Issue #143, opened 2015-02-25
Sprites of hanging corpses are not centered correctly


AXDOOMER:
Sprites of hanging corpses are not centered correctly. This causes the sprites to appear to be moving around on the ceiling when the player circle-strafes around them. 

![hang](https://cloud.githubusercontent.com/assets/6194072/6364736/ef16f7dc-bc78-11e4-82d8-71cf8f4f525d.png)

Proluiz:
same of #135 .

Issue #142, opened 2015-02-25
Candles are drawn inside the floor


AXDOOMER:
Unlike the original Doom, candles in FreeDoom are drawn inside the floor. This may be the result of auto aligning of sprites, which on some editors, put the sprites a few pixels inside the floor. In the original Doom, candles are drawn over the floor.

jewellds:
I've had a go at fixing the offsets with pull request #144. It's still slightly below the floor line, because if you place it too high it looks like it's floating, but now it's centred and a bit higher than it was.

Issue #135, opened 2014-11-04
Hanging Freedoomer has wrong offset


Proluiz:
The X posing of hanging freedoomer (DB actor named has: Hanhing pir of legs) is wrong for some reson plese try to fix the x offset :(

PR #144, opened 2015-03-04
Reshaded lamp sprites and fixed offsets

jewellds:
The first commit improves the shading on two lamp sprites, as discussed here: http://www.doomworld.com/vb/freedoom/71902-re-shaded-lamp-sprites/

The second commit adjusts the offsets on all hanging corpses, as well as the candle, fixing #143, #135  and #142.


fragglet:
Thanks!

PR #145, opened 2015-03-06
FreeDM MAP08 update

AXDOOMER:
I made some updates to my map and the file has been sitting on my computer for a long time. I created a fork and now I want to pull it in the FreeDoom project. At the same time, I changed my map's name in Fraggle's BEX file.


chungy:
Thanks :)

PR #162, opened 2015-08-13
Add Hebrew translation of freedoom2.desktop

GreenLunar:


chungy:
Thanks. I've merged all of these.

GreenLunar:
Yes, you are correct, but I always end up with several commits instead of one, at least by using the web interface.

chungy:
Multiple commits is fine, I was just mentioning keeping them in the same branch. I don't have experience with the GitHub web UI for this stuff though, maybe it's a limitation you can't overcome only with it.

It's only a bit more work for me to do when merging them, it's not a real issue.

GreenLunar:
By the way, after reading the comment in this file, how does level 32 end?

I have got into a the biggest room, so far, with a flying can generating monsters, and I am stuck there.

GreenLunar:
Of https://github.com/freedoom/freedoom/releases/download/v0.8/freedoom-iwad-v0.8.zip

PR #165, opened 2015-09-18
patches: New wall03_3, Improved shading on door2_1 and door2_5

jewellds:
This commit contains two updated patches, and one replacement.

Door2_1 and door2_5 are updates of my two door textures, to fix some problems with the shading. 

Wall03_3 is a completely new patch to replace the existing wall03_3, which wasn't metrically compatible with Doom's, and frankly looked a bit odd.

More information on the forum thread here: http://www.doomworld.com/vb/freedoom/72597-texture-improvements/


chungy:
Thanks :)

Issue #150, opened 2015-04-25
c2m9 - No yellow skullkey


Proluiz:
Skullkey is not really needed to finish map but his missing make the cyberdemon area unreachable

Issue #60, opened 2014-02-25
C2M2 issues


paulgilb:
A couple of issues with C2M2:

2 monsters do not warp in (and thus cannot be killed).
The teleport secret cannot be reached, which also makes the supercharger unobtainable.

PR #172, opened 2015-11-29
Fix: level issues #60 and #150

NickZ:
Fixes issue #150.
key in blue skullkey room was accidentally set to blue
skullkey, resulting in two blue skullkeys. Changed to yellow skullkey.

Fix: C2M2 Issues #60
Fixed missing sector tag on monster pen teleporter, able to kill every
monster now.
Added hidden door to teleporter, moved secret tag off teleporter to
ajoining sector, fixed teleporter to teleport to hidden room.


chungy:
Thanks

Issue #35, opened 2014-01-15
MAP30 Split


Catoptromancy:
MAP30 should much shorter and almost directly lead to boss fight. It should also be harder as it is also shorter.
Can rip boss fight out of current MAP30.

The leftovers of MAP30 could be converted into an phase1 episode map. e3m7 or e3m9 and leave it almost intact.

PR #173, opened 2015-11-29
Per #35: Simplified Map30, created C3M7 from scraps

NickZ:
As per #35, I have simplified Map30 so that it only contains the players' starts, and the boss room. I then created C3M7 out of everything from Map30 except the boss room, replacing the FreeDoom2 things and textures.

C3M7 still needs to be tweaked and rebalanced for FreeDoom1.


SuperSomariDX:
HoM behind you at spawn on C3M7 by the way.

PR #175, opened 2015-11-30
C1M1: Removed "collapsing wall"

NickZ:
The "collapsing wall" effect in the first hallway in the game was poorly
done. It just fell instantly with no sound effects or animation, and
just ended up looking like a visual anomaly. Not something you want to
show the player on their first impression, so I took it out.


fragglet:
I think I mentioned this to @jmtd a while back and he felt the same way. But yes, I agree.

jmtd:
I can't recall my thinking at the time but I think it was the best that could be achieved with vanilla effects only, which is weird because I used plenty of boom effects elsewhere. The intention was to suggest that the place was falling apart, to make the player feel uncomfortable and not dawdle, but It's a fairly superfluous effect.

PR #180, opened 2015-12-02
Updated content

Blastfrog:
See https://www.doomworld.com/vb/post/1528750 for more details.
This commit only partially covers sodafd_v3.wad, I'll fill out more soon.


Blastfrog:
Never mind, redoing this as multiple commits.

PR #183, opened 2015-12-03
Add GeckoYamori's sound effects

Blastfrog:
Also included are two new sounds by me, replacing the current rocket launcher sound.


chungy:
Would you mind splitting up your commit so that each one contains an independent resource to be added/replaced?

PR #187, opened 2015-12-04
titlepic/create_caption: try alternative fonts

rjy:
Do not FTBFS if the user does not have ghostscript / gsfonts installed.
Instead try some other fonts commonly installed by Linux distributions.

The patch is written to be easily extendable with extra font names, but
to be independent of the order of "convert -list font" output. Each
listed font is searched for in turn, and the first one found is used.

### DIFF
--- a/graphics/titlepic/create_caption
+++ b/graphics/titlepic/create_caption
@@ -1,11 +1,25 @@
 #!/bin/sh
 
+find_font() {
+	local font fontlist
+	fontlist=$(convert -list font | awk '$1=="Font:" { print $2 }')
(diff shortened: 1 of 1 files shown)


rjy:
The Real Reason: I got terribly sick of the build failing because I don't have gsfonts installed.

rjy:
"local" is also part of dash, so I figured it was portable enough to use.

chungy:
shellcheck doesn't complain about it. It's usually pretty good at caching non-POSIX extensions. (It does complain about unquoted variables, but for this use, it's not a real problem)

PR #188, opened 2015-12-04
Many new graphics

Blastfrog:
Haven't bothered with the symlinks just yet.


chungy:
Per IRC discussion, I've pushed rewritten/cleaned-up versions of these commits instead. thanks

PR #193, opened 2015-12-09
More stuff

Blastfrog:
New gib sound, updated texture, fixed aspect ratio of pistol, and addressed some issues with map05.


chungy:
I'm not convinced that the change to the pistol is needed. It's another superficial "more like Doom" change that's unnecessary. Thanks for the rest though.

PR #195, opened 2015-12-10
Icons and a switch

Blastfrog:
Yep


chungy:
I had to rewrite your commit that tried to change symlinks. Please do not mix Cygwin file managemnt and non-Cygwin versions of Git.

Blastfrog:
Weird thing is, I was using Cygwin, with the native version of Git. There must've been an error in the installation somewhere, I dunno.

NickZ:
I recommend mobaxterm. It's very easy to run a local terminal and install
packages, just type "apt-get install git" and such

PR #196, opened 2015-12-10
MAP05: Gameplay and compatibility fixes

fabiangreffrath:
- Remove the extraneous secret room accessible right from the start.
- Increase the floor height of sector holding the red fake key, so it
  is better visible from the outside.
- Entirely remove the dull rectangular sector that once surrounded the
  Doom logo.
- Change the texture of the wall leading to the backpack secret.
- Remove the teleport from the super-secret room (accessible from the
  backpack secret) that lead to the northern-most sector behind the
  bars, instead add a switch to the super-secret room to lower the three
  bars on the left of that sector.
- The three bars on the right of that sector are now lowered when the
  switch lowering the fake red key and unveiling the real red key is
  activated.
- Move the invisible secret switch revealing the rocket launcher
  secret to the switch at the northern end of the northern-most
  corridor (it is still hard enough to discover, after all).
- The three eastern blue bars are now directly switchable (just as
  their western counterparts), remove switches from the walls on
  either sides.
- Turn the sector surrounding the yellow key back into damaging lava
  (it is in a hell area after all), but leave it escapable.
- Also, turn the sector surrounding the rocket launcher back into
  damaging nukage.
- Replace the two Barons guarding the red key by Cacodemons.
- Replace a stuck demon in the yellow key area by a Chaingunner.
- Change linedef on lift into yellow/red key area to type 88.

- Baptize the map "Waste Water Basin" in 'lumps/fraggle/freedoom.bex'.

TODO: Thing placement!


NickZ:
Did you make your changes before Sodaholic submitted his for this map? https://github.com/freedoom/freedoom/commit/72aeffbfe5c63e5d9574b9654f6b2827d60a14ba

chungy:
Fabian: Would you please review the tools that you're using to make the edits? I just removed extranous lumps from your edited version of the map. It's not the first time I've had to do it; the last times you submitted changes, they broke MAP01 entirely due to GL Nodes being included in the IWAD that didn't match the map.

fabiangreffrath:
> Did you make your changes before Sodaholic submitted his for this map?

Initially yes, but I rebased my map entirely on his submission.

> Would you please review the tools that you're using to make the edits?

Oh, I am sorry. I use Eureka+GLBSP and I guess that the latter adds these additional lumps that cause all the confusion. Is there a better/non-GL node builder available for Linux?

chungy:
> Oh, I am sorry. I use Eureka+GLBSP and I guess that the latter adds these additional lumps that cause all the confusion. Is there a better/non-GL node builder available for Linux?

I usually use [ZenNode](https://github.com/Doom-Utils/ZenNode). zdbsp is popular too and should work.

PR #201, opened 2015-12-17
levels: map05: gameplay tweaks

rjy:
weapons

- move chaingun to outside blue key building's (secret, locked) back entrance
- (don't think a supershotgun is needed -- no sufficiently nasty monsters)

route simplification

- red key and rocket launcher swapped
- red key now on donut pillar outside water pump building
- no need to make long and obscure backtrack from water pump building.
- cacodemons now teleport to donut pillar when switch approached.

rocket launcher secret

- rocket launcher visible in gloom on ledge, where red key used to be
- add a couple of lamps to highlight it
- blocking bars changed to resemble columns of water, as a hint
- secret switch in the wall behind the donut pillar removes these columns
- water columns on pump machine also lower
- add a third lowering water column next to the switch in the rock

other fixes

- yellow key monster teleporter redesigned to deploy its payload faster

DW: https://www.doomworld.com/vb/post/1534058


chungy:
Thanks

PR #204, opened 2015-12-19
levels: c2m1: remove surplus secret

rjy:
Sector 168, the nukage surrounding the entry teleporter, had the secret
area type. I assume this is an error, since it contains nothing and is
easy to miss. Change it to a damage type, as was likely intended.


NickZ:
i'll change you to a damage type


WITH MY FISTS

PR #205, opened 2015-12-20
More stuff

Blastfrog:
I dunno if the flame sound effect can actually be used. Depends if the Creative Commons Attribution license doesn't conflict with Freedoom's modified BSD license. The rest is fine, though.


NickZ:
Pushed updated missile sprites (without the update front graphic), the explosion sprites, DSSLOP, barrel explosion, and textured metal on the saw. the rest of the stuff needs some work.

PR #209, opened 2015-12-24
Create the remaining missing PC speaker sound effects; Change a few e…

GhostlyDeath:
…xisting sounds.

 * The item pickup sound has changed to not be a truncated powerup sound.
 * The second and third sets of Zombie and Serpent sounds are also now off-pitch.


fragglet:
Nit: for future submissions, please follow the convention for Git commit descriptions as documented in README.adoc.

fragglet:
Thanks.

Issue #97, opened 2014-09-08
Build process is broken while creating helptext.png


FredBezies:
I try to build wad from git, and it crashes when creating helptext.png

Here is the error log

`python smtextgen helptext.png 320x200        \`
`150,5   "file:helpttl.gif"                \`
`10,25   "Weapons"                         \`
`80,25   "file:../../sprites/shota0.gif"   \`
`150,25  "file:../../sprites/sgn2a0.gif"   \`
`210,20  "file:../../sprites/mguna0.gif"   \`
`10,40   "file:../../sprites/launa0.gif"   \`
`80,40   "file:../../sprites/plasa0.gif"   \`
`150,40  "file:../../sprites/bfuga0.gif"   \`
`240,40  "file:../../sprites/csawa0.gif"   \`
`10,67   "Ammo"                            \`
`53,67   "file:../../sprites/clipa0.gif"   \`
`68,62   "file:../../sprites/ammoa0.gif"   \`
`110,67  "Shells"                          \`
`160,67  "file:../../sprites/shela0.gif"   \`
`180,62  "file:../../sprites/sboxa0.gif"   \`
`220,67  "Backpack"                        \`
`290,55  "file:../../sprites/bpaka0.gif"   \`
`10,90   "Missiles"                        \`
`76,78   "file:../../sprites/rocka0.gif"   \`
`95,82   "file:../../sprites/broka0.gif"   \`
`170,90  "Energy"                          \`
`220,90  "file:../../sprites/cella0.gif"   \`
`240,81  "file:../../sprites/celpa0.gif"   \`
`10,115  "Health"                          \`
`63,113  "file:../../sprites/bon1a0.gif"   \`
`78,112  "file:../../sprites/stima0.gif"   \`
`96,107  "file:../../sprites/media0.gif"   \`
`128,107 "file:../../sprites/pstra0.gif"   \`
`170,115 "Armor"                           \`
`220,113 "file:../../sprites/bon2a0.gif"   \`
`240,105 "file:../../sprites/arm1b0.gif"   \`
`280,105 "file:../../sprites/arm2b0.gif"   \`
`145,140 "Map"                             \`
`175,130 "file:../../sprites/pmapa0.gif"   \`
`10,140  "Overdrive"                       \`
`84,132  "file:../../sprites/soula0.gif"   \`
`110,132 "file:../../sprites/megaa0.gif"   \`
`10,163  "Night vis."                      \`
[...]

Issue #85, opened 2014-06-23
DM12: Misaligned texture


AXDOOMER:
I forgot to report this one last time. 

Check this video, you'll see where it is between 0:25~0:26 when the player turns left and fires the BFG on the metal "support" texture (the one that's misaligned) to kill the other player. 

http://youtu.be/YJjlO9A1mDs?t=25s

Issue #48, opened 2014-02-01
Some sprites have speckles when playing with D2TWID


fragglet:
Screenshot: http://i.imgur.com/9U7X25O.png

I get this when playing with D2TWID, but only with Chocolate Doom. When using PrBoom+ they don't appear. Only appears on some sprite frames; it's particularly noticeable on the Baron / Spider sprites. 

D2TWID has replacement PLAYPAL / COLORMAP lumps so that may be to blame.

NickZ:
closed with #214

chungy:
Still an issue, actually.

mc776:
Are there any sprites left that still do this?

mc776:
Link for convenience https://www.doomworld.com/idgames/levels/doom2/megawads/d2twid

mc776:
I'm also getting this with PrBoom+.

Is this only one colour in the palette or are there more?

mc776:
Why I think the only affected thing is colour 255:

PrBoom+
![456](https://user-images.githubusercontent.com/24984517/236759318-3c743405-0df0-4c85-a90f-47add412a347.png)

Chocolate (btw how do you get DeuTex -as to use a specific IWAD? The easiest way to check these is with a dead octaminator but I'm stuck with release)
![123](https://user-images.githubusercontent.com/24984517/236759321-e441e5c6-99be-498f-80b2-86220f491676.png)

By the way I did actually test #967 and as far as I am at all able to tell there are no cyan speckles anywhere.

mc776:
I think I linked this issue without understanding what linking does. Reopening until we have an agreement on a proposed permanent workaround.

mc776:
It doesn't seem like any sprites have this problem anymore, not even any of the recent stuff (checked fist, backpack and bronzes as the highest-risk items).

I think we can close this for now and re-open it or post a new issue (or even just a PR since we all know what to look for now) if it ever comes back up again.

PR #214, opened 2015-12-28
levels: made C1M2 vanilla compatible

andwj:
* Simplified lighting gradients in a few places, especially the
  numerous circles in the room to the E (large visplane overflows)

* Changed translucent glass window in S of map.

* Reworked the diagonal translucent window (near 434,-1364) and lowered
  the room on SE side to accommodate MIDGRATE texture.

* Replaced generalized linetype in backpack secret (in W of map).

* Lowered the corner area just past the broken sky room.  By chance this
  fixes the odd-looking stairs a bit further on.

* Made the translucent glass in SE of map simply be opaque (the player could
  not see much through it anyway).

* Changed doorway near (900, -1700) to look more like a doorway (it was
  previously looking too much like a window).

* Replaced translucent window just S of the broken sky room -- now solid
  (this window was high and player could not see much through it).

* For teleporting monsters into blue key room, replaced BOOM silent teleport
  linetype #269 with vanilla compatible #97 (vanilla has no silent ones).

* Replaced BOOM scroll linetype #85 with vanilla linetype #48.

* In blue key room, changed monster closet to use a vanilla linetype instead
  of a BOOM generalized one.  Also made closet have a lowering floor instead
  of a raising ceiling (to preserve the geometry there).

* Cleared a line with tag #7 and a BOOM generalized type, there was no matching sector.

* Replaced BOOM generalized type for switch just past red-keyed door,
  reworking the door so the lights do not get cut off.

* Fixed linedefs with missing textures.

* Large rework of the "blue room" -- can no longer use BOOM colormap fx,
  hence made all the walls, floors and ceilings use blue textures.


chungy:
thanks

fragglet:
Side note, please don't write #<number> in a commit or PR description unless referencing an issue - the tag numbers you've mentioned here have added irrelevant cross-references to several bugs.

andwj:
On 12/28/15, Simon Howard <notifications@github.com> wrote:
> Side note, please don't write #<number> in a commit or PR description unless
> referencing an issue - the tag numbers you've mentioned here have added
> irrelevant cross-references to several bugs.

Obviously I would not have used # if I knew about this feature of github.

-- Andrew

chungy:
I think the # symbol isn't too bad, just avoid using it in the future.

However, a more important stylistic thing: try to keep your description lines within 72 characters. Going over that seriously makes it look awkward in a `git show` or `git log` display.

andwj:
On 12/29/15, Mike Swanson <notifications@github.com> wrote:
> However, a more important stylistic thing: try to keep your description
> lines within 72 characters. Going over that seriously makes it look awkward
> in a `git show` or `git log` display.

Ok, noted.

Cheers
-- Andrew

PR #216, opened 2015-12-30
new Short Techno Lamp

CWolfRU:
New Tecno lamp (short)


chungy:
thanks. I've rewritten the commit message to match the expected style.

PR #223, opened 2016-01-04
Sprites: raymoohawk's tall lamp fix

CWolfRU:
Raymoohawk's fixes for tall lamp


chungy:
I'll appreciate if the commit style guidelines were followed better. Particularly the history rewriting (shouldn't have all the revert-revert-revert commits in a PR like this) and references to other commits by hash instead of action stamp; additionally, the directory is "sprites" and not "Sprites" and the last one should be "buildcfg" too.

I'm rebasing myself to fix these, but please keep it in mind for the future. :-)

PR #224, opened 2016-01-05
sprites: new backpack and techno pillar

CWolfRU:
New backpack and techno pillar


jmickle66666666:
I don't think that backpack is an improvement. It's not very readable, while the older one is very clear. 

The new techno pillar is a cool design but still needs some shading work imo.

chungy:
Please keep your commits topical: don't combine multiple different things into the same one.

All that said, I generally agree and I think a traditional column shape works better for a sprite that has the same face from all angles.

PR #230, opened 2016-01-18
graphics: updated statusbarface

raymoohawk:
added the new mazmon statusbar face


chungy:
I don't really like this one, I'm going to have to decline it for now.

Also, it's typical style for the person that made the resource to be the author of the commit, not the committer.

raymoohawk:
ok, got it, tho i still think there should be a designated person for doing commits. doing both the art part and the technical part is kinda tedious, at least for me

PR #235, opened 2016-01-24
Patches: mazmon switches

raymoohawk:
added mazmon's sw3s0 and sw3s1


chungy:
Please use --author if you do commits on behalf of others. I've rewritten this commit to reflect that. Thanks :)

Issue #239, opened 2016-02-06
Sprite not properly aligned on its axis


AXDOOMER:
I don't know which sprite this is, but here's a screenshot. 
![doom04](https://cloud.githubusercontent.com/assets/6194072/12869871/9c1c8d48-ccf9-11e5-91f4-bc601e0fb1d4.png)
When you turn around it, it moves, so this means the sprite is not centered on its X axis.

andwj:
The sprite is CEYE (frames A,B,C)

I will fix it shortly.

PR #242, opened 2016-02-13
sprites: fixed offsets on CEYE and TLMP

andwj:


chungy:
Thanks. You can just include the text "closes #239" or such in the future. It would automatically close the issues in the tracker once pushed to this upstream repo :)

PR #247, opened 2016-03-06
make the build reproducible

a-detiste:
Hi,

This tiny tweak is needed to make the build reproducible.

Appart from having high packaging standards; having reproducible
build also sometimes help to identify some weird/old bugs;
and here that doesn't hurt either in anyway.

https://wiki.debian.org/ReproducibleBuilds/TimestampsInDocumentationGeneratedByAsciidoc


chungy:
thanks

PR #268, opened 2016-07-20
sprites: new pistol

Voros2:
Uses Blastfrog's new pistol


Voros2:
I know this was rejected at first by Nickz, but Nickz at Doomworld later revealed that he liked it afterwards. 

So, Nickz, if you are reading this, why don't you commit this into the branch? The rest of us will see what Freedoom truly feels like with this pistol, with the autobuild.

chungy:
The sizes of all the files are different. Is that accounted for in `buildcfg.txt`?

Voros2:
I was wondering about that. Unfortunately, I'm no genius and I can't find any tutorials on how build.cfg works. 

How does it work?

jmtd:
Hey @Voros2 , thanks ok, thanks for making the effort to raise PRs! The buildcfg.txt file is an input file into the program deutex, which builds Freedoom. It contains a list of all the resources that are to be copied into the WADs.

If you look at the [buildcfg.txt file](https://github.com/freedoom/freedoom/blob/master/buildcfg.txt#L2315), it has entries for the pistol sprites as follows

```
PISGA0  -128    -100
PISGB0  -109    -83
PISGC0  -123    -88
PISGD0	-122	-88
PISGE0  -128    -100
```

Those numbers are (I think) offsets that control where the pistol will appear on the screen and they're sensitive to the size of the pistol graphics. These values might need tweaking for the new pistol, so this Pull Request should include changes to the buildcfg.txt file.

If the new pistol sprites were original in a WAD which had the new pistol aligned properly, the easiest thing to do would probably be to extract that WAD using deutex (which will generate a wadinfo.txt file in the same format as freedoom's buildcfg.txt) and copy the offsets out of that.

Voros2:
Ok then. I'm closing this pull request and the rest. Gonna have fix my fork again.

PR #269, opened 2016-07-29
New logo

Voros2:
Jewellds' new Freedoom title logo.

Replaces m_doom.gif


Voros2:
I doubt the offset will have to changed for this to work properly.

CraftedPixel:
I like the original because it looks more 3D. Also, your logo doesn't fit with the dimensions correctly.

Voros2:
I'll provide the offset values then.

But I won't unless the pistol gets merged/rejected again.

fragglet:
Looks good. It's better to use this as it's less Doom-esque. Also matches the FreeDM logo which is nice.

PR #270, opened 2016-08-01
New player death sounds

Voros2:
Uses SeventhSentinel's new sounds. Original and better than a zombie death sound and the Wilhem Scream.


Voros2:
How come this hasn't been added yet? It is a legit submission.

PR #271, opened 2016-08-05
level: map09

Voros2:
New map for Phase 2.

I am not the author of this map. To contact the author, the email is sajber@sajber.info

This map was specifically made for Freedoom.


Voros2:
I also forgot to mention: The author calls the map "Datacenter". I think the "Mine HM09" name should be replaced with this instead.

PR #275, opened 2016-08-11
New MIDIs

Voros2:
Uses blueworrior's new MIDI submission.

I assigned the tracks to slots that use double duty on music tracks, and made sure it was appropiate enough for the slots.


Voros2:
Added some more MIDI from blueworrior that wasn't submitted before.

Voros2:
Blueworrior gave me higher pitched versions of the MIDI.

fragglet:
Cool. Blueworrior isn't listed in CREDITS - can you update it? We want to make sure everyone who has contributed is credited.

fragglet:
Thanks!

PR #277, opened 2016-08-13
lumps: new demos

Voros2:
Some of the demos have barely any gameplay, so I made some demos for those specific ones.
And in demo1 of Phase 1, the linedefs using Boom features are shown as HOMs. I made another demo for this.

All were recorded in PrBoom+, compatibilty level 9, skill UV.


chungy:
Thanks, but demos are only really made during release time. Interim development between versions doesn't really matter to maintain demo sanity. Additionally, it is standard that demos are only recorded in vanilla compatibility for the IWADs.

Voros2:
Ok then. Tell me when you guys are about make your next release, and I can prepare demos for it.

PR #279, opened 2016-08-20
revised worm death

Voros2:
hexereticdoom made a worm death scene using red blood instead of the current orange blood.

It looks better and makes more sense really.


fragglet:
You haven't committed this correctly - your commit adds new files rather than modifying the existing ones.

Voros2:
Oh shit, i forgot that it should be lowercase.

I'll reupload it in a few hours.

Voros2:
Fixed.

fragglet:
When deleting files, you can delete them all in a single commit - you don't need to make a commit per file.

Voros2:
I know, but I couldn't find the revert button for some reason :P, so I resorted to deleting them one by one. But at least it's fixed now.

Voros2:
What about the crediting Hexereticdoom?

fragglet:
Oh, you're right - can you submit another pull request to fix it?

Voros2:
Yes I can, but I'll need to gather his information first. Until merge sajbear and Voros(:P), as given in another PR.

PR #316, opened 2016-10-15
New interpic

Voros2:
Since Sodaholic/Blastfrog is gone for the month, I'd like to submit this new intermission screen he made.


chungy:
Thanks, looks good.

Issue #323, opened 2016-10-19
Aquatex textures affected by Tutti-Frutti effect.


jewellds:
Many of the new Aquatex textures are 64x64 (or smaller) and exibit the Tutti-Frutti effect in engines that have that bug (EG: Vanilla, Chocolate Doom). This could be fixed by expanding them to be 128 pixels tall.
![doom49](https://cloud.githubusercontent.com/assets/6622575/19507918/deb308b0-95cd-11e6-8a93-73600bfe268c.png)

Additionally, PIPE15 and PIPE16 are completely blank.
![doom60](https://cloud.githubusercontent.com/assets/6622575/19507922/e691d21e-95cd-11e6-9897-997205c235e4.png)

Test wad for Aquatex: [aquatestv5.zip](https://github.com/freedoom/freedoom/files/538100/aquatestv5.zip)

Voros2:
It seems I forgot to upload pipe15 and pipe16. That can be fixed.

As for the 64x64 textures, removing them from the IWADs would be a better idea. Or converting them into flats (if not already). 

You say this is only in vanilla/chocolate?

jmickle66666666:
The 64x64 textures are mostly just texture versions of the flats, iirc. It can be fixed by stacking 2 of them as patches in 128x64 textures. The original texture set was designed for boom, so there's bound to be a little work making sure they're all fine for vanilla.

Voros2:
Yes you do remember correctly. So I'll just remove them. Ok with you? It'll save space too.

jmickle66666666:
Well they are still handy as textures so I'm not sure removing them is necessary.

Voros2:
Well, I'd like to see this issue get fixed soon, and there's plenty of fabulous Aquatex textures inside, more than 200. Removing some non-vanilla-compatible textures, especially when they're just flats converted to textures, shouldn't be a big problem to mappers (because they've got over 200 more to choose from). 

Q: if the height is 64, but not the width eg 128, will it still have tutti frutti effects?

jmickle66666666:
The height must be 128 for it to avoid the tutti frutti effect. You can create a texture definition and stack two of the 64x64 patches to make it a 128x64 texture without having to waste any filesize. This will entirely fix the issue.

Voros2:
Oh yeah! That can fix patches easily. I'll get to work with that then.

Voros2:
Do the X offsets need to be changed also?

jmickle66666666:
[...]

PR #324, opened 2016-10-19
Aquatex fix

Voros2:
Meant to be a solution for https://github.com/freedoom/freedoom/issues/323 (tutti frutti errors).
Redefined several texture definitions in textures.cfg so that all textures have a height of 128.


fragglet:
Is this confirmed to fix tutti-frutti?

My memory of vanilla tutti-frutti issues is that textures cannot tile properly if they are not a power of 2 (16, 32, 64, 128). So if you have a texture that is 12 pixels tall or 24 that can be a problem, but if it's 64 it shouldn't make any difference.

Voros2:
I did what jewellds showed on the issue 323. Most of the textures are now 128 in height.

fragglet:
[Looks like I was wrong!](https://doomwiki.org/wiki/Tutti-frutti_effect)

> If a wall is taller than its texture, then the texture is tiled vertically. But the Doom rendering engine starts new tiles only at heights of 128, 256, 384, etc. If the texture is less than 128 units tall, horizontal bands will be left undefined.

fragglet:
Merged - thanks.

Voros2:
Shouldn't https://github.com/freedoom/freedoom/issues/323 be closed now, or are you keeping it open for any further problems?

PR #325, opened 2016-10-20
New cursor for FreeDM

Voros2:
A cursor created by andrewj that never got added. Honestly, I believe it's nice for FreeDM, since it's all about deathmatch.


fragglet:
Cursor?

buildcfg.txt
```diff
@@ -801,8 +801,16 @@ M_RDTHIS	0	0
 M_OPTION	0	0
 M_QUITG	0	0
 M_NGAME	0	0
+#ifdef ULTDOOM
+#ifdef DOOM2
```
fragglet: #ifdef doesn't combine in this way. This will remove the M_SKULL lumps from the main Freedoom IWADs.

It's better if you use #else, eg.

#ifdef FREEDM
...
#else 
...
#endif

Voros2:
Like that?

fragglet:
Much better - thanks!

PR #327, opened 2016-10-28
Change 1337 to 777 

Voros2:
1337 is a bit too childish and is basically a reference to a stupid internet meme. So, for now, a simple triple 7 would be nice. I'll leave it upto you two, the maintainers.


chungy:
I don't agree with it being childish. Freedoom was born on the Internet, and around the time that 1337 was at its peak. I find it appropriate and nice :-)

fragglet:
If you want to fix something childish that's a clear mistake, please rededicate your energy to replacing the "AGM" textures.

Voros2:
Whatever you say.

Issue #329, opened 2016-10-30
Shotgun sprite error


uhbooh:
![doom14](https://cloud.githubusercontent.com/assets/22185239/19839853/222a2cfe-9ec0-11e6-955b-b8827552dfac.png)

The three shotgun pumping frames are taller than 128 pixels, and thus suffer from the repeating sprite glitch. This thread explains how to fix it.

https://www.doomworld.com/vb/doom-editing/71482-importing-tall-sprites-to-work-in-vanilla/

Talon1024:
Actually, if you look at SHTGC0 in SLADE, you'll notice that the 'glitch' you're pointing out is actually part of the sprite.

AXDOOMER:
How is it part of the sprite?

Voros2:
Simply, the sprite is too tall for vanilla limits.

Talon1024:
OK, I just tested the sprites in vanilla Doom 2, and I'm noticing glitches on SHTGB0 and SHTGD0, but not SHTGC0, unless they are blending in with the floor texture.

I don't see any glitches in the picture above, though. I had assumed the "glitch" you were pointing out was just weird shading on the hand, which is part of SHTGC0.

AXDOOMER:
There is clearly a bug, but maybe it can be fixed without changing the position of the sprite like @Voros2  did. 

> https://www.doomworld.com/vb/doom-editing/71482-importing-tall-sprites-to-work-in-vanilla/

@Voros2  clearly didn't check this thread before submitting his fix.

Voros2:
To be honest, it is a bit confusing, both the thread and the Doomwiki article. I'm still trying to grasp the whole concept of this.

AXDOOMER:
We need to fix this bug to the source, which means fixing DeuTex. The wiki article explains what needs to be done to fix it. 


```
In Doom and Doom 2, id Software specifically bypassed this issue by splitting these long columns 
into two posts: one of 128 pixels, and a second with the excess. Plutonia and TNT's IWADs, on 
the other hand, were apparently compiled from the source graphics with the third-party tool 
DeuTex, which did not take this potential vertical wrapping into account.
```

https://doomwiki.org/wiki/Shotgun_sprite_wraps_vertically_in_Final_Doom

fragglet:
Can we just crop SHTG[BCD]0 to be shorter than 128 pixels? Or will it screw up the animation too much?

Voros2:
[...]

PR #330, opened 2016-11-02
Shotgun fix

Voros2:
Meant to be a quick fix for https://github.com/freedoom/freedoom/issues/329 by chainging the offsets.


Talon1024:
[There is no bug or glitch](https://github.com/freedoom/freedoom/issues/329#issuecomment-258562689).

AXDOOMER:
This is not a fix, this is only a workaround. The real bug in in DeuTex.

Voros2:
Will this not be merged? Else, that small glitch will be present in vanilla/chocolate Doom

PR #332, opened 2016-11-05
Nex credo maps

Voros2:
Jayextee has kindly allowed the usage of his maps from the mapset Nex Credo to Freedoom.
Here is the revised map lineup I put together with tips from uhbooh.
All maps are vanilla, have difficulty setting implemented and has player starts.


Voros2:
Any news on adding these great maps?

fragglet:
Looks good, but these replace some existing levels, and maybe it would be good to document why these particular levels have been chosen for replacement.

Given that these come from a particular mapset, is it worth rearranging some levels to keep them together (like we did with C4)? Or is it fine to separate them?

Voros2:
Well, simply for the players' sakes. They require a good knowledge of the levels first, thus players will probably get stuck and ragequit (that's bad). The automap itself is basically a mess for some of them. That, and most of those maps would be near-impossible to vanillafy.
I did rearrange them to fit Freedoom's progression (if any) as much as possible. I plan on editing Freedoom's maps soon so that they feel more progressive than random, while using the recently added Aquatex textures. 
What do you mean by seperating them?

fragglet:
What I meant was that for Double Impact we imported the whole episode as E4 and kept the levels together. Is it worth doing the same for Nex Credo - keeping them together as a sequence - or not? I guess preserving Freedoom's progression is a good reason.

We still don't have a C3M5. Is it worth moving one of the levels that's being replaced to that slot? Then we'll have a complete set of levels!

Voros2:
Whether the sequence should be kept? That is a good question. By keeping the sequence Freedoom would probably lose some good maps along with "bad" ones. But not keeping the sequence, may or may not help Freedoom. It depends on Freedoom's progression really.

The C3M5 problem. I'm sure over time, someone will step to fill it in (Soundblock, the creator of the Echelon megawad, had a go with it, but couldn't find a way to Hellify it though). Nex Credo is a Doom 2 mapset, so I don't think that would be wise.

PR #336, opened 2016-11-21
Endpic and victory2

Voros2:
Both endpic and victory are pretty much the same. So I took screenshots to make it seem better, by making it seem like a scene of the battle has been frozen. I tested them and it looks fine ingame.


fragglet:
My main criticism of VICTORY2 is that the close-up of the player sprite is very pixellated. Unavoidable I know, but it wouldn't look so bad if the shot was taken from a different angle.

Voros2:
I tried it in different angles, and frankly, they don't look very good either. Putting the camera near the Assault Tripod makes it feel awkward. Putting it on the side (so that both of them can be seen) loses a bit too much quality IMO. Putting it next to the player felt like the best approach, as it makes the player the main point of interest of the picture. 

Plus, everything is pixelated in Freedoom, so I don't see how thats a problem.

PR #338, opened 2016-11-22
Cwolf's new IoS sprite

Voros2:
Since Cwolf has apparently gone off into the army, id like to submit his sprite on behalf of him.


fragglet:
We're getting rid of RMS? :(

chungy:
Since the boss brain is only an insider joke (both in Doom and Freedoom), I don't think we ought to replace it with a serious sprite.

Voros2:
But what if it becomes serious? The current IoS level is a clone of Doom 2's IoS level. It's not that impressive, ESPECIALLY now, where people have sourceports with mlook enabled. 
Sodaholic mentioned once that he would like to see a different IoS level:

The final level of phase 2 would take place inside this being and should be a completely different gimmick than raising a lift and shooting rockets. Perhaps have a series of switches you have to activate that temporarily expose the boss's weakpoint, and use the Keen actors as destructible shielding that you have to gradually widdle down to get to the Romero head.

Voros2:
Quick test ingame:
http://m.imgur.com/TOk1WcH?r

gamegenten:
I like this new one. I use the BossBrain actor on multiple levels and was going to replace the RMS head sooner than later.

AXDOOMER:
That was a nice head replacement, but like John Romero's head, it's nothing serious. I wonder if RMS would agree to have his head on a sprite in a project distributed under the BSD license though. He's even the bad guy and you must kill him to win the game. I prefer the "serious" sprite.

PR #340, opened 2016-11-25
SSG offset fix

Voros2:
The SSG wasn't properly centered and two of the frames were a bit (only a bit) "janky" during the reloading sequence. So, I centered it, fixed those two frames with their offsets and attempted to make the whole animation more fluid and natural-like. Tested ingame and it works.


Voros2:
Hi. Whats the hold up?

PR #348, opened 2017-01-03
Egyptian tex fix

Voros2:
Agenten made some small edits to the Egyptian textures. Basically, agenten tried to make them look neater by removing the jarring yellow using Slade's color remapper. The change is mostly subtle but you'll be able to see the difference when comparing them side by side.


fragglet:
This looks like an improvement, but if you compare with 8f2843caf72d5596a30609a4243313d0d7a3e343 (the original commit that added the Egypt textures) you'll see that there are some Egypt textures that haven't been updated. Specifically DRFRONT and STWALL have been missed. As this changes the tone of the wall color for these textures, those two textures are going to look inconsistent unless they're also updated to match.

Voros2:
STWALL and DRFRONT added.

fragglet:
Great! Thanks.

Issue #10, opened 2014-01-02
FreeDM should have its own music


fragglet:
Currently FreeDM uses the same music as Freedoom. Some of the music tracks are okay for deathmatch but others are too slow. For deathmatch levels we want something more fast-paced and rockin'.

It might be possible to selectively reuse some of the existing music tracks to use for FreeDM to put together a more appropriate soundtrack.

mc776:
After #1009 it seems DM14 and Map32 are the only duplicates left.

rostuhan:
Oh, you've meant D_DM14 and D_DM32.

mc776:
huh, well I guess there are no duplicates between FreeDM and singleplayer now.

If someone does want to do a new track for one of those DM maps this is not meant to discourage that but I think we can close this issue, subject to someone else verifying this because I do *not* trust myself to check 32x27x32 possible combinations right now.

mc776:
Just did another round of checks, can't find dups of anything at all, I think this is safe to close.

mc776:
I just realized while looking at #1125:

> --FreeDM--
> Title: (same as Phase 2)
> Intermission: (same as Phase 2)
> Story: (same as Phase 2)

PR #349, opened 2017-01-08
Velvetic midi

Voros2:
KevinHEZ has generously donated his Velvetic MIDI Pack, so here's some replacements.

The FreeDM ones provided felt more fast paced than the current one, hence these two tracks were assigned to FreeDM.

C1M9 and MAP18's MIDIs get cut off in Chocolate Doom, so those two should be replaced.

The rest are mainly for replacing unsuitable tracks.


Voros2:
It can be considered to partly address https://github.com/freedoom/freedoom/issues/10

PR #355, opened 2017-01-17
maps: vanilla Phase 1 and 2 maps

Voros2:
On behalf of Catoptromancy, and by his request, I would like these vanillafied maps get merged into Freedoom.
C2M4 is by Protox.
Map21, 27 and 29 were done by Amarande.
Map13, 14, 17, 19 and 24 are Jayextee's work. A few were subject to minor edits by Cato.
The rest are all Cato's work.

I hope these get merged soon.


Voros2:
Please note map29 is actually limit-removing, due to some drawsegs overflows. Nothing fatal.

Voros2:
Map14 is a filler map and Jon requested if his maps would be removed rather than converted to vanilla. 
Map31 is a small castle and not vanilla compatible due to segfault.

Both maps made by Jayextee

Voros2:
Map13 and map19 are both huge and confusing maps. They need to be replaced with maps that don't require the player to know past knowledge of the map in order to complete it.

Voros2:
Map17 has unfavourable gameplay according to players (and design, it seems). While it's fun to jump of all those buildings int there, I guess the disadvantages outweigh the advantages. Thankfully, the replacement has a "capitol" vibe and is more interesting too. Map24 clashes with the rest of the maps in a bad way, by being too good. It's something you would see in a custom PWAD map, rather than an IWAD.

teamgenten:
I like map17, the design is actually pretty cool. The mob and weap placement is bad.

Voros2:
This is now ready to be merged.

PR #358, opened 2017-01-19
patches: TNT patches

Voros2:
Three patches made by @uhbooh a long time ago, that weren't merged.

The font for DISASTER suits well, and makes good use of that old font. Hopefully it won't be too much trouble to merge these patches over the placeholder ones.


chungy:
Using the newer font would be better. Please fix.

Voros2:
Another reason why the old font was used:
 
>I used the old font because the original >patch had the words drawn in blood. The >old font resembles blood more than the >new one so I used it.

chungy:
Rather not mix in the old font at all. Can it be drawn so it looks hand-painted?

AXDOOMER:
@chungy What do you think about these? A applied a "slur" filter. Slurring produces an effect resembling melting the image downwards (like if it was dripping). The font is more or less not recognizable, but I had to play with the options so such low resolution text would still be readable. 
disaster2
disaster3

chungy:
better, but I still think it's a bit obvious. I had merged these patches and made my own disaster thing by hand-drawing something in GIMP: https://github.com/freedoom/freedoom/commit/ffc71c8362990603af004255eee6ea2088320864

AXDOOMER:
What's wrong with the old font? I think "disaster area" drawn in Paint with the paint brush is even worse.

AXDOOMER:
Is this acceptable? (hard to read, but at least the old font is not recognizable)

disaster

chungy:
I honestly think faffing about with this one is just trying to paint lipstick on a pig.

I can concede that my own replacement isn't all that great (solid color, yay...), but I'd rather see it improved (or a new texture in similar spirit) than trying to shoe-horn this one in. Mind you, DISASTER isn't used in a single Freedoom level (test_level.wad not counting), and only once in tnt.wad MAP27. It's a fairly low-key and seldom-used one.

AXDOOMER:
Yeah right, we'll spend more time drawing a new one the first time it's going to be used in a level. It's probably never used in any PWAD.

Issue #245, opened 2016-02-21
A particular sector in Phase 1, E1M1 in the toxin pit isn't a damaging floor.


TrustyGun:
This one, exactly. This is the first waterfall, but it is the same for the other waterfall. You can walk on top of it after the toxin pit rises.
![screenshot_doom_20160221_072758](https://cloud.githubusercontent.com/assets/7079721/13202551/ccb128b4-d86c-11e5-9611-c7505d0bed88.png)

CraftedPixel:
hmm someone must have forgot lol

Issue #241, opened 2016-02-07
VISPLANE OVERFLOW in 1st level


CraftedPixel:
Well it is sort of impossible to look outside without a visplane overflow error.
Can someone help :D

fragglet:
Which level do you mean? C1M1 or MAP01?

chungy:
and which version?

Freedoom is not yet vanilla-compatible, neither in the latest release nor the development tree. If you're looking for help in converting levels, you might try Doomworld's WADs and Editing forum.

CraftedPixel:
I just learned how to make maps, but could not even get doors to work :D
I am talking about the episode 1 level 1 when looking outside will confuse the engine and crash
The editor I am using detected many errors in the level
Also trying to figure out how to open up a wad file to get all the sprites for editing
P.S.
And how to get doors working

Issue #227, opened 2016-01-06
Fix C1M1 Rocket Room


SuperSomariDX:
![stuckarea2](https://cloud.githubusercontent.com/assets/6821693/12148175/0f0b1c5e-b46b-11e5-8bc7-70a2a8e7cecf.png)

The Deathmatch rocket launcher room has a place to fall into the nukage on one side. It is the diagonal angel closes to the slimefall room.

There are three possible things that can be done.

A: Leave as as due to it's damaging sector properties
B: Make The corner an impassible wall.
or 
C: (A Cosmetic Change - Minor) Add a small pipe in the area that can represent the nukage going through a machine of some sorts, and big enough to block a fall off.

The small room itself is quite bland honestly, but that's a personal issue.

PR #360, opened 2017-01-23
map: new C1M1

Voros2:
I understand the historical value of C1M1. Its a great map.
Jon said once that he would like you guys to consider removing his maps rather than converting them(source: https://www.doomworld.com/vb/post/1689762 ). And recently, YukiHerz updated a map he made for Freedoom. 

This map introduces to the player many key elements of the game eg acid is bad, using lifts and switches, different light levels, the power of enemies, the key system etc.

Addresses https://github.com/freedoom/freedoom/issues/245 https://github.com/freedoom/freedoom/issues/241 https://github.com/freedoom/freedoom/issues/227 

It has a wonderful outdoor area and utilises the new Aquatex textures nicely :) and its completely vanilla.

I sincerely hope you'll merge this into Freedoom.


fragglet:
Hey @Voros2 and @chungy, I haven't been paying much attention to Freedoom lately but I just checked out the new release and I have to say that I love this new E1M1. I particularly like how it makes use of the Aquatex textures that @jmickle66666666 made and I think it provides a really compelling vision for the rest of the project.

The visual "cohesion" (I don't know what's a good word to describe it) is really strong in this level. That is, the monsters, the textures and the level all seem to be in harmony and to have a common visual theme (subtle brown and greys combined with bright primary colours). This is something we should really be aiming for in the project in general since having a common theme gives it a high quality feel - it looks less like a mixture of different conflicting resources. It's the kind of result I've been hoping for in the past when I've talked about having stronger art direction. Other people also seem to have noticed this [in the Doomworld release thread](https://www.doomworld.com/vb/doomworld-news/92937-freedoom-v0-11-emerges/).

I suspect the use of Aquatex is a big part of this, though it also comes down to level design and how the textures are used. But assuming that I'm right, it would be a very good idea to promote the use of Aquatex throughout the IWAD levels. I know there's already work being done to make levels vanilla compatible but often reskinning isn't a huge amount of work (often not much more than texture search and replace) - maybe it can be done. Honestly I think this is a big deal - we should be holding this level up and telling people - **this** is Freedoom's visual style. Let's make all our levels like this.

Voros2:
It can be done, and once I get a computer, that's what I'll do. That's always been my plan: to edit the maps so that they use Aquatex along with stock textures in a 50:50 ratio, depending on the map.

PR #362, opened 2017-01-29
New reworked c2m3.

Proluiz:
This is the final revision and include new edits: (from p4_c1)
Dm Exclusive gameplay.
Very slight skill tweek.
More ammo balance.


Voros2:
Not to mention that it's fully vanilla now.

PR #364, opened 2017-01-30
lumps: updates

Voros2:
There's been a recent influx of contributors, and they haven't been credited yet. Also, made minor cleanups.

The dehacked files now contain intermission screens that should be friendly to 4:3 monitors, level name changes to fit the new levels, if not already, minor stuff, and fixing a typing error in buildcfg.txt.
Plus some new demos for, hopefully, the imminent release.


chungy:
Removed the demos as I implied on the forums.

PR #365, opened 2017-02-07
buildcfg: FREEDOOM lump for Phase 1 and 2 only

Voros2:
FreeDM contains both FREEDM and FREEDOOM lumps. These two lumps are identical to one another, so make FREEDM be the only lump present in FreeDM.


chungy:
The contents aren't really the important part. The existence of the FREEDOOM lump is good for detecting that it is a Freedoom IWAD.

Voros2:
Then is the FREEDM lump needed?

chungy:
Neither are strictly needed for the IWADs to work, but they are useful for source ports' game detection code.

Voros2:
Isn't that the purpose of the FREEDM lump in FreeDM?

chungy:
Partially, for FreeDM specifically. There's no harm in having both lumps in the IWAD.

Voros2:
There's no harm in having __EUREKA lumps either. The FREEDM lump is supposed to be used for recognising FreeDM in source ports, and apparently the FREEDOOM lump too. Why have both, when both serve the same purpose?

chungy:
__EUREKA lumps have no utility function, these ones do.

Removing them isn't happening.

Issue #373, opened 2017-02-16
Remove newline escape sequences from C6TEXT


Voros2:
There are two newline escape sequences at the end of C6TEXT, so it continues displaying text, such a part of the obituary message for the stealth technospider. I incorporated the fix in the C6TEXT fix pull request, but it seems it wasn't added.

Since the new v0.11 has this bug, maybe a small v0.11.1 would be worthwhile?

chungy:
Thanks, this is fixed. Indeed, if/when we do a 0.11.1, this would be one of the fixes.

PR #374, opened 2017-02-16
dehacked: Remove \n\n\ 

Voros2:
This fixes a crucial part of C6TEXT. The \n\n\ causes the engine to continue looking looking for text to display onscreen, even though the required text has already been displayed. This makes it look as if it's broken.

Also included the new par time for map26.


Voros2:
Addresses https://github.com/freedoom/freedoom/issues/373

PR #384, opened 2017-02-21
changed pestmeister blood frames

luckypunk:
Don't know if this will work. This should just be 6 changed gifs. I still don't quite get git.

If it works, I hope these look slightly better.
[bos2smno0.zip](https://github.com/freedoom/freedoom/files/791788/bos2smno0.zip)


chungy:
Thanks. You're doing fine, but use `git commit --amend` next time to make changes to your commit. Don't need to make a new one :)

luckypunk:
> use git commit --amend

Thanks! That will probably help. I'll try to remember this for next time.

PR #395, opened 2017-02-25
Sprites: change sprites by cwolf and raymoohawk

luckypunk:
https://www.doomworld.com/vb/post/1720049


Voros2:
Better. Best to add a small description alongside the commit too, as it goes into the history.

Voros2:
Well, better than nothing. Please add a message to your commits.

luckypunk:
@Voros2 

> Well, better than nothing. Please add a message to your commits.

Too late now, but for the future: How would that message look, for example?

Voros2:
https://github.com/freedoom/freedoom/commit/fd216d0234c3c4b11ccb8bdbdfb94b57173ae49d As a reference compared to https://github.com/luckypunk/freedoom/commit/9f2955e78a9e1a0611b5c11c39837934b94f3cb6 

But it's too late. Looks like chungy will have to do it himself.

luckypunk:
Thanks.

I'm supposed to split that up into 14 individual commits and then write a two-liner comment for each?

That seems a bit unreasonable.

I should also mention that I'm submitting these for raymoohawk and personally I think that most of these are worse than what's already in there. I wouldn't even know how to justify these.

Voros2:
IMO, these sprites are making Freedoom more cohesive as its own game, not just a bunch of unrelated resources like before. Same artist, same quality, good consistency. 
But this isn't the place to argue over that. Here, you just want these to be added to the repo.

PR #397, opened 2017-02-25
levels: move map06 to map18, add new map06

Voros2:
The current map06 is too difficult for such an early slot, so it has been moved to map18. For the new map06, it uses map04 of Nex Credo, made by Jayexetee. Level is vanilla compatible and tested.


chungy:
Please please break your lines at 72 characters next time, but thanks.

Voros2:
Are you talking about the commit description?

PR #398, opened 2017-02-26
patches: new Plutonia textures

Talon1024:
These new textures are based on various public domain photographs. Look in the commit messages for the specific photos used.


chungy:
Thanks. Merged, but please break your lines at 72 characters next time

PR #403, opened 2017-02-28
sprites: various by cwolf and raymoohawk

luckypunk:
https://www.doomworld.com/vb/post/1720049


luckypunk:
don't commit these yet. probably screwed up buildcfg.txt

luckypunk:
everything should be fine now

PR #404, opened 2017-02-28
sprites: cand & cbra by cwolf and raymoohawk

luckypunk:
https://www.doomworld.com/vb/post/1721082


luckypunk:
don't commit these yet. probably screwed up buildcfg.txt

PR #406, opened 2017-03-03
patches: new Plutonia textures

Talon1024:
Check the commit messages for what these new textures are based on.

Also, check out [my post on ZDoom Forums](https://forum.zdoom.org/viewtopic.php?f=37&p=982583&sid=6c00345906f1121a43900def9b193192#p982583) containing many of my source files, and true colour versions of these textures, as well as some unused textures.


AXDOOMER:
Those are very nice! I'll use them on my map.

Voros2:
One of the MC patches cause a lava fall animation to break. I'm not sure which one as of yet, but the conflicting patch should be removed as the lava fall animation is one of those common animated textures used.

Talon1024:
Yeah. Plutonia's texture definitions hijack DBRAIN1 and DBRAIN4 for the start and end of the MC animation respectively, and they also hijack FIREBLU1 and FIREBLU2 for the start and end of the AROCK animation.

Doom 2's texture definitions use DBRAIN1-DBRAIN4 as the lava animation for the final boss' weak point (using RWDM11A-D patches).

Voros2:
Well the change is noticeable in a very bad way, as the lava fall animation is broken now. Probably best to revert back to the old MC patches.

Talon1024:
The DBRAIN animation can be fixed by using the Doom 2 texture definitions for it, as well as stretching its height to 128. Also, there's one MC patch I didn't replace, and that is MC9, since it is not used in Plutonia's MC animation.

Maybe Plutonia compatibility could be achieved by getting rid of the Plutonia texture definitions in Phase 1/2, and creating a new Freedoom IWAD or PWAD with the Plutonia texture definitions.

Voros2:
The last thing the maintainers want is to build more IWADs, last I remember. Not to mention Final Doom IWADs were basically Doom 2 with some new textures and new levels. I'm sure they chose to use the Final Doom definitions for a reason and I don't think you have a choice other than reverting back to the old MC textures.

PR #410, opened 2017-03-09
musics: new FreeDM tracks

Voros2:
Made by jupiter_ex. Both tracks in FreeDM
are unsuitable because of the slow tempo
present and gets cut off in Chocolate Doom.


chungy:
Thanks

PR #413, opened 2017-03-11
titlepic: use image instead of text, change footer

Voros2:
Rather than using the bland fonts for the titlepic, instead generate 
images using Freedoom's font and textgen. These images get overlayed
where the "Phase 1" and "Phase 2" white-lettered words used to be.

Also, replace the website URL on the titlepic with a copyright notice,
as the website URL is already mentioned on the help screen (where
it should be). And removed the "Version: " part of the version footer
text because it's not necessary. Finally, make footer text orange 
because it stands out well in a nice way :)

Preview: http://imgur.com/a/9PtzQ

### DIFF
--- a/graphics/titlepic/Makefile
+++ b/graphics/titlepic/Makefile
@@ -6,12 +6,18 @@ m_doom.gif: ../m_doom.gif
 m_dm.gif: ../m_dm.gif
 	convert -transparent '#00ffff' ../m_dm.gif m_dm.gif
 
-fd1title.gif: titlepic.gif m_doom.gif
-	./create_caption titlepic.gif m_doom.gif "Phase 1" $@
+t_phase1.gif: ../t_phase1.gif
+	convert -transparent '#00ffff' ../t_phase1.gif t_phase1.gif
+	
+t_phase2.gif: ../t_phase2.gif
+	convert -transparent '#00ffff' ../t_phase2.gif t_phase2.gif
+
+fd1title.gif: titlepic.gif m_doom.gif t_phase1.gif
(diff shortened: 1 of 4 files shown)


graphics/titlepic/Makefile
```diff
@@ -6,20 +6,27 @@ m_doom.gif: ../m_doom.gif
...
 
 fdmtitle.gif: ../captainw/fdmtitle.gif
 	./create_caption ../captainw/fdmtitle.gif $@
 	cp $@ ../
 
 clean:
-	rm -f m_dm.gif m_doom.gif fd1title.gif fd2title.gif fdmtitle.gif \
-	      ../fd1title.gif ../fd2title.gif ../fdmtitle.gif
+	rm -f m_dm.gif m_doom.gif fd1title.gif fd2title.gif fdmtitle.gif      \
+	      ../fd1title.gif ../fd2title.gif ../fdmtitle.gif ../t_phase1.gif \
```
fragglet: The 'clean' rule should only delete things that are generated by this makefile. It looks like you've added t_phase1.gif and t_phase2.gif to this list and they're generated by textgen in the other directory. Can you fix?

Voros2:
But the t_phase images aren't supposed to be as lumps in the iwad. Doesn't that mean I have to to clean it? I'll fix it anyway.

Voros2:
Should be fixed now.

chungy:
The clean rules in the Makefiles are only about cleaning up built files after `make clean`, not what does or doesn't go in the WAD, that's determined instead by buildcfg.txt :)

PR #414, opened 2017-03-11
Titlepic: use more discrete breathing orifices

Ferk:
This changes the helmet of the Freedoom guy in the titlepic to be more similar to the how the sprite looks, by making the breathing orifices more subtle and less muzzle-like.


AXDOOMER:
Definitively better. It looks less like Darth Vader's helmet and a lot more modern.

PR #420, opened 2017-03-16
sprites: change pestmeister death animation

luckypunk:
still not good, but better than before


luckypunk:
Thank you! That's exactly what I was going for.

Issue #419, opened 2017-03-16
Increase height of all spheres 


Voros2:
The spheres are essential power ups that should be visible easily. The recent sprite update lowered the offsets to make it look like the spheres are on the ground. Given the lack of looking up and down, it was a bit hard to see. By making the spheres appear at the same level as the player camera (32), the spheres are immediately visible, even from a large distance. Please implement this.

PR #423, opened 2017-03-17
buildcfg: increase sphere heights

Voros2:
As mentioned in https://github.com/freedoom/freedoom/issues/419 
this raises the height of the four spheres so that they are more visible
and recognisable as power ups.


raymoohawk:
i may have misunderstood, but wont this overdraw the sprites on some maps? or clip them in opengl?

Voros2:
I get what you mean, and no it won't. The height was increased by a mere 7 pixels. Imagine the current sprites in game and raised by a few pixels. Nothing big to cause clipping of any sort.
Doom doesn't have freelook (in vanilla that is), so this was needed for these spheres. Doom's soulsphere has a similar height too, and other spheres too.

PR #430, opened 2017-04-03
wad-diff: Added wad-diff and other utilities to diff WAD files

selliott512:
This is my attempt at adding utilities that can be used to visually diff WAD files. See the "I've created a utility that displays differences in WAD revisions visually" thread for details.


selliott512:
I accidentally included map07.wad in this pull request. I'll create another one without map07.wad, but otherwise the same.

PR #432, opened 2017-04-03
wad-diff: Added wad-diff and other utilities to diff WAD files

selliott512:
This is my attempt at adding utilities that can be used to visually diff WAD files. See the "I've created a utility that displays differences in WAD revisions visually" thread for details. This is the same as PR 430 except it no longer incorrectly includes a change to map07.wad.


NickZ:
This looks really nifty, and it looks like it could be really handy, especially for the upcoming level changes. However, I cannot get yadex to work on any of my computers (which isn't surprising, given how dead the project is and it's not currently packaged in debian). I prefer the current method that you have going with the independent app, using OMGIFOL instead. What I'd recommend is that you add the diffing functionality to the independent app.

selliott512:
That makes sense. I've thought about submitting a new patch with wad2image (the independent app), but I wasn't sure there was sufficient interest. Maybe I'll try it since you've expressed interest.

The approach for this pull request is to diff two screen shots using red (removals) and blue (adds) to highlight changes. As it is wad2image illustrates changes with animated GIFs. Which do you think would be preferred, the color approach, or animate GIF? Maybe I could do both with a variable or build target to indicate which.

Also, since wad2image is GPL version 2 or later, which I did because it includes some Yadex files, maybe it would be best to not check it into Freedoom. Then again if it's just used for the build maybe the license isn't as concerning. In any case if it is best to not check a copy of wad2image into Freedom it's still pretty easy to install. The build could then use the externally installed wad2image.

NickZ:
> The approach for this pull request is to diff two screen shots using red (removals) and blue (adds) to highlight changes. As it is wad2image illustrates changes with animated GIFs. Which do you think would be preferred, the color approach, or animate GIF? Maybe I could do both with a variable or build target to indicate which.

Why not both? :B

If I had to choose one or the other, I would probably go with the red and blue static image as the default.

>Also, since wad2image is GPL version 2 or later, which I did because it includes some Yadex files, maybe it would be best to not check it into Freedoom. Then again if it's just used for the build maybe the license isn't as concerning. In any case if it is best to not check a copy of wad2image into Freedom it's still pretty easy to install. The build could then use the externally installed wad2image.

Yes, this is my thinking as well, especially since you're now relying on a bunch of third-party libraries.

selliott512:
I couldn't help but notice that there's only been one commit to Freedoom since March, which seems odd given the amount of activity on the forum. Do you think if I did this there would be a reasonable chance of it being committed?

Voros2:
If it's an improvement, why not? Although I wouldn't say that you're PR isn't being merged because it's not good enough. I'm sure the maintainers have their reasons for that 4 month hiatus.

NickZ:
@selliott512 If it's good enough and @chungy says it's ok, I will ensure that it gets committed.

scripts/wad-diff
```diff
@@ -0,0 +1,168 @@
...
+# The set of WAD files considered can be further reduced by the optional
+# "wad-regex" argument. For example, if a diff would be produced for both map05
+# and map07 but you only want to see the diff for map05 you could pass "05" for
+# wad-regex.
+
+# Globals
+
+dname="${0%/*}"
+bname="${0##*/}"
+tmp_dir="/tmp/$bname.$$"
```
chungy: Use `mktemp` for temporary directory creation.

scripts/image-diff
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
```
chungy: Please add a license identifier. Examples can be found in the other scripts, but it is generally just `SPDX-License-Identifier: BSD-3-Clause` to declare the file as using the Freedoom license.

scripts/wad-diff
```diff
@@ -0,0 +1,168 @@
+#!/bin/bash
```
chungy: If it is at all possible, change to use `/bin/sh` (and use shellcheck to verify syntax). I won't consider it bad to require bash, however.

chungy: License identifier required here.

scripts/wad-diff
```diff
@@ -0,0 +1,168 @@
...
+fi
+if [[ $# -ge 2 ]]
+then
+    wad_regex="$2"
+else
+    # Anything by default.
+    wad_regex=".*"
+fi
+
+if ! mkdir -m 700 "$tmp_dir"
```
chungy: Replace with `mktemp` as noted above :)

scripts/wad-to-image
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
chungy: License identifier required here, change to `/bin/sh` if possible.

scripts/wad-to-image
```diff
@@ -0,0 +1,153 @@
...
+
+# wad-to-image - Convert a WAD file to an image file.
+# by Steven Elliott <selliott512@gmail.com>
+#
+# The WAD file is displayed in Yadex, which must be installed, and then
+# "import" is used to capture the image in the Yadex window.
```
chungy: Yadex is an unmaintained decrepid program that doesn't even compile for me ;)

Perhaps Eureka (Yadex derivative) or omgifol can do the job better.

scripts/wad-to-image
```diff
@@ -0,0 +1,153 @@
...
+    exit 1
+fi
+
+unset yadex_win_id
+for ((i=1; i <= 50; i++))
+do
+    sleep $small_delay
+
+    # The "head" is to get the most recent one.
+    yadex_win_id=$(xwininfo -tree -root | grep -w Yadex | head -n 1 | \
```
chungy: This won't work on Wayland systems (or Windows or Mac OS X ...). Omgifol to generate map images may be better than trying to capture a graphical window.

selliott512:
Keep in mind that a lot of this is likely to change. As hinted at by my discussion with NickZ the Yadex dependency is going away in favor of my new wad2image script, which is OMGIFOL based. chungy, unfortunately this means much of your code review feedback will no longer applicable. But I do appreciate many of the things you had to say. I'll submit a new pull request when the time comes since it will be sufficiently different. The new approach will minimize changes to Freedoom. Instead the functionality will be in wad2image as much as possible.

Makefile
```diff
@@ -100,7 +100,11 @@ else
 	JSON=$(JSON) VERSION=$(VERSION) scripts/makejson
 endif
 
-clean:
+# Clean files produced by "wad-diff".
+wad-diff-clean:
+	rm -f wad-images/*.png
```
selliott512: I was just trying to make my changes as isolated as possible, but I'd be happy to add to the main target.

scripts/image-diff
selliott512: This functionality will probably be incorporated into the `wad2image`, but I'l make sure to include that to any script I add to Freedoom.

scripts/image-diff
```diff
@@ -0,0 +1,81 @@
...
+# Functions
+
+def open_bw_image(name):
+    image = Image.open(name)
+    return image.convert("L")
+
+def is_on(pixel):
+    return pixel >= threshold
+
+from PIL import Image
```
selliott512: Good catch. I'm not sure how it ended up here. I think I was experimenting with having a method that made the import conditional by catching an exception if it failed, but that's not what I'm doing now. However, this file is going away.

scripts/image-diff
```diff
@@ -0,0 +1,81 @@
...
+
+if len(sys.argv) != 4:
+    print >> sys.stderr, "Usage: image-diff image-1 image-2 diff-out"
```
selliott512: Indeed that does not work on Python 3. I'm trying to recall what my intent was for supporting different Python versions for this script. In any case, `wad2image`, which will replace this, was tested on Python 2 and 3.

chungy:
No problem, half expected it when I took so long :)

Nick mentioned on IRC that you may be actively developing this script outside. Does it make sense being inside this repository, or can you generalize it and we could simply have an external dependency?

Makefile
chungy: Understood, but no reason to be intimated by changing what is here. [Be bold!](https://en.wikipedia.org/wiki/Wikipedia:Be_bold)

scripts/wad-diff
selliott512: How did I not know about shellcheck? Thanks for that. I'll try it on my future shell scripts. As to `#!/bin/sh` vs. `#!/bin/bash` - I agree that it's helpful to make the script to work on as many systems as possible, and they all have `/bin/sh`.  However, pure Bourne shell can have unpleasant syntax. Also, `/bin/sh` varies more system to system than `/bin/bash`. On Fedora it's often a symlink to `/bin/bash`. I believe on Ubuntu it can be a stripped down shell depending on the install. Also, this particular script will change significantly and become a part of wad2image.

scripts/wad-diff
selliott512: I'm aware of `mktemp`, but I use the approach above because the I like the having the PID in the name. If the PID is logged somewhere, or seen from `ps`, it makes it easy to relate the running process to it's temporary files. However, there's something to be said for following standard secure programming practices so that others can make sure the program is secure even when you've convinced yourself that it's secure. So for this script, which will probably be changed a bit, renamed, and moved to `wad2image`, but will still exist, I'll do as you suggest. But I may revert back to my `tmp_dir` ways in future scripts I write. I'll see how it goes :-)

scripts/wad-to-image
selliott512: The Yadex dependency is going away. Don't worry about the references to Yadex in `wad2image`. That's just some Yadex text files, and not Yadex itself.

scripts/wad-to-image
selliott512: I was hoping it would work on Linux (where I tested it), or the Cygwin environment you recommend for Windows with an X Server and related tools installed (which I have not tested), but I'll take your work for it that it won't work. It's going away in any case.

selliott512:
The new approach I'm leaning toward will be to have all functionality within `wad2image` and for Freedoom only modify Make targets. From your last comment I'm guessing you agree.

chungy:
Sounds like it. I'll close this PR, then.

selliott512:
Just letting you guys know that I haven't forgotten about this. I've been on vacation, but you'll probably see another pull request in a week or so.

selliott512:
I still haven't forgotten about this. Just letting you guys know assuming anyone is still listening. Maybe this weekend if all goes well.

NickZ:
Yup, I saw that you had started making the changes on your fork. It's gonna
be sweeeet.

A request: can you make this diff against the current HEAD of master by
default? I have an idea for this with Travis-CI: When autobuilding a pull
request, it will execute wad-diff and will upload the artifacts to an s3
bucket and an automatically post them in the PR thread.

selliott512:
I'm glad you like it. I was curious if anyone was watching. :-)

I'm not sure what you mean by "default", but when the time comes I'll rebase the Freedoom portion against latest HEAD. I'll probably squash it into one commit. Are you concerned that I have it in a branch I created? I think the merge will work smoothly in spite of that as long as I update it appropriately.

NickZ:
Oh, no, I was talking about the wad-diff command itself. That's the default
behavior I want, so that when people submit a PR with level changes, we can
easily get the png images.

selliott512:
Oh, I see. As it is I plan on adding at least two targets to the Makefile that include "wad-image" and "wad-image-diff" which will produce images in a particular subdirectory. "wad-image-diff" depends on git to determine the changes. But I'll work with you to come up with something helpful.

PR #435, opened 2017-04-13
textures.cfg: fix DBRAIN animation.

Talon1024:
Set DBRAIN1's size to 64x32, and remove MC patches from it.

This should address the issues brought to attention in #406 and [this thread on Doomworld](https://www.doomworld.com/forum/topic/93689-the-lava-flow-texture-is-bugged/).

Plutonia compatibility can be achieved by creating a separate PWAD with Plutonia's texture definitions. This is why the MC and AROCK animations work when playing Freedoom with Plutonia 2 or PRCP. Plutonia 2 and PRCP both include the Plutonia texture definitions, in addition to a few new ones.

### DIFF
--- a/lumps/textures/textures.cfg
+++ b/lumps/textures/textures.cfg
@@ -1200,28 +1200,8 @@ CRACKLE2        64      128
 *   RW44_2      0       0
 CRACKLE4        64      128
 *   RW44_4      0       0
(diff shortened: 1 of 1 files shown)


NickZ:
I've tested this, and I can confirm that this fixes the issue without creating any new ones.

Issue #436, opened 2017-04-15
Please remove the new annoying death sounds


AXDOOMER:
It's really annoying in deathmatch where it's played repetitively. I fail to understand how they are an improvement on the previous death sounds. 

commit: https://github.com/freedoom/freedoom/commit/2246f2d5526702d9cd79d1b6e1c56df0712973c3

rjy:
>I fail to understand how they are an improvement on the previous death sounds

I think there was plenty wrong with the previous death sounds. One was a repeat of one of the zombie deaths, an obvious mistake. The other was the Wilhelm scream, a tired cliché.

AXDOOMER:
Yeah, but now it's even worse. It discourages me from deathmatching in FreeDM.

rjy:
Alexandre-Xavier Labonté-Lamoureux wrote:
>Yeah, but now it's even worse. It discourages me from deathmatching in FreeDM. 

I wouldn't know. If only Odamex were active when I am awake...

It's a pity Freedoom can't use D_EPain (from DoomEd), it's hilarious.
https://github.com/fragglet/DoomEd/blob/master/D_EPain.snd

AXDOOMER:
The previous sound sounded more like somebody who was dying. Who screams at the top of his lungs when he dies? 

https://www.youtube.com/watch?v=nAdniWncWu4

NickZ:
Fixed in #437

PR #437, opened 2017-04-15
Tweak speed/pitch of player death sounds

Ferk:
This is in response to #436
I'd rather the sounds are improved than removed, so I've shortened and tweaked them to make them less annoying. 

In the case of ``dspdiehi`` I added some short fade-out, removing part of the end, because it was getting distorted when sped up.

I don't have good recording equipment to completely remake them / make my own, so they are actually based on the same sounds, but I think this is a significant improvement.

Check it out and let me know if you think it should be edited further.


AXDOOMER:
I removed the loud part from the first one and made the second one less noisy: https://ufile.io/d864f

I used your modified sounds.

Voros2:
Are they 8bit, mono and 11025hz? I get noise after compiling with Deutex?

AXDOOMER:
Maybe that's why I felt there was some noise in Ferk's sounds. I couldn't really improve them so I'm not going to redo the transformations that I did again. The sounds should be replaced with something that matches better with the FreedomGuy current voice.

Ferk:
Ow, sorry. I've converted, built and tested the WAD in Retro Doom. When playing the WAV file in VLC there's a bit of noise (as was in the original) but there's no noticeable noise in-game.

To minimize the noise in the second sound (I mean ``dspdiehi``), I've just used some fading to shorten it without really affecting its speed/pitch.

Maybe it would be better to get new sounds from the ground up. But in the meantime I'd rather see them tweaked than removed, which was what #436 suggested.

AXDOOMER:
@chungy The latest Freedoom didn't build: http://freedoom.soulsphere.org/

Voros2:
Deutex was updated:D but the updates Deutex isn't in Freedoom's server.

You'll be happy to hear the tall sprite error has been fixed (mostly) and PNG support added.

chungy:
Yup, we know. Build Freedoom yourself or wait for the server to be updated.

AXDOOMER:
> You'll be happy to hear the tall sprite error has been fixed (mostly) 

Oh yes indeed. :smile:  Good job!

PR #456, opened 2017-07-20
musics: map01.mid

Voros2:
This was originally in c2m5, but the author requested if it could be 
moved elsewhere.


NickZ:
but I like the old map01 :(

Voros2:
I liked a number of other stuff too, but progress. BTW, the old map01 is in c2m5 now, so it's not gone :) (although now you have map01 and c2m5 playing the same track) the short length of this midi is suitable for map01. Not to mention that this midi was made for map01 and not c2m5.

I originally placed it in c2m5 in order to minimise the sharing of one track over two or more maps, as much as possible.

Issue #470, opened 2017-08-03
Including Hambourgeois textures


Voros2:
If you would look into this thread: https://www.doomworld.com/forum/topic/88143-the-hambourgeois-content-mill/

You'll find that he made some patches and flats that would make for nice additions to Freedoom. More specifically, an attempt to fix some problematic textures in Freedoom.

Consider replacing patches/flats with some/all of his textures?

chungy:
Looks good to me.

Issue #171, opened 2015-11-28
Textures that are too close to the originals


NickZ:
RW32_2 and RW15_1 are too close to the originals and should be replaced.

https://github.com/freedoom/freedoom/blob/master/patches/espi/rw32_2.gif
https://github.com/freedoom/freedoom/blob/master/patches/espi/rw15_1.gif

Comparison (Doom on the left, FreeDoom on the right)
![comparison](http://i.imgur.com/ml3YXVr.png)
![comparison2](http://i.imgur.com/7owrX57.png)

NickZ:
fixed in #471

PR #471, opened 2017-08-04
textures: new stuff by Hambourgeois

Talon1024:
New textures by Hambourgeois to replace some ugly and/or problematic
textures. These new textures are from this Doomworld forum thread:
https://www.doomworld.com/forum/topic/88143-the-hambourgeois-content-mill/

Should address #470


NickZ:
Thanks @Talon1024! That takes care of #171 as well. 

BTW, github has a nifty feature where if you include the magic word "Fixes" in front of an issue number, like "Fixes #470", it will automatically close that issue when the PR is merged.

PR #472, opened 2017-08-06
scripts: fix-gfx-offsets

NickZ:
This is a script that will take in wildcards (e.g. "vile* pisga*") and update
buildcfg.txt with the offsets contained in the specified png files.


chungy:
I don't think processing wildcards in the script is a good idea or necessary. That'd just require escaping them in the shell so the script can process that specially... taking in a plain list of files seems like a better idea IMO.

Voros2:
Why is this needed? I'm a bit confused.

NickZ:
@Blastfrog specifically asked for support for this.

Blastfrog:
This would definitely be a useful tool for mass updating sprite offsets easily and painlessly. The intent is to use a tool like SLADE to align the sprite, export it as PNG, and then use this tool to update the offsets in buildcfg.

I agree with chungy, though, a plain list of files is preferable to wildcards. Maybe have an option for both?

Ferk:
In Unix systems, the wildcards will automatically be expanded by the shell, so as long as the script accepts any space-separated list of files (quoted/escaped if there are spaces in the filename), it should work

PR #480, opened 2017-08-24
Added targets, variables and help for wad2image integration.

selliott512:
This is an updated and significantly changed attempt to integrate wad2image into the Freedoom build. Some highlights since the last pull request I submitted (#432):

- Moved most functionality into wad2image. For example, my previous pull request added scripts to the "scripts" directory. The new approach is for wad2image to be copied to or symlinked to the "scripts" directory, but otherwise there are no new scripts.
- In addition to animated GIFs and the black or white images seen in my [Freedoom forum post](https://www.doomworld.com/forum/topic/93601-ive-created-a-utility-that-displays-differences-in-wad-revisions-visually/) color coded difference images can be generated on top of reduced color saturation images. I've added an example to [wad2image's home page](http://selliott.org/utilities/wad2image).
- More flexible make targets to generate images from the build.
- Various wad2image bug fixes.

The latest version of wad2image (0.9.3) is needed. I think it makes sense to refer to it with a [latest link](http://selliott.org/sites/default/files/utilities/wad2image/wad2image-latest.zip).

I'm curious how you feel about the size of "Makefile" being increased by 88 lines mostly for help. I can move that to a separate text file, but I figured it might be handy to have a help build target.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /CREDITS.txt
 lumps/freedoom.lmp
 lumps/freedm.lmp
+/scripts/wad2image
 /wads/
 wad*.txt
 *.bak
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 lumps/freedoom.lmp
 lumps/freedm.lmp
 /scripts/wad2image
+/wad-images/
 /wads/
 wad*.txt
 *.bak
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ else
 	JSON=$(JSON) VERSION=$(VERSION) scripts/makejson
 endif
 
-clean:
+clean: wad-image-clean
 	rm -f	*.html deutex.log $(OBJS) \
 		./COPYING.txt ./CREDITS.txt \
 		./wadinfo.txt ./wadinfo_phase1.txt \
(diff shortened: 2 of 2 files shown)


selliott512:
NickZ,

I'm glad you like it. Thanks for the feedback. I think it should be possible to do all of that. I'll add another commit to this pull request soon. Some comments on your comments:

I think I can do the $PATH thing. I think I'll search for "wad2image" in the path and strip off the trailing "bin/wad2image". The result will be wad2image's home/installation directory. "git-wad-diff.sh" is in that home/installation directory, which the build runs. It will be use user's responsibility either set up the $PATH accordingly, or to pass in WI_HOME for the home/installation directory.

Not displaying the images by default is easy. Instead of WI_NO_SHOW=t there will be WI_SHOW=t if you do want to see them.

Sending the images to some other directory is easy. "wad-images" it is.

As to the IWAD - Fortunately wad2image has the concept of a search path (--wad-spath argument) for WADs including the IWAD, so I think I just need to change the search path in the build start with the "wads" directory. In "wads" it will look for "*doom2.wad" so it will end up finding "wads/freedoom2.wad" if it exists. I'm guessing it makes sense to not have a dependency for this - if the IWAD is there it will be used, otherwise not.

selliott512:
Also, I would totally understand if you didn't like having so many "@echo" statements in your Makefile. I could move it to a text file and then just have "wad-image-help" cat out that text file. Just let me know and I'll do that as well.

selliott512:
I think I got everything you requested in the commit I just pushed. Let me know if there's anything else.

selliott512:
I just pushed one more commit. The help was not quite accurate. Also, I wanted it to display the current value of each variable in the help, so now there's actually a reason to have the help in the Makefile. Also, I wanted to check that a few key variables were set to avoid unknown behavior if they're not. If having this many commits in a PR bothers you I can squash them into one.

NickZ:
Thanks for implementing the changes.

For the $PATH stuff, I was thinking more along the lines of wad2image and git-wad-diff should be thought of as standard command-line utilities. As such, they should both be in the "bin" folder, and should not have file extensions, such that all I would have to do is add the bin folder to $PATH. I could see someone wanting to package wad2image as a debian package, and this would make distribution much easier.

selliott512:
In my experience software can either be packaged with system package manager where files end up in /bin, /lib etc., which is what you're describing, or in can be a self contained collection of files suitable for tar, zip, whatever that can be expanded to somewhere like /opt. I opted for the later for wad2image in order to make usable on as many operating systems possible. As for the file extension I did that for Windows users so that it would have a nice Python icon in Windows explorer, so that you can get a short help message by double clicking on it, and so you don't have to explicitly type "python" in cmd.exe. I agree with you as a matter of principle about the extension, because the interface shouldn't be tied to an implementation, but I wanted to make things as easy as possible for users that may not be IT inclined.

I'd be flattered if someone wanted to package wad2image for Debain or some other operating system, but there's a bit of work to do before then. wad2image depends on various files it's shipped with that would have to be moved somewhere. Also, it uses a bundled copy of OMG, which a forked version. So the packager would have to decide if they wanted to package that separately or whatever. In any case I'm pretty easy to find me by searching for "wad2image" if someone wanted to do that. I'd help them.

In order to partially address your concern I've updated wad2image (the latest version is now 0.9.4) so that a symlink to it may be created to it. The symlink can be whatever you want, without the extension, for example:
`cd /usr/local/bin`
`ln -s /opt/wad2image/bin/wad2image.py`
`ln -s wad2image.py w2i`
The above is from the installation instructions I just add to [wad2image's home page](http://selliott.org/utilities/wad2image).

If in the future someone does package wad2image so that it's in the repo I can submit another patch to update Freedoom to use it.

Is that good enough?

NickZ:
Yes, I think that's fine, as long as I can easily create symlinks.

This is ready to merge. Thanks for all your work on this!

selliott512:
Thanks for your feedback. You've helped improve wad2image. I'm excited to have contributed to Freedoom.

One small glitch - the top level directory determination does not work as well as I hoped. I've just submitted PR #488.

chungy:
I don't really like the `WI_HOME` infrastructure here. Seems overly complicated and fragile -- eg, if `wad2image` and `git-wad-diff` are in the `$PATH` but not the "expected" locations by manual install, it just won't work.

It's better to just expect both files to be in the `$PATH` and run them directly as `wad2image` and `git wad-diff`. If really needed, we can use a `$(WAD2IMAGE)` variable to specify the path to an alternate one.

selliott512:
The `WI_HOME` logic is a bit complicated, but did you see PR #488? In it I list the cases that I tested.

`wad2image.py` depends on supporting files in the ZIP file it's shipped with. If for some reason `wad2image.py` is copied out of that directory structure to somewhere in the path it won't work, but it will fail with an error message explaining the problem. I'm not sure how to avoid that with wad2image without having a more complicated or platform specific install process.

If you symlink `wad2image.py` it will work. If you copy it `WAD2IMAGE_HOME` needs to be set.

PR #481, opened 2017-08-28
Fixed map names so that they match the filename.

selliott512:
I mentioned this recently on the [forum](https://www.doomworld.com/forum/topic/95748-level-wads-that-contain-maps-that-dont-match-the-filename/). In order to see that only the map names were changed I find it helpful to add this to .git/config:
`[diff "wad"]
	textconv = hexdump -C`
And for .gitattributes:
`*.wad -text diff=wad`
The differences can then be seen with
`git diff -U1`
Should WAD diffing be configured that way in general?


selliott512:
In the [forum](https://www.doomworld.com/forum/topic/95748-level-wads-that-contain-maps-that-dont-match-the-filename/) I gave a one liner to get the list of affected maps. Here's an improved version that uses DeuTex instead of Yadex:

`(cd levels; for w in c*.wad dm*.wad map*.wad; do l=${w%.wad}; l=${l^^}; l=${l/DM/MAP}; l=${l/C/E}; bad=$(deutex -doom2 ../bootstrap -wadir $w | grep -P '^(MAP\d\d|E\dM\d)' | awk '{print $1}' | grep -v $l); if [[ -n $bad ]]; then echo -e "$w\t$bad"; fi; done)`

Maybe I could make it into a script and put in `scripts`.

chungy:
That would be helpful, otherwise they'll rapidly go out of sync again.

Makefile
```diff
@@ -253,6 +253,15 @@ wad-image-help: wad-image-common
 	@echo "      make wad-image-diff WI_VERBOSE=t WI_GIF=t WI_SHOW=t WI_CMD=animate \
 WI_COMMIT=\"0c004ce~..0c004ce\""
 
+# Test that WAD files have the expected map names.
+test-map-names:
+	tests/test-map-names
```
chungy: This file belongs in the `scripts` directory

selliott512:
Ok, I've added a test script. A few things about the change:

I ended up putting it in a new directory `tests` instead of `scripts` since I thought it might be nice to make a distinction. Any future test scripts can be put in that directory.

I added make targets `test` and `test-map-names`. I figure it will be handy for people to simply type `make test`.

It's a bash script instead of sh. I was going to make it sh, but doing so is hard. I figure anyone that's actually doing a build should have bash installed.

The main build doesn't depend on `test`, but it could if you want.

Makefile
selliott512: I didn't write my explanation comment fast enough, which might of helped. How would you like it to look? Is that a good enough reason to have a new `tests` directory? If not and you still want it in `scripts` do you still want the `test` prefix for the file name and build targets?

tests/test-map-names
```diff
@@ -0,0 +1,73 @@
...
+    expected_name=${expected_name/C/E}
+    if [[ -z $expected_name ]]
+    then
+        echo "Could not determine the expected name for WAD \"$wad\"." 1>&2
+        exit 1
+    fi
+
+    # What the map name actually is as indicated by the name of the first lump
+    # for the map.
+    actual_name=$(deutex -doom2 "$boot_dir" -wadir "$wad" | \
```
fragglet: Probably this is sufficient:

if ! deutex -wadir "$wad" | grep -qi "$expected_name"; then
  ...

selliott512:
Here's what the output of `make test` looks like prior to the maps being fixed (without this change):
```
tests/test-map-names
Filename    Expected    Actual      
--------    --------    ------      
dm31.wad    MAP31       MAP01       
dm32.wad    MAP32       MAP01       
map06.wad   MAP06       MAP04       
map08.wad   MAP08       MAP01       
map13.wad   MAP13       MAP09       
map14.wad   MAP14       MAP03       
map18.wad   MAP18       MAP06       
map19.wad   MAP19       MAP08       
map24.wad   MAP24       MAP06       
map32.wad   MAP32       MAP01       

10 WAD files in levels directory "/home/sle/git/freedoom/levels" have unexpected map names.
Makefile:258: recipe for target 'test-map-names' failed
make: *** [test-map-names] Error 1
```
And here's what `make test` looks like with the maps fixed (with this change):
```
tests/test-map-names
All WAD files in levels directory "/home/sle/git/freedoom/levels" have expected map names.

All tests passed.
```

tests/test-map-names
selliott512: I just added a top level comment showing what the output looks like, which includes the `actual_name`. Admittedly the `actual_name` is obvious to people familiar with this project, but it's a way for people to be told exactly what the map name should be. It's also a way to make sure this script is doing the right thing.  So that's why I'm grabbing it explicitly.

Also, I couldn't get it to work without the `-doom2` option.

selliott512:
Since this has been open for a while I thought I'd clarify that I'm happy to make changes, but I'm not sure how you'd like it. Do you like the "test" target and Makefile changes, or should I get rid of that? If we keep the "test" targets then maybe that's a good reason to have a new "test" directory, but how do you feel about that? Move it to "scripts"?

Issue #485, opened 2017-09-04
wadinfo-builder.py can probably be removed


fragglet:
This script was put in place a long time ago when Freedoom was very incomplete and there were a lot of missing resources. Its main purpose is to point missing resources at "dummy" resource files (you'll notice that each directory contains a `dummy.png` or `dummy.wav`, `dummy.mid`, etc.). Nowadays we have a complete set of resources so I doubt it's necessary any longer. If it is still masking some missing resources it would be simpler to just manually edit `buildcfg.txt` to point to dummy until the resource is submitted.

Doing this should simplify the build quite a bit; the diagram in `BUILD-SYSTEM.adoc` should become one step shorter.

fragglet:
A quick survey of the dummy references being produced by wadinfo-builder.py:

Missing levels:
```
wadinfo_phase1.txt:e3m5 = dummy
wadinfo_phase2.txt:map26 = dummy
```
Missing sound effects:
```
wadinfo_phase2.txt:dspedth = dummy
wadinfo_phase2.txt:dsbspwlk = dummy
wadinfo_phase2.txt:dsflame = dummy
wadinfo_phase2.txt:dsflamst = dummy
```
These ones are weird. It seems that some textures are embedding sprites as patches, and as a result they get included in the generates [patches] list.
```
wadinfo_phase2.txt:; bosfa0 - no dummy
wadinfo_phase2.txt:; tfogf0 - no dummy
wadinfo_phase2.txt:; tfogi0 - no dummy
```

fragglet:
The rogue sprites-as-patches are actually from Final Doom textures:

* Plutonia's `A-BROWN4` texture embeds `BOSFA0` (fancy spawn cube on door front).
* TNT's `SLAD10` and `SLAD11` textures embed `TFOGF0` and `TFOGI0` (gives a "sparkle" effect on the UAC logo; I'm not sure what this even looks like in Freedoom).

Voros2:
Since we're on the subject of removing a script, I have always wondered if the music-duplicates script is needed anymore. It certainly isn't used anymore and there're no more duplicate tracks other than a 1-3 being shared by FreeDM and Phase 2 (which can be replaced with the pending music submissions).

PR #487, opened 2017-09-04
textures: Don't include sprites in pnames list.

fragglet:
A useful vanilla-compatible trick is that it's possible to include
sprites as patches inside texture definitions. It turns out that
both Final Doom IWADs added textures which do this, and Freedoom
inherits this as a result. We shouldn't include these sprites in the
pnames.txt that gets generated for inclusion in wadinfo.txt, since
they're already defined in wadinfo.txt under [sprites] and won't
be found under patches/ anyway.

This should help as a step towards resolving #485.


chungy:
Thanks

PR #489, opened 2017-09-04
build: Point missing resources at dummy resources.

fragglet:
Currently the wadinfo-builder script does this automatically, but
there are so few missing resources now that it's going to be simpler
to just get rid of it. These are the few remaining resources which
don't exist in the Freedoom tree.

Part of #485.


chungy:
Thanks, looks good to me.

PR #491, opened 2017-09-11
More new textures

Talon1024:


fragglet:
Size doesn't match on crlwdt3 - is that deliberate?

fragglet:
Never mind, I see that was already brought up.

PR #495, opened 2017-09-16
Level tweaks and fixes

Talon1024:
Check the commit messages for info on individual tweaks/fixes.


Voros2:
There's also a vanilla MAP28, edited by riderr3 (the creator of Urania): 

http://www.geocities.ws/voros/stuff/MAP28fd2_v2.wad (This is the second version, which doesn't conform to the vanilla savegame limit.)

http://www.mediafire.com/file/j1mbcorr2yd3rgr/MAP28fd2.wad (This is the first version, that takes out a room and some minor stuff to conform to the savegame limit)

Talon1024:
There's also Catoptromancy's map28fd2_v2_c2.wad.

PR #505, opened 2017-09-30
Rebuild nodes and reject table of MAP15

Litrivin:


Litrivin:
This pull request can be aborted, it is no longer needed.

PR #512, opened 2017-11-20
New patches

Talon1024:


Talon1024:
I think I'm ready for this stuff to be committed now.

PR #532, opened 2018-05-19
graphics: tweaks on freedoom guy standard HUD face

Ferk:
Doom engines stretch the images with 20% extra height. This, together with the already slim jaw and sharp chin in the face of the freedoom guy, makes it look very juvenile, stretched.

This change tweaks the face images to have a thicker jaw and a more mature look. It also tweaks the lip position a bit and uses a softer dark tone for the previously pitch black outline along the chin.

See forum post for comparison:
https://www.doomworld.com/forum/topic/100786-tweaking-freedoom-guys-face/


NickZ:
Looks good. Is this still WIP?

Ferk:
It has been ready for a while, but it occurred to me yesterday that it would be good to tweak also most of the other faces to ease the transition.
Give me until the end of the week, I think I can get to complete it by then.

Ferk:
Ok, I'm done tweaking the other faces that were slightly looking down in an angle.
They now match in thickness with the one changed so the transition is smooth as it was before the change.

fragglet:
Looks good, thanks!

PR #535, opened 2018-06-25
lumps: add further basic MAP33 support

guynamederick:
I am aware the purpose of the dummy MAP33 is to prevent PrBoom+ from crashing as it assumes the freedoom2.wad (and freedm.wad) IWAD is that of the BFG Edition of Doom 2. However, with the presence of a MAP33, ZDoom-based ports would not read HUSTR_33 as it plays freedoom2.wad as if there were only 32 mapnames like in the original Doom 2 IWAD. As such, MAP33 would be titled as "Unnamed." The ZMAPINFO file is to give MAP33 the mapname "Empty Facility" properly (MAP33: Empty on FreeDM).

For DeHackEd, I edited PHUSTR_1 to display the MAP33 mapname as Chocolate Doom and its derivatives would not read HUSTR_33 but instead go to PHUSTR_1. The source code of Chocolate Doom names MAP33: Betray this way when launching with the BFG Doom 2 IWAD. The PHUSTR_1 mapname won't cause any issues to mapset accompanied by a DEHACKED file that contain their own PHUSTR_1 mapname.

Edit: Replaced MAP33 dummy map with a modified MAP01 and MAP02 from test_levels.wad. DeHackEd mapnames changed to "Test Facility" to reflect change.

Doomworld thread topic: https://www.doomworld.com/forum/topic/101305-adding-further-map33-support/


chungy:
I don't think any of this is really necessary (afaik gzdoom never tries to go to map33 on its own, and doesn't crash). Certainly replacing the dummy level with the test levels is out of the question completely.

guynamederick:
If you certainly feel that MAP33 should remain unsupported, I'm okay with that. This is more or less minor changes based on how different source ports read mapnames. The modified test map is more of an afterthought.

Issue #530, opened 2018-05-11
Semi-official update of Double Impact


DoomeroHispalense:
https://www.dropbox.com/s/i9u8xc7sk5efe6g/dbimpact-fix.zip?dl=1
https://www.doomworld.com/forum/post/1872886

It seems that _RottKing_ has released an updated version of _Double Impact_ for the DWIronman League, which fixes two possible soft-locks and E1M8 is modified so that this map will not require anymore these dummy barons to trigger the 666 tag.

I have posted this just in case you (plural) are interested in updating the episode 4 of _Freedoom: Phase 1_ with this semi-official update.

Issue #521, opened 2018-01-01
C4M8 isn’t 100% vanilla, there are visplane overflows


ghost:
C4M8 isn’t 100% vanilla. After the player manages to unlock the final section once and for all, there is at least one location where visplane overflows do happen. I have attached (link 1) one demo recorded with Chocolate Doom. Obviously, the visplane overflows crash Chocolate Doom and abort the demo recording.

I strongly believe that Visplane Explorer cannot detect it (according to its developer, C4M8 was OK, link 2) because it cannot simulate the activation of every linedef. It can simulate the opening of every door and this isn’t exhaustive enough to predict certain visplane overflows.

Link 1: https://www.dropbox.com/s/xqvbhuue3r0mfcu/c4m8.zip?dl=1
Link 2: https://www.doomworld.com/vb/post/1535100

NickZ:
Hmm, that's interesting. Would you be interested in trying to fix it?

Litrivin:
I would do it, but I lack of a desktop computer at this moment.

NickZ:
closed in #536

PR #536, opened 2018-07-02
levels: Update C4M4 and C4M8

guynamederick:
Reponse to Issue #530: Both to update Chapter 4: Double Impact with the fixes included in the hotfix version. For C4M4 it only changes the teleporter special actions to be repeatable, as prior you can only trigger them once, which can result in a soft-lock. C4M8 is (mostly) the same as the hotfix version of Double Impact's E1M8, and Freedoom's past edits are replaced in favor of the hotfix changes (No more dummy Barons and what not).

Response to Issue #521: For C4M8, the final sector crashes vanilla. From the view (applies to the other side to the right) shown in the linked screenshot, this is where vanilla would crash given the amount of sectors present in the last room. I have edited it to remove some details so that it's unlikely for the game to crash when looking at this part of the map from a distance. I have tested C4M8 with ChocoRenderLimits.

Visplane spot: https://i.imgur.com/ZbRFceF.png

I have also changed the current endpic to match up with the current Freedoom assets, as it is awkward to see all the new sprites only to see an endpic with the old and outdated sprites and textures.

### DIFF
--- a/graphics/text/create_caption.py
+++ b/graphics/text/create_caption.py
@@ -12,7 +12,7 @@ import os
 font = ImageFont.load_default()
 
 
-txt1= "© 2001-2017"
+txt1= "© 2001-2018"
 txt2= os.environ['VERSION']
 background_image = Image.open(sys.argv[1])
 background_image.load()


chungy:
Thanks for it. I cleaned up the commit messages a little bit, in particular the "Chapter 4 Update" one included very long lines. Please keep them within 72-ish characters per line.

guynamederick:
I'll take note of that, thank you.

PR #542, opened 2018-07-25
graphics: New HUD face

Haaslok:
Replace long-time HUD face with completely new ones


NickZ:
I think this is a good start. I think it needs some work though. The faces look inconsistent; the front pain face in particular. There are also alignment issues, some of the faces go outside the box.

Haaslok:
I need more direction than "needs some work". Can you be more specific? Whats wrong with the pain face?

Also, how do I dictate offsets to sprites in the github? I know how to do it in Slade. 
I'm new to Doom modding, so I don't know these things.

unfa:
I can't really tell why, but this sprite looks to me like a face of a woman:
Screenshot_20190519_140549

MissLav:
yea i like the current face as well its nice

unfa:
I wonder - maybe it's already a thing people do, but I'd try to take pictures of an actor doing the expressions and then drawing over that as a template.

unfa:
I've got some acting skills, I could try and provide some face shots, but I have long hair, so that might be an issue (unless you want to go this way - could be a interesting twist to the character).

MatthewTheGlutton:
I am in favour of keeping the current (devbuild) faces, especially given how much of Freedoom's branding currently has it (cf. the newly formed Discord guild) it would have to be a huge, unequivocally better in every way improvement to totally replace it now.

chungy:
I basically agree with all the other feedback. This is of poorer quality than the present HUD face. And yes, I'm fond of the current one.

Ferk:
I agree the face is iconic, but specially the hair is an improvement.
I suggested it before in the forums but: @Haaslok what about taking the bits using some of your changes to improve the current face rather than reworking it?
image image

MatthewTheGlutton:
In defence of the current, scraggly, messy hair, it looks less cartoony and more like someone who's been out in the field for a while (and stuck in a helmet for a prolonged period to boot).

PR #544, opened 2018-09-14
dehacked: Turn on full brightness for a few frames

fragglet:
The firing frames for the zombie and assault tripod look weird if they're not 
shown in full brightness, so add a dehacked tweak to set them to full
brightness. Doom has the same bug but it isn't quite as noticeable.

This could theoretically cause a mod compatibility issue since unexpected
frames might be fully lit. However, (1) the worst that can happen is a minor
cosmetic issue, and (2) I don't know of any actual mods this affects (maybe
there aren't any). I think the benefit outweighs the potential downside.

### DIFF
--- a/lumps/dehacked.lmp
+++ b/lumps/dehacked.lmp
@@ -4,6 +4,25 @@ Patch format = 6
 
 # SPDX-License-Identifier: BSD-3-Clause
 
+# The following changes turn on full brightness for a few firing frames,
+# since these they show gun flashes and it looks weird if they aren't full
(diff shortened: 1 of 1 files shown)


chungy:
Looks good to me, thanks.

Nit: your commit message has long lines, which makes it wrap unnaturally in the terminal. Your text editor should have a setting or operation to automatically wrap at 72 chars.

fragglet:
Thanks, I did the whole thing through the GitHub UI, and it's difficult to
judge line length since it uses a proportional don't. I'm usually pretty
diligent about line lengths.

PR #545, opened 2018-09-23
Add "hexdump" diff documentation and instructions

selliott512:
In order to assist with diffing binary files, particularly WAD files,
instructions and comments have been added in order to assist with
diffing binary files with the "hexdump" tool.

This can be thought of as a companion to PR 481 since that change introduces binary changes where diffing with "hexdump" works well.

I did this because figuring out how to do custom diffing in git is not obvious. If you think it's too verbose then feel free to close it out. You can still refer back to this if you ever want to do it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,8 @@
 # Auto detect text files and perform LF normalization
 * text eol=lf
 
+# Append "diff=hexdump" to one of the following to enable binary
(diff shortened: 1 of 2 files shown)


chungy:
I will close this inactive PR; the other script fulfilled the purpose you were seeking anyhow :)

PR #546, opened 2018-09-29
sprites: improved player front walk animation

Ferk:
The previous animation always looked to me like he was holding his pee. So I tried to improve it. 
I also redured a bit the sudden swing with the weapon that it does in the middle of the walk.


Ferk:
I've also committed an improved animation for the back view:
Before - After:
.gif.642d94d3c8979d070117e5472ff8a5b2.gif)

PR #547, opened 2018-09-30
Add fix-map-names script and build targets

selliott512:
Add scripts/fix-map-names to fix and test map names. Map names are
tested with make target "test-map-names" and fixed with "fix-map-names".


NickZ:
This looks good and works as advertised. Thanks!

Issue #443, opened 2017-04-23
MAP08 100% kills?


JohnnyTheWolf:
Has it been reported that it is currently impossible to get 100% kills in MAP08? It appears a dozen or so former humans are not getting activated and teleported outside of their closed-off area.

Voros2:
This has been reported and is a known bug.

igdegoo:
this issue is fixed by #548 and should be closed

PR #548, opened 2018-10-15
levels: Update MAP08

guynamederick:
Made some tweaks and some fixes to the current MAP08. Tested with Chocolate Doom 3.0.0 and 100% stats is possible, so issue #443 is resolved in that part.

Changes:
-Difficulty tweaks: More monster variety and items were added/removed, some depending on the difficulty of course.
-Some more detail: Map was a little plain to look at, so I thought to edit and add some bits. One of the secrets (the outdoor one) was pretty much empty so I added something to it.
-Monster traps: Tried to improve the chances of monsters spawning in the map, though the way it's set up can still result in some late spawns.
-Rebuilt nodes with ZokumBSP

If there's any suggestions to be had here, I would like to know.


NickZ:
Very nice! Merged.

PR #549, opened 2018-11-19
New Music for map:c2m8

Ajanddino:
Added my own music to map c2m8. It's a more up-tempo song for the level.


NickZ:
very nice! I like this track a lot; it plays much better with the frantic gameplay going on in C2M8 with trying to avoid 2 tripods trying to kill you.

PR #551, opened 2019-02-02
levels: create c3m5

BeachThunder:
c3m5 now exists


NickZ:
Hey, thanks for the PR! This is a really nice level!

A couple of things before I merge this in:

* We should not be giving away the SKAG (BFG) in episode 3 (i.e. without the player discovering a secret), simply because it makes the end boss way too easy. I would change the SKAG to a plasma rifle, and the current plasma rifle pickup to a chaingun.
* The dark cavern is currently way too dark on the default gamma settings. I would bump up the lightness a bit so that it's navigable on gamma level 1.

After those changes, I'll merge this in.

Calinou:
Nice level :slightly_smiling_face:

I think I spotted a texturing error in the lava cavern (the one behind the yellow key door). The first step uses a lava texture instead of a gravel texture:

screenshot_doom_20190217_140451

BeachThunder:
@NickZ done and done. I also made a few other minor adjustments while I was at it. 

@Calinou, it wasn't an error, I wanted to make it look like lava was covering the bottom step. But, yeah, in retrospect, it's probably better if it's a gravel texture. Thanks.

NickZ:
Thanks! Merged!

PR #552, opened 2019-02-24
Convert GZDoom BIGFONT to the new system

K0stov:
With this pull request, I want to raise awareness about a new [**BIGFONT**](https://zdoom.org/wiki/Font) system that’s set to appear in GZDoom (it already has appeared, but so far only in devbuilds). To make this a bit clearer: **BIGFONT** is the text format used by header texts in GZDoom’s various menus, as shown in the screenshot below. Until now, it was primarily based on the [FON2](https://zdoom.org/wiki/FON2) format (there’s a lump named **DBIGFONT** in the Freedoom IWADs that is based on this format).

BIGFONT

Graf Zahl, the author of the source port, has overhauled the BIGFONT system; while it used to rely on the old FON2 binary format, it now depends on individual lumps that are based on Unicode. What I’ve done here is extracted the BIGFONT from **freedoom2.wad** and turned all characters into individual sprites that are named after their code points in Unicode.

I’m not entirely familiar with the folder structure of this repository and how the scripts in the [text](https://github.com/freedoom/freedoom/tree/master/graphics/text) folder work (which is where I’ve put these sprites), but for future builds, it should at least provide a base of some sort.

I’ve marked this PR as **[Do not merge]**. First, we should discuss whether there’s a better way to do this, considering the current GZDoom BIGFONT files included in the Freedoom wads are limited compared to what’s in the repository: I’ve noticed that these characters are [already cut up](https://github.com/freedoom/freedoom/tree/master/graphics/text/fontchars) into separate lumps, complete with uppercase and lowercase variants, as well as various other characters that could be useful, since GZDoom is undergoing major advancements in [localization](https://forum.zdoom.org/viewtopic.php?f=4&t=63664) and [translation](https://forum.zdoom.org/viewtopic.php?f=49&t=63737). Maybe it’s better to use those characters instead, and rename them based on Unicode.

Then, of course, we wait until the new BIGFONT system makes it into an official GZDoom version.


NickZ:
@Nemrtvi what's the status on this? I think GZDoom introduced this in 4.0?

Nemrtvi:
> 
> 
> @Nemrtvi what's the status on this? I think GZDoom introduced this in 4.0?

Yes, it did.

axredneck:
Is it a fix for issue #553 ?

MatthewTheGlutton:
Has anyone posted about this on the ZDoom forums? Might be good to get the GZDoom devs to take a look. (I'd post but I'm pretty clueless as to how all this stuff works so I wouldn't even know how to describe it except link and "any thoughts?")

Nemrtvi:
> 
> 
> Has anyone posted about this on the ZDoom forums? Might be good to get the GZDoom devs to take a look. (I'd post but I'm pretty clueless as to how all this stuff works so I wouldn't even know how to describe it except link and "any thoughts?")

I did, when I first made this PR:
https://forum.zdoom.org/viewtopic.php?p=1092851#p1092851

Graf says it's preferable for Freedoom itself to include these fonts. I've provided the means for it with the PR, but someone else might want to expand it to a complete font set that I've seen somewhere among Freedoom's files.

chungy:
It would make most sense to generate any necessary lumps directly from the PNG files (we should do it for DBIGFONT too, or can this sort of system be a replacement?).

I'm not seeing any kind of documentation on the ZDoom Wiki nor in GZDoom itself about the formats or expectations. Is there a place I should look?

Nemrtvi:
> 
> 
> It would make most sense to generate any necessary lumps directly from the PNG files (we should do it for DBIGFONT too, or can this sort of system be a replacement?).
> 
> I'm not seeing any kind of documentation on the ZDoom Wiki nor in GZDoom itself about the formats or expectations. Is there a place I should look?

Don't know about any documentation, so it's high time we start writing something. In the meantime, if you have any questions, I recommend leaving a post in this topic:
https://forum.zdoom.org/viewtopic.php?f=4&t=63664

To add a symbol using this system, you need to name a lump after its Unicode symbol counterpart (on the site below, it is shown by the "Unicode number" when you click any symbol - for example, the capital letter A needs to be named 0041.lmp)
https://unicode-table.com/

Ferk:
My impression is that this new system requires using a pk3, since it depends on a specific placement within the pk3 folder structure (ie. `/filter/game-***/fonts/bigfont`), if it cannot be used in wads then it won't be possible to add it to freedoom.

selliott512:
I took the liberty of converting this to a draft instead of the "[Do not merge]" prefix. It makes it easier to search for PRs by state.

mc776:
Is anyone working on this anymore? Should it just be closed as stale and the issue revisited if/whenever?

mc776:
What exactly are these lumps? Are they just PNGs named "lmp", or Doom graphic lumps, or something else entirely? What do you even open and edit them with?

K0stov:
They're Doom graphic lumps, if I remember correctly.

mc776:
I wonder if there's some way to have these start out and PNGs and then let Deutex convert them to graphic lumps during the build process - it would make reviewing FD's "source" a lot easier.

mc776:
All these lumps represent characters that are already represented by graphics in FD, right? Would it make more sense, instead of having separate files this way, to add them to the buildcfg using `=` to point to the graphics?

Or maybe we can somehow convert everything to this system... like just reorganize the entire contents of the fontchars directory and all the text graphic creation scripts? Is there any technical reason why we can't unify everything like this?

mc776:
> (EDIT: or practical reason why we shouldn't? Just thinking about in the future if there might be more translation work that ends up touching on level names or something... or heck, even a level name with a heavy metal umlaut...)

> [I'm not sure any UMAPINFO credit displayer out there even supports anything you can't type on a US keyboard with nothing but Shift.](https://github.com/freedoom/freedoom/issues/1455)

PR #555, opened 2019-04-15
Player crouching sprites for GZDoom

MatthewTheGlutton:
Includes moving, firing and pain, but not death or xdeath (since GZDoom does not use them by default anyway).

As of this commit of GZDoom https://github.com/coelckers/gzdoom/commit/c1a3d248f608141eee16f0b25c3c22c8c198f496 these will no longer be overwritten by GZDoom's own crouching player sprites at runtime even if provided in an IWAD.


NickZ:
Thanks, merged.

PR #556, opened 2019-04-18
Sprites: New Ammo and Shell boxes

YukiHaze:
IMO the current ammo and shell boxes have a strange fade shading to them (This applies to a lot of pick ups, but these ones in particular looked really off), so I created these as replacements.

If I did anything wrong when creating this pull request please let me know, it's my first time doing this.
Compare


MatthewTheGlutton:
Looks good, but three criticisms immediately spring to mind:

1. Does Freedoom ever establish that the bullets and shells are 9mm and 12ga? (the pickups just say "ammo" for... whatever it is that is shared by both the pistol and the minigun) It might be better to just leave it ambiguous.

2. The text is (at least against this white background) a lot less immediately readable with those outlines and gradients (basically the same "strange fade shading" but for the text alone). The labels on the old graphics are very simple and clean and stand out immediately.

3. The old shellbox, though not as technically well drawn, has a much more pleasing design. It just *feels* like something I'd be happy to pick up a box of at the corner store on my way home from work... and thus more plausible as packaging. (It's also totally ambiguous what it is when it just says "SHELLS" rather than the 12 which, even knowing it's shells, might make one momentarily think there were supposed to be 12 in the box.)

Ferk:
The old shellbox also has visually thicker shells (3 pixels width each shell), the shading even gives them a hint of cylindrical shape. in the new one the shells have the same width as the normal ammo (2 pixels), just more flat.

Personally, the only thing I would improve in the old graphics is the color scheme of the shells box. I don't really think the fade shading looks weird, it actually gives it some shadows that make them feel like actual objects placed in the map, otherwise it'd look like a flat texture. Original Doom AMMO sprites also has a fading shade, although in a different direction.

MatthewTheGlutton:
I personally really like the McShells vibe... kinda hints at some of the implied setting without really saying it.

MatthewTheGlutton:
Hideous Destructor uses the ammo box sprite as a literal generic "ammo" box that might contain anything. Is anyone aware of any other mods that do this? If not, and the Freedoom sprite is ultimately changed to be specifically labeled as pistol ammo, I can supply a different lockbox sprite for HD, but if other mods also rely on the genericity of AMMOA0 that will be a consideration.

In any event, I think any change to these sprites in the direction suggested here should make the shell box noticeably larger since it's carrying 20*7=140 projectiles versus the clipbox's 50. The player's implicit encumbrance ratio is also 4 pistol rounds for every shell (200/50); multiply that by the pickup amounts and it's 1x50=50, 4x20=80.

MatthewTheGlutton:
To address some of the issues I had with this PR I frankensprited up a couple things:
image
Thoughts?

chungy:
we ended up having a consensus with #604, which was already based on your edits. Closing, but thanks :)

PR #562, opened 2019-05-16
dist: Add releases and content rating to AppStream metadata

### LINKED ISSUE #522: Add OARS entry in the appdata
As a violent game, Freedoom should have the Open Age Ratings System set in its AppData so users could check what's appropriate for them (or their children).

Getting the OARS XML entry is very easy, just use the form in the [OARS website](https://hughsie.github.io/oars/) and then add the resulting XML to the AppData file. Here's an [example](https://github.com/flathub/com.viewizard.AstroMenace/blob/master/astromenace.appdata.xml#L24).

I'd gladly submit a PR but this needs to be done by someone who knows the game well (in order to answer the questions in OARS about the game's content).

Calinou:
Those tags are now required by Flathub.

This closes #522.

### DIFF
--- a/dist/freedm.appdata.xml
+++ b/dist/freedm.appdata.xml
@@ -43,4 +43,10 @@
       <caption>DM18: Energy Facility</caption>
     </screenshot>
   </screenshots>
+  <releases>
+    <release version="0.11.3" date="2017-07-18"></release>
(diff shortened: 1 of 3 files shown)


chungy:
thanks

PR #565, opened 2019-06-01
change some instruments of d_map30

igdegoo:
for some reason i felt like doing this so i did. sorry about the mess in commits im a github noob but it shouldnt be a problem


igdegoo:
ima make a new pr without the messed up commits

PR #566, opened 2019-06-01
change some instruments of map30

igdegoo:
For some reason i felt like doing this so i did


chungy:
Thanks for the submission, but I should note that the near-total lack of information about this change made me delay reviewing it. It would be helpful if you include more next time, especially more than a commit message of only "h" :)

On listening to it, I do find the change to a choir interesting. Not a bad change I think. I'll merge it and rewrite the commit message.

PR #568, opened 2019-07-01
dehacked: Use three periods instead of a comma

coltongit:
Use three periods instead of a comma for the inner monologue, as this helps make the text flow a bit better.

### DIFF
--- a/lumps/dehacked.lmp
+++ b/lumps/dehacked.lmp
@@ -398,7 +398,7 @@ E3TEXT = The abomination explodes into\n\
 E4TEXT = Despite fighting an army of monsters,\n\
          you managed to find a ship here.\n\
          Looks like freedom is yours.\n\n\
-         "Time to get out of here,"\n\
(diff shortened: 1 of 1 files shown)


MatthewTheGlutton:
That looks grossly unprofessional and does not help the flow at all.

Ferk:
https://english.stackexchange.com/questions/73216/punctuation-of-direct-speech-edge-cases

MatthewTheGlutton:
Thanks for the link. It actually helps me understand why I have such an intuitive revulsion towards this PR.

The only ellipses cited there are used for things that are clearly cut-off sentence fragments. "Time to get out of here" is a full sentence all on its own. (I can't even *think* of any way to add to it that isn't some purely optional modifier ("...now that I have the Rune of Space Magic!") or a second clause that would have to be joined by an "and".)

chungy:
I think it's fine. But still, set your text editor to break your commit messages properly (merging after fixing that) :)

PR #572, opened 2019-07-09
 More intermission text adjustments 

MatthewTheGlutton:
I still hate that "...", but I'm starting to think the aggravation is not really that itself but the use of a direct quote for "you" the player for something they never actually said.

So looking at the E4 text again to see how it could be avoided, I noticed a bunch of other little weird things... the past tense on the first sentence clashing with the rest of the present tense; the "despite" doesn't really fit with the logical progression; and while I know exactly what "plump down" was intended to convey it seems a bit nonstandard. It's also missing punctuation at the end, in contrast with the other intermission texts.

I've also modified C5 and C6 for similar reasons.

### DIFF
--- a/lumps/dehacked.lmp
+++ b/lumps/dehacked.lmp
@@ -399,11 +399,11 @@ E4TEXT = After yet another army of monsters\n\
          you manage to find a functioning ship.\n\
          Looks like freedom is yours.\n\n\
          You practically melt into the soft\n\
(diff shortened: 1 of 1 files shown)


MatthewTheGlutton:
Not really related to this PR but just so it doesn't look like I'm always complaining: I really like the symmetry between the player raising their weapon and the monsters raising their weapons going between C5 and C6.

PR #573, opened 2019-07-09
dehacked: Fix minor punctuation errors

coltongit:
Noticed minor errors in the new text, so went ahead and fixed those here.


MatthewTheGlutton:
I actually had those commas there originally, but with them at the end of the line like that I found myself lurching like someone trying to learn to drive manual, so I deleted them as being unnecessary since the deletion created no grammatical ambiguity.

I have no strong thoughts about this either way.

chungy:
I agree that a little grammatical leeway is acceptable to make it visually more appealing. Sorry to close, but thanks for the effort :)

PR #580, opened 2019-08-18
Add a manual for Freedoom.

fragglet:
I've been working on this over the past couple of weeks and I feel like it's complete enough now for review. Current PDF output can be found here:

https://soulsphere.org/random/freedoom-manual.pdf

It's intended to be an equivalent to [Doom's original manual](https://doomwiki.org/wiki/Doom_instruction_manual) that new players can use as a reference. I'm intending to expand it in the future with some additional information Freedoom-specific sections.


chungy:
This is excellent work :)

I've thought about a game manual in the past, especially if printed copies are ever done... I'm glad you've done this.

NickZ:
@chungy can you notify the debian packager that freedoom now requires asciidoctor-pdf as a build dependency?

chungy:
When it's required, we'll get there :)

Ferk:
Awesome! I miss those manuals that always came with the game in the good old days :P
Even for people who already know Doom, this would also help clarify all the name changes, and what's what.

Maybe it would be a good idea to add a sort of disclaimer indicating that "Freedoom by itself contains no game logic, the game behavior depends on the engine being used and how accurate is it to the original Doom engine. The instructions shown here are just a representation of such original behavior." 
Or something like that.

Epictyphlosion:
@fragglet, I've tried to send you an email regarding your automatic Slige generator and how you got it to work, but the email on your website apparently doesn't work. Is there any way I can contact you other than here?

PR #582, opened 2019-08-19
Fix shell icons in the manual

### LINKED ISSUE #581: Manual uses the wrong icons for shell ammo boxes
The manual added in #580 uses the wrong icons for shell ammo boxes (according to the PDF output linked in the PR): 

image

Specifically, they use the same icon as the energy ammo boxes.

Calinou:
This closes #581.


fragglet:
Whoops! Thanks for fixing.

Issue #577, opened 2019-08-04
Backpack has wrong offset


Proluiz:
While i was mapping a noticed a small glitch in the offsets from backpack.
Screenshots/examples:
![image](https://user-images.githubusercontent.com/6380147/62430455-3d1b0580-b6f3-11e9-8264-65a1ebb4e992.png)
![image](https://user-images.githubusercontent.com/6380147/62430463-46a46d80-b6f3-11e9-9340-f1667190f1f6.png)

Proluiz:
Since nothing has pogressed so far I've decided to check the sprite myself. (BPAKA0)
Actual offsets:  8x26
![image](https://user-images.githubusercontent.com/6380147/63356180-b9415a00-c33d-11e9-9af6-09090ee0566d.png)

Better offsets: 11x26
![image](https://user-images.githubusercontent.com/6380147/63356169-b0e91f00-c33d-11e9-8ce9-fd3ee637f821.png)

I dont know how to edit it otside slade i need you guys to do it.

PR #584, opened 2019-08-20
buildcfg: Fixed the offset of backpack

Proluiz:
Backpack is now centralized to the actor from 8 to 11.
Based in issue: #577


Proluiz:
err, ignore it... #583

PR #585, opened 2019-08-21
Fixes to the skill levels

MatthewTheGlutton:
Extreme Carnage should have a period to be consistent with the other skills. Yes, I know the previous labels are full sentences, but given we all know the final skill must end with a ! it does end up being left as the only unpunctuated one.

I'm starting to dislike "INSANITY!" for multiple reasons, but it also seems kinda flat and uninspired after decades of eXtReMe / "totally random crazy monkey spork uwu" whatever in popular culture.

"Mayhem" might be a better descriptor anyway for the additional speed and damage and randomly respawning monsters.


chungy:
Looks good to me, thanks

PR #593, opened 2019-09-01
Organize .gitignore

selliott512:
Organize .gitignore into three sorted sections based on a comparison of
a clean and full build.

I noticed that manual/freedoom-manual.pdf was not being ignored. Instead
of just fixing that I thought it might be helpful to organize .gitignore.

I compared directories and extensions in a clean and full (dist and
manual) build like this:
```shell
mkdir /tmp/fd

cd ~/git/freedoom-clean
git clean -fxd
find . -type d | grep -v "^\\./\\.git" | LC_ALL=C sort -u > /tmp/fd/dirs-clean
find . -type f -name "*.*" | grep -v "^\\./\\.git" | sed -r "s|^.*\\.||g" | LC_ALL=C sort -u > /tmp/fd/exts-clean

cd ~/git/freedoom-full
find . -type d | grep -v "^\\./\\.git" | LC_ALL=C sort -u > /tmp/fd/dirs-full
find . -type f -name "*.*" | grep -v "^\\./\\.git" | sed -r "s|^.*\\.||g" | LC_ALL=C sort -u > /tmp/fd/exts-full

diff /tmp/fd/dirs-*
diff /tmp/fd/exts-*
```


chungy:
Some of these really belong in their appropriate subdirectories; for instance, there is already a manual/.gitignore but it probably should be changed to ignore `*.pdf` and `*.html`.

Would you mind organizing them with per-directory ignore files?

selliott512:
Yes, I can try to do that.

As a first pass to understand what work remains after my suggested top level .gitignore I deleted all other non-top level .gitignore files. I then added patterns until "git status" returned clean and until there were no files in the index that were ignored. The later tends to suggest a .gitignore bug since checked in files shouldn't be ignored. To check the later I created a [script named git-ls-ignore-index](https://github.com/selliott512/git-items) . I understand you want things in subdirectories, but just to share what I have so far If I were to add the following lines to my suggested to top level .gitignore with no other .gitignore files then both criteria are met:
```
/bootstrap/*.wad

/graphics/*wi*.png
/graphics/credit.png
/graphics/fd*title.png
/graphics/help.png
/graphics/m_*.png
/graphics/prboom.png
/graphics/t_phase*.png
!/graphics/m_dm.png
!/graphics/m_doom.png
!/graphics/m_lscntr.png
!/graphics/m_lsleft.png
!/graphics/m_lsrght.png
!/graphics/m_skull1.png
!/graphics/m_skull2.png
!/graphics/m_therml.png
!/graphics/m_thermm.png
!/graphics/m_thermo.png
!/graphics/m_thermr.png

/graphics/text/*.png

/lumps/*_deh.lmp
/lumps/*pname.lmp
/lumps/colormap.lmp
/lumps/dmxgus.lmp
/lumps/fd*txtr*.lmp
/lumps/genmidi.lmp
/lumps/playpal.lmp

/lumps/colormap/*.lmp

/lumps/genmidi/*.lmp

/lumps/playpal/playpal.lmp

# "stub" files or "mkdir" in Makefile needed for the following.

/lumps/textures/freedm/

/lumps/textures/phase1/

/lumps/textures/phase2/
```
I can move everything that has a path into the corresponding subdirectory.

As to *.pdf and *.html, there's a top level *.html, so maybe leave that top level? *.pdf is only in the manual, but would it be an error if someone checked in a PDF elsewhere? I can put it in the manual if you want.

Also, I added *.orig, *.swp and *~ because I always have them in my .gitignore files.

The "!" logic is unavoidable without having a huge list of files.

Ideally checked in files and generated files would have better separation, but i understand that this is how it evolved.

Also, the last three directories listed (under lumps/textures) have .gitignore files with no other files, so they're effectively stub files that assure that the directory exists. A "mkdir" of those directories in the build is probably the cleanest solution, so I can try to do that.

chungy:
I have been thinking of this today, and if it's organized as such, I don't mind a single top-level gitignore after all, especially one that looks as clean as that. I prefer that the project is self-consistent either way. Good work!

> Also, the last three directories listed (under lumps/textures) have .gitignore files with no other files, so they're effectively stub files that assure that the directory exists. A "mkdir" of those directories in the build is probably the cleanest solution, so I can try to do that.

If you'd like, go ahead and do that. I didn't realize we had such directories laying around. The top-level wads directory used to be like that too, but it's created as needed now.

selliott512:
I like a single top level .gitignore too. Everyone knows where the ignored patterns are that way.

I hope the comments aren't too pedantic. I think they are helpful.

I ended up leaving PDF top level. If someone wants to check in a PDF it's easy to change, but I can change it if you want.

I did a "make dist", "make install" and "cd manual; make". Hopefully that's sufficient to build everything that needs to be ignored.

Incidentally, I noticed two Python files that seem to need Python 3 now possibly as a result of Blacken:
```
lumps/genmidi/genmidi.py
lumps/genmidi/mkgenmidi
```
I only looked into it very briefly, namely changing the first line to "python3" for my testing.

To keep things neat I squashed and did a push force since GitHub does not seem to have a way of squashing on merge.

selliott512:
Also, I added the creation and clean of "the last three directories" I mentioned to lumps/textures/Makefile.

chungy:
All looks good to me, thanks.

Extra comments:
> I ended up leaving PDF top level. If someone wants to check in a PDF it's easy to change, but I can change it if you want.

PDFs aren't source documents anyhow, I don't expect there to be any good reasons to add them.

>Incidentally, I noticed two Python files that seem to need Python 3 now possibly as a result of Blacken:

Interesting. Maybe  a `from __future__ import print_function` is all that's needed? My system lacks Python 2 these days, and I probably won't really mind to change all the executable calls from `python` to `python3` either. [Python 2 is mostly dead](https://pythonclock.org/) and I don't think it's worth caring about it anymore.

PR #594, opened 2019-09-01
Remove Doom 3: BFG Edition support

chungy:
Might have been a nice idea to be able to run under absolutely any
Doom engine, this has caused problems with PrBoom-Plus for
compatibility, with that engine detecting our game as a BFG Edition
IWAD and applying certain workarounds that don’t actually make sense
in this case.

Requesting feedback on this. Support, disapproval?


guynamederick:
Until PrBoom+ ever gets updated or some fork follows it, I'm all for removing BFG Edition support so mod compatibility issues don't occur. It bothers me that the SS to Zombiemen change is even a thing, I'm glad Crispy Doom removed that "fix" long ago.

Scythe 2 MAP31 features a custom marine enemy that runs fast and shoot plasma. When playing with freedoom2.wad, they get replaced with zombies instead.
scy2_31

chungy:
There isn't really an "until" -- I don't expect to ever restore BFG Edition compatibility. I'm thinking it over, and overriding the SS monsters makes it too high of a cost. Thanks for the feedback.

PR #597, opened 2019-09-01
SkullButtons

MissLav:
Edit to SkullButtons to look more like dooms


chungy:
There is a technical problem with your commit, that you added new files with uppercase characters for their name. The old ones in lowercase would still be used by the build system.

Beyond that, I'm not fond of the edit, to be frank. They come out as too dark I think.

MatthewTheGlutton:
I like the shading on this a lot more than the current one, which (especially with the pure white without any yellowing) always had that low-budget early 90s render look to it.

Looking *more* like Doom is a minus for me though. But the Freedoom skull still looks a lot more human and less stylized which detracts a lot from the similarity.

As an experiment I tried adding a jaw to it but it didn't look right at all once I squished it back down to the original dimensions. (Would've been an interesting reimagining of the Doom imp though.)

Just plain scaling it down left a lot of space on either side, though, which could be filled by other ornamentation.

So that said, just purely as a thought experiment I'll leave this here: 
image

selliott512:
Not to discourage creative effort, but from a legal standpoint should Freedom have commit message that admit the intent is to "look more like dooms"?

chungy:
No, and the project generally actively avoids making anything look close to Doom.

I'll close this PR, it is surely disappointing @MissLav but I hope it doesn't discourage you from future work :)

Issue #586, opened 2019-08-21
Textures: BIGBRIK1 and BIGBRIK3 differ from each other, unlike doom2.


Proluiz:
Bigbrik1 needs to be repatched to match the current version of bigbrik3, it make so easy to noticed the diferences in some map mods.

In Doom2:
![BIGBRIK1](https://user-images.githubusercontent.com/6380147/63427878-915cff80-c3ec-11e9-951d-94a274921029.png)![BIGBRIK3](https://user-images.githubusercontent.com/6380147/63427885-9621b380-c3ec-11e9-8f10-bcd237a2b29f.png)![BIGBRIK1](https://user-images.githubusercontent.com/6380147/63427878-915cff80-c3ec-11e9-951d-94a274921029.png)

In Freedoom2:
![BIGBRIK1](https://user-images.githubusercontent.com/6380147/63428009-dbde7c00-c3ec-11e9-8b97-85941c7c93a3.png)![BIGBRIK3](https://user-images.githubusercontent.com/6380147/63428012-de40d600-c3ec-11e9-8e9c-fa8c3e207031.png)![BIGBRIK1](https://user-images.githubusercontent.com/6380147/63428009-dbde7c00-c3ec-11e9-8b97-85941c7c93a3.png)

chungy:
Looks more like an intentional difference to account for the light source... I'm personally in favor of keeping it like this.

I could maybe see it making secrets easier to find, but are there examples where it makes maps too easy?

AXDOOMER:
It doesn't make sense bright up the pixels on a texture with a light source considering that light should propagate to adjacent wall textures as well.

The only valid example is the following case, where the resulting light of a light source can be contained within the texture itself. (Doom 2 MAP27)
![Screenshot_2019-08-21_14-57-13](https://user-images.githubusercontent.com/6194072/63460253-2a5b3d00-c424-11e9-9fa2-9665937ede2e.png)

It's up to the author of the map to make design decisions about light and I don't think Freedoom should include design decisions that will break what a map was supposed to look like in the original Doom. If the author wants to show that a wall emits light (or bright up the pixels on a wall), he will add a sector with a different light value. 

Other designs decisions were made in the past, like changing the candle sprite for a little lamp, and I think this inevitably breaks how some maps were intended to look.

NickZ:
Makes sense to me. Do you want to take a crack at fixing it?

fragglet:
[...]

PR #600, opened 2019-09-06
Redraw of RW23_3, RW23_4 and RW38_4

MatthewTheGlutton:
I noticed while looking at issue #586 that these patches were looking a little distressingly similar to their Doom2 counterparts. Not the proportions, I know a lot of mappers will use those so they're non-negotiable, but the noise on the bricks' surface - the teal bits are virtually identical.

This attempt to re-draw isn't a perfect clean room, I had to look at the originals to see where the brick lines were and how tall the light was, but I can guarantee that every pixel was drawn independently of the Doom2 textures. (The original colours were sampled from Freedoom's BSPIA2A8 and FIREA0)


chungy:
looks good to me

PR #602, opened 2019-09-08
new skull buttons SW2_#

MissLav:
messed up on the naming of the edited files


MissLav:
also i used [yuki's)](https://github.com/freedoom/freedoom/pull/556) new / old font

PR #603, opened 2019-09-08
new skull buttons sw2_1,2,7,8

MissLav:
added new skull buttons based on MatthewTheGlutton's edit 
oldnewskull

Matthew's edit
Matthewskull

also used [yuki's](#556) new / old font 
had to remove and re add pull request due to capitalizing the name of the buttons


MissLav:
replaced flat2 
oldnewflat2

doomfighter33456:
awesome

selliott512:
I think your skull buttons look great, but I have a small concern - they seem like they might be harder to see. I know with both FD and the original Doom one of my frustrations is overlooking buttons, so I need all the help I can get, but maybe that's just me.

A git tip - if you create a separate branch in your fork for each change then your "master" branch won't be tied up while your pull requests are pending.

chungy:
I had the same concern, but they seem to be alright in-game.

Another comment though, your branch is quite messy. I've gone ahead and squashed it down to one commit, but if you could use rebase to keep the history tidy, that'd be great :)

PR #604, opened 2019-09-10
[sprites] Chemical propellant ballistic pickups

MatthewTheGlutton:
After a bit of messing around further to my comments on #556 I ended up doing an overhaul of all the clip and shell sprites. (Sorry YukiHaze your proposed improvements do look technically better than the current but I can't agree with the substantive changes.)

Clip is still a clip and not a mag. I've got a headcanon that they can be linked to each other to form a chain for the minigun. My hope is that the new version will look more like a block of something with a distinct shape where the old one was kinda noisy and indistinctly outlined.

Ammo box is reduced to something that would plausibly fit only 50 pistol rounds without a lot of padding. If the "AM MO" is a bit cheesy I've got another version with just an abstract label.

Four-shells is mostly unchanged, but the one on the bottom has always been too invisible for me (I always do a double-take wondering who thought the pickup had only 3 shells) so I've made it pop out a bit more. I've also fixed the offset in buildcfg.txt so it's horizontally centered.

Shellbox no longer has the McDonald's fry packaging but I still insist that the colours are a neat addition to an otherwise bland FPS cliché. The box is now a 4x5 block with the same scale as the small shell pickup.
beforeandafter


MatthewTheGlutton:
Is there any particular historical significance of the 3+1 arrangement? Because this looks a lot better to me:
shelb0big

selliott512:
I like the shells.

I think the ammo box looks cool, but is could it possibly be a bit wider and easier to read?

My inclination is to make things large and visually distinct even if it makes it less realistic. Others may disagree.

MatthewTheGlutton:
If it were wider it could be easier to read!
secondambox

AXDOOMER:
It's so hard to decide... Now I like the one with the bigger font a little more because it's sharp and more readable. The E and S don't look too nice on the other ones (although I think the A and M on the ammo box sprite are 100% fine). 

Can you get a screenshot in PNG format where you look straight and the boxes are at equal distance? 

Also, it would be good to have more opinions.

Calinou:
The one on the left is the most readable to me, so I'd go for that one as well.

Ferk:
It's definitely an improvement over the current one either way.
Personally I'd go with the middle one. It's more readable than the tiny one (that almost spells "SMELLS") but not overly stylized, which keeps it a bit more realistic and make the shells themselves look bigger in contrast.

Something that might be worth trying out is adding a small transparent pixel in between the shells on the top row, similar to how it was in the original, to give it more depth.. but it's a tiny detail that's not needed if you prefer it otherwise without it.

MatthewTheGlutton:
I went back and forth about roughing up the top edge of the shells, and barely declined to so each time, but this is definitely looking better.

Using the medium-size text with the small-text perspective on the box. This preview is scaled correctly.
shela0_embiggened

chungy:
Seems good, I'm going to merge this, though also rebase; it'd be nice if you could do that yourself more often instead of making extraneous commits. Also you seem to have accidentally overwritten shela0, but I reverted it.

MatthewTheGlutton:
I tried looking up what rebase meant and the graphs just made my head spin. I'm guessing it flattens the PR to a single commit that contains everything?

Is there a way to do it from the Web interface?

chungy:
That's one function it can do, and I have no idea. I never used the web interface.

caligari87:
More simply, rebase temporarily removes your changes, fast-forwards to the point the master is at, and then re-applies the changes as if you'd made them at the current point instead of in the past.

Squashing turns your entire PR to one commit.

A PR can be rebased or squashed in the web interface *upon merging* (see the drop-down arrow next to the button) but is generally more polite to do on the submission-side to make sure there's no conflicts.

PR #605, opened 2019-09-11
Fix the offsets of Missile Launcher.

Proluiz:
Just a small fix for Missile launcher, it's a little offset to the left, I also raied the height of it a little, so it's not "sinked" on floor.


Proluiz:
i have a screenshot too:
Screenshot_Doom_20190911_110047

PR #609, opened 2019-09-15
red/green colourblind friendly sboxa0.

MatthewTheGlutton:
I'm really sorry for not checking this earlier, but I just tried to view the new sprite using a red/green colourblind filter I had lying around and it turns out to be almost completely illegible:
colourblind
The darkened sprite in this PR is on the right, current on the left.

The new pistol/chaingun ammo box is fine. I didn't bother checking the others since there's no text on them.


chungy:
merged as 2769d7c31b0f6c4668be3fedcc7a5e3b8c376142

PR #629, opened 2019-10-10
textures: Add back old Espi STAR* patches.

fragglet:
These were Espi's second set of STAR* textures that he made for
Freedoom, and he ultimately replaced them with the ones that we have
now. However, I always liked these textures and the fact that they have
a very different and distinctive look compared to the original Doom
STAR* textures.

While this doesn't roll back STAR* to using these textures (and I don't
think we should), it adds them back as a second set of textures under a
different name (s/STAR/ESPI/). It would be nice if some of the Freedoom
maps end up making use of these in places and a nice tribute to Espi who
contributed so much to Freedoom in its early years.


fragglet:
Timing is also appropriate since it's been just over 10 years since Espi passed away.

MatthewTheGlutton:
Which STAR* textures are the current ones again? I'm having trouble finding them on the repository.

fragglet:
Look in /patches for the files named SWxx_x.png, eg.

https://github.com/freedoom/freedoom/blob/master/patches/sw11_1.png

chungy:
I think this is a good idea, and a great tribute to the man too.

PR #630, opened 2019-10-11
textures: Add a STARBR1 texture.

fragglet:
Weirdly there's a STARBR2 but no STARBR1, and there are also these
SW11_4 and SW11_5 patches which are not used in anything except for
in COMPUTE3 as part of a montage texture. It seems like a natural fit.


Proluiz:
there's no screenshot of something to we see it?

fragglet:
Here's what it looks like:

patchpatch

Compare to what STARBR2 looks like:
patchpatchpatchpatch

And STARGR1:

patchpatch

STARGR2:

patchpatchpatchpatch

PR #632, opened 2019-10-15
sprites: Replace player third-person sprites.

MatthewTheGlutton:
E/F sprites widened for better scaling to 1:1.2.

Antenna added to back plate as shown on title pic. Player-coloured light by artistic licence.


chungy:
This is pretty good work. It takes me a bit of getting used to the chubbiness compared to the older sprite, but I think I am ok with it.

I have a couple of critiques though:
1. the detailing on the gun is totally absent, just being black. It seems a little unusual to me.
2. More of the player sprites are based on the skinnier version and I think they should be updated also to reflect changes here.

MatthewTheGlutton:
1. Will do! I kinda assumed the gun was supposed to be a literal "black box" but I can detail in something more generically rifley.

2. I'll try for pain first, then the death sprites. The crouching sprites could definitely use an update as well...

I've tried looking at the sprites to see where I could cut something down to look more... triangular... but nothing really occurs to me that doesn't make me think "where are the muscles holding this up"... I'll keep an eye out while doing revisions.

MatthewTheGlutton:
Proof of life update: I think I've substantially completed the drawing for all the regular player sprites, including death but not including gibbing. Now to do some touchups on the crouching sprites...

fragglet:
Hey, just wanted to say that I built the latest version and tried it out watching some multiplayer demos and it looks really good. Looking forward to seeing the finished product.

MatthewTheGlutton:
I think I'm pretty much done at this point.

How do I squash all this into a single commit again?

MatthewTheGlutton:
While waiting for feedback I realized I should do one more arbitrary-ish change just so I can re-squash with a half-decent commit message, so I've addressed the PLYC*4*6 walking frames being noticeably brighter than the other angles.

chungy:
Looks all good. Thanks!

Issue #633, opened 2019-10-20
Does anyone actually use the wad-image targets?


chungy:
This issue ought to be directed partly at @selliott512 -- I know he uses it (he made it!), but I don't know if anyone else does.

I spent over an hour trying to get wad2image installed in a way that the make targets actually expect. This was complicated because it seems like installing wad2image like a normal program isn't actually supposed to work, and the make file heavily depends on putting it inside Freedoom's script directory (a solution I don't particularly care for). Even after I got something working with a combination of environment variables that are not obvious to me, the utility of it in Freedoom seems... limited. Perhaps a consequence of how I manage the repository, but even things like "compare the latest changed WAD" don't really appear to reliably work like it was expected to work, basing off of file system time stamps that are very suspect to change when checking out different points of the repo history, and not behaving as I would really expect it to behave.

In another case I actually expected it to have: create images of all the Freedoom maps, it seems to be totally lacking (I could script my way to getting that, but... eh).

With apologies to selliot, I think there was a good idea here, but it's buried under many layers of complexity and I am tempted to strip this out of the Freedoom Makefile.

Calinou:
I think generating images is better done using external tooling/scripts anyway, so I'd be fine with removing it. It's not strictly part of the Freedoom project per se.

selliott512:
I'm sorry it's been difficult to use. Maybe i can help. First, a few steps to get it going even if you already did:
```
type -p animate # In ImageMagick.
# An alternative to the following to avoid changing "scripts" is something like 
#   ln -s /home/sle/git/wad2image/bin/wad2image.py /usr/local/bin
# so is symlinked in the path. It can figure it out from there.
ln -s $HOME/git/wad2image scripts # Wherever you have wad2image checked out.

# A tree was moved, which this produces an animated GIF to show.
git checkout e8c621d9e50ca187cc74d2d1c335a0146879bbab -- levels/map05.wad
git reset levels/map05.wad # Modified, not staged, as if it had been copied in.

# Second to last example from "make wad-image-help"
make wad-image-diff WI_VERBOSE=t WI_GIF=t WI_SHOW=t WI_CMD=animate
```
The above should pop up an animated GIF showing that the tree was moved.

As to the motivation - I find working with binary files to be disconcerting since I'm used being able to "diff" changes. If someone claims they made a minor change to a level WAD how do you know what they actually did? Playing through the level seems time consuming. Maybe this is getting to be less of a concern as the level WADs are getting more stable. 

[...]

PR #637, opened 2019-10-27
Makefile: Explicitly define the shell (bash) and its env.

sirocyl:
This makes the "command: command not found" error no longer occur on
systems where the default shell is not `bash`, or does not support the
`bash` builtins. Additionally, a sane environment is set for safer
shell command use and scripting.

The environment settings for `bash` here do the following:
  - Terminate on errors, including signal `ERR` traps. (-e, -E)
  - Ensure that shell variables are set before use. (-u)
  - Print the shell command, and arguments, as they're run. (-x)
  - Fail on errors in any command in a pipeline. (-o pipefail)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
+SHELL=bash -Eeuxo pipefail
 VERSION=$(shell git describe --abbrev=8 --dirty 2>/dev/null || echo v0.12.1)
 WADS=wads
 ASCIIDOC=asciidoc


sirocyl:
Small aside - this does add `bash` (or a closely compatible shell) as a dependency, although that was already done with the use of `command` earlier - if this is not desired, the use of `bash` builtins should be eliminated.

fragglet:
Thanks! I learned a new Makefile trick with this PR :)

chungy:
I know it's a bit of my fault for not commenting when I saw the PR, but I was thinking of withholding this until I decided on #633 :)

tis is fine, though, it's a good idea to have it fixed regardless. Also I'm with fraggle, I learned something.

PR #641, opened 2019-11-03
c1m1: make door secret

Jayman2000:
On C1M1, there's a SHAWN1 door that leads to a secret. The door is now marked as
a secret, so it won't appear on the automap.


chungy:
Thanks!

PR #646, opened 2019-11-24
textgen: Tweaks and improvements to kerning table.

fragglet:
This improves on the existing rules and in particular resolves a TODO to add more rules for lower-case characters.


graphics/text/textgen
```diff
@@ -155,12 +155,14 @@ def generate_kerning_test(font):
...
 
-    cmd = font.generate_graphic(" ".join(pairs), "kerning.png")
+    image = font.generate_graphic(" ".join(pairs))
+    image.save("kerning.png")
 
 
 if __name__ == "__main__":
 
     font = TextGenerator("fontchars", kerning_table=FONT_KERNING_RULES)
+    #generate_kerning_test(font)
```
chungy: Commented code is useless code. It seems like you've used this to do some sort of manual testing. Maybe it would be worth extending the program so a program argument could trigger a test mode, or split out an independent textgen_test script?

PR #650, opened 2019-11-27
C1M1, MAP05, and MAP08: Bug fixes, minor decoration

ConsumingCritic:
For C1M1, I have added a caution bar to a high platform that contains the blue key where anyone unfortunate enough to not watch their step may fall down and have to go around to get up there again. It is decorative, but it also adds a sense of realism, and it can potentially serve as a heads-up to players about to run off the platform (just as it would hypothetically to anyone at AGM). I have also added similar caution bars to a staircase in MAP05.

Also, when I was testing the game on my Android device, I had recently noticed two areas that needed fixing. In MAP05, there is a secret area containing an automap where it is possible to collect the automap without causing the game to register the secret as being found. I have drawn a quadrilateral that is adjacent to the area to fix that. In MAP08, there is a wall that when used reveals a secret area containing a 100% health bonus. I do not know whether I had missed something important (i.e. a switch that allows vanilla access to it), but I discovered that the only way to reach the bonus is by jumping onto the platform. There is no way to access it using the vanilla Doom engine, so I lowered the front part of it to fix that. That should help make Freedoom more compatible with the original Doom engine.

C1M1
screenshot
MAP05
map05
MAP08
map08


Calinou:
You can access the ledge in MAP08 by straferunning (running diagonally), as you move faster this way.

selliott512:
I like the caution bars - making a very dangerous world just a bit safer. :-) 

That jump in MAP08 is a pain. I find myself attempting it multiple times, so l like the idea of making it easier one way or another.

Exceptional screenshots and explanation.

selliott512:
I'm not one of the project maintainers, but my $0.02 on the mechanics of git in case it's helpful:

You have a merge commit as one of your three commits which can be confusing to deal with. If your change gets merged as is then there will be a new merge commit that points to this branch which then has another merge commit in it. Since your merge commit is only meaningful in your repository if it were me I'd be inclined to `git reset --hard` just before it, `git cherry-pick` whichever commit is needed, `git diff @{u}` to make sure there is no unintended change, and then `git push -f` to update your feature branch.

Squashing is nice unless you think there will be a need to make a distinction between the commits after they are merged.

Again, just my opinions on git.

I tried it in GZDoom. Looks good to me.

ConsumingCritic:
@Calinou So that is what I was missing. The Doom engine was designed to allow running faster by moving diagonally. I knew it was possible in a handful of first-person shooters, especially older ones such as Star Wars Jedi Knight: Dark Forces II. I did not know, however, that Doom could also do that. I tested the level myself, as well as MAP19 with a similar secret that is said to be accessible by "straferunning", and I found that it is possible. You can disregard my change regarding MAP08. I would like to point out that not everyone knows the exploitation of this game mechanic, and if there is anything I would do to change the level, it is adding subtle hints to running diagonally to that secret area. Perhaps add arrows to the floor all pointing in one direction and stack them diagonally so that the player can get the idea of straferunning. Can we do something like that? I can do the work for you, but it does not hurt to ask.

selliott512:
IMHO your change is nice as is. It's already a secret area by being obscured by the panel. No need to also require trick jumping. I find trying to get the jump right frustrating.

ConsumingCritic:
> You have a merge commit as one of your three commits which can be confusing to deal with. If your change gets merged as is then there will be a new merge commit that points to this branch which then has another merge commit in it. Since your merge commit is only meaningful in your repository if it were me I'd be inclined to git reset --hard just before it, git cherry-pick whichever commit is needed, git diff @{u} to make sure there is no unintended change, and then git push -f to update your feature branch.

How do I do that? Sorry, but I am still quite new to GitHub, and I do need help.

selliott512:
Regarding git merge commits - probably don't worry about it for now. I looked through the various *.adoc files and I couldn't find any detailed guideline, so it's really just my opinion as someone who's not a project maintainer.

Briefly though, the procedure you can find many places online to "squash" commits will by default get rid of the merge commits, but again I wouldn't worry about it for now since the maintainers are quite capable of fixing it up however they wish. If they want you to do something I can try to be helpful.

PR #651, opened 2019-11-29
map11: Fix unreachable SKAG and monsters

selliott512:
Removed unreachable SKAG near the start as well as the bars in front
of it. Added a nominal health potion for symmetry with the pillar to
the right without overly affecting balance.

Near the end of the map there are six monsters on pillars that can get
knocked off. Once knocked off, either from player or monster
projectiles, they're very hard to kill making 100% kills difficult.
Teleport them to where the player can kill them when they get knocked
off.


selliott512:
If teasing the player with the SKAG was intentional then feel free to ignore this PR, but it felt like something that is now defunct to me.

With this change the REJECT lump is now 0 bytes instead of 38,504 bytes. All other maps have a non-zero REJECT lump except for C1M1. If you want I can try to figure out what to pass to ZDBSP to make it have a real REJECT lump, but then perhaps C1M1 should be fixed as well.

selliott512:
By the way, I'm fine with wad2image being removed, but I thought I'd share how I used it to show a change in this case:
```
wad2image.py --iwad ~/git/freedoom/wads/freedoom2.wad --show-cmd animate -d gif -vs map11-{old,new}.wad
```
After first copying the level to "old" and "new" files before and after the change.

Here's the [image](https://selliott.org/ptmp/map11-unreachable-items-e9sDCOhUxaew/map11.gif) produced.

Proluiz:
if i'm not wrong erick is going to soon rework on my version of map11, someone commited this version before i finished it, because it the skag is unreachable.

selliott512:
If I'm understanding correctly you had plans to make the SKAG accessible, but an early version of your MAP11 got checked in? Maybe you were going to add a teleport there?

Is there a MAP11 in progress? If so, this could be checked in as a temporary remedy, and then overwritten. Or it could be closed out.

If no MAP11 is currently in progress, then this PR can be merged?

With regard to the second part of this PR having to do with monsters falling off pillars - all I did is make the linedefs around the pillars teleports to a teleport up on the platform where the player is. Maybe if we end up closing this PR the owner of MAP11 can consider something like that?

selliott512:
With regard to the REJECT lump issue mentioned above I've since determined that ZokumBSP is more commonly used by the Freedoom project than ZDBSP. I've rebuilt the nodes and now REJECT is 38,365 bytes, much closer to what it was before. Likewise with all the other lumps. I've pushed it.

selliott512:
Rebuilt the nodes with ZenNode since that seems to be the correct node builder.

PR #654, opened 2019-12-14
Makefile: Add "rebuild-nodes" make target

selliott512:
Add "rebuild-nodes" make target to rebuild nodes for all levels that
match a specified pattern. Document "rebuild-nodes" in
BUILD-SYSTEM.adoc. Modified map06.wad so it adheres to the "no padding
between the lumps" constraint.


selliott512:
Recently it took my a while to figure out how the built nodes work in Freedoom, so I thought I'd add a make target and some documentation.

If there are any additional options that you like to pass to ZokumBSP just let me know and I'll add them, but as is it generates nodes that are almost exactly the same size as the existing ones.

For map06.wad there was a 24 byte region that said "DeuTex 4.4.902 ¸OJM 94"  as well as some 1 and 2 byte padding regions. I verified that the remaining 11 lumps are exactly the same after removing the non-lump regions. I wrote a tool called [wad-tweak](https://github.com/selliott512/wad-tweak) to filter out such regions as well as some other things:
```
wad-to-lump.py -vip map06.wad _standard_
```

chungy:
Hmm, interesting that you chose ZokumBSP. I've used [ZenNode](https://github.com/Doom-Utils/zennode) instead and indeed most/all the levels in Freedoom presently have ZenNode-built nodes.

ZokumBSP claims to be experimental which has drawn me away from relying on it for Freedoom, plus the lack of a top-level Makefile gave me doubts about its usability on Linux (I admittedly haven't downloaded nor tried to compile it). Do you mind shedding light on this?

selliott512:
To be honest I don't have a preference. I picked ZokumBSP because it was my best guess from looking at the git log where I noticed that in those cases where it was mentioned ZokumBSP was used more recently:
```
$ git log -10 -i --grep node -- levels | grep -iE '(^Date)|node|bsp' 

Date:   Tue Aug 27 12:16:54 2019 -0700
    levels: rebuild nodes and run fix-map-names
Date:   Sat Aug 4 14:19:23 2018 -0500
    -Rebuilt nodes and blockmap with ZokumBSP
Date:   Sat Jul 14 17:31:44 2018 -0500
    -All: Rebuild nodes and blockmap with ZokumBSP
Date:   Mon Jul 2 22:01:02 2018 -0500
    -Rebuild nodes and blockmap with ZokumBSP
Date:   Mon Jul 2 13:54:41 2018 -0700
    levels: rebuild nodes for maps that need them
Date:   Sat Dec 23 01:36:30 2017 -0500
    Also, rebuild these levels' nodes and reject tables with ZenNode.
Date:   Fri Nov 17 17:51:58 2017 -0800
    levels: rebuild nodes where necessary
Date:   Mon Aug 21 01:15:34 2017 -0400
    - Rebuild nodes and blockmap with ZokumBSP
Date:   Sun Aug 6 11:07:07 2017 -0700
    levels: ZenNode WADs with invalid nodes
Date:   Sun Mar 19 12:46:02 2017 +0000
    add a control sector and rebuild nodes, which would bloat the diff!)
```
If you want I can change the documentation and the Makefile to ZenNode which I've verified locally works just fine. I can also see that almost all of the levels were built with ZenNode last since running ZenNode on all of levels only causes 7 levels to have changes. I can amend and push force so there is one commit.

As to the top-level makefile - that seemed odd to me also, but ZokumBSP's makefile is `src/zokumbsp/makefile`.

Incidentally, I noticed that both ZenNode and ZokumBSP have a bug where they hang reading from stdin if stdin is not a TTY. It's probably not an issue since it will usually be run interactively, and if it's run on the build server stdin will probably be closed, but if you think it's a problem I can add `< /dev/null`. I hesitate to do that in case stdin is ever useful to it. Maybe there are cases where it asks questions, or it reads data from stdin.

selliott512:
I think I answered my own questioning by seeing that using ZenNode only changed 7 of the levels. I've updated the change to ZenNode and kept it in one commit.

By the way, if you would've preferred that I kept the original commit and add another so you can see the update more easily I'm happy to do that next time. I just think GitHub doesn't always squash on merge in a way that's helpful.

selliott512:
Rebuilt the nodes with ZenNode since that seems to be the correct node builder.

PR #656, opened 2019-12-22
Makefile: Warn if deutex does not support PNG

selliott512:
Check if deutex supports PNG based on its help output and fail the
build early if not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ OBJS=$(FREEDM) $(FREEDOOM1) $(FREEDOOM2)
 
 .PHONY: clean dist
 
-all: $(OBJS)
+all: deutex-check $(OBJS)
 
 subdirs:
 	$(MAKE) -C lumps/dehacked # graphics/text depends on generated dehacked files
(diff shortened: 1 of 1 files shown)


selliott512:
If you attempt to build Freedoom on Fedora 31 with the deutex in the repo the build passes, but produces invalid WAD files. The problem is the deutex in the repo does not support PNG. I think it would be helpful for the build to fail early in this case. Here's an example of failed build output with this change:
```
/bin/deutex does not support PNG. Try building deutex with the PNG 
libraries (libpng and libpng-devel or similar packages) installed. 
deutex can be downloaded from https://github.com/Doom-Utils/deutex.
The full path to duetex can be specified by passing 
DEUTEX=/the/path/to/deutex to make when building Freedoom.
make: *** [Makefile:50: deutex-check] Error 1
```
Admittedly the build displays many "PC90 could not find file" warnings, but that may be hard to understand.

Running the check by itself on the command line only takes 5 milliseconds on my system.
```
time (deutex -h | grep -qw PNG)
```

PR #658, opened 2019-12-26
c3m2: Add boxes of shells

selliott512:
c3m2 is quite ammo starved when the ammo conserving angle grinder in
the secret area near the start is not found. Add two boxes of shells,
and convert some loose shells to a box of shells for a total of three
boxes of shells.


selliott512:
This is admittedly a boring solution since perhaps additional weapons would be more interesting, but I figured it was a reasonable and safe incremental improvement.

I've checked in the Discord chat room and the consensus was that this map lacked ammo, so I added boxes or shells until it seemed right. An [animated GIF of the change](https://selliott.org/ptmp/c3m2-more-ammo-DQVmx4pSnRyj/e3m2.gif).

PR #659, opened 2019-12-26
c2m3: Remove extra exit

selliott512:
Remove the extra exit for c2m3 that was once a secret exit by changing
the extra exit into a secret area with a large health pack.


selliott512:
I think the extra exit is confusing. Even if we plan to rework this map this is as simple and easy change for now. An [animated GIF of the change](https://selliott.org/ptmp/c2m3-remove-extra-exit-vNQWGwfAjS0E/e2m3.gif).

I thought it would make a nice secret area, and a large health pack seemed like a reasonable payoff for finding it without disturbing the game too much, but I can change it.

MatthewTheGlutton:
As long as all exits are clearly marked as to prevent their inadvertent triggering I don't see the point of this. The second possible exit makes things more interesting than it does confusing, giving the player more options to explore (and sometimes a lifeline in some difficult monster/weapon mods where something can go horribly wrong halfway)

selliott512:
The current state of the map was not by design. It's a result of the maps being renumbered, and a secret exit being changed into a non-secret one as a quick solution. But that doesn't necessarily invalidate your point.

In pretty much every map I've ever played there has been one normal exit, and possibly one secret exit. So I guess it's a question of how valuable following that convention is verses doing something different. If the consensus is that having two exits is helpful I'll understand.

As to the how clear it's marked It's somewhat clear. It's a green portal looking thing.

selliott512:
As to inadvertent triggering - the player can see the main exit before coming across the secondary exit, so the player is likely to assume that the secondary exit is not an exit if the player is used to there being only one exit.

selliott512:
I see now that C2M9 has the same issue - it has at least two normal exits one behind a blue skull key door, and one behind a yellow skull key door. Maybe it's for the same historical reasons.

A possible compromise - the exit I removed could be changed to a teleporter teleporting close to the real exit.

PR #662, opened 2020-01-03
sprites: make the dark soldier's death more violent

boogiebogus:
I thought that it didn't make much sense that the dark soldier just didn't bleed at all when it died, so i took some time to change up the death sprites. Now its head explodes, and arm falls off, and a big hole opens up in its torso (no guts, just blood)


NickZ:
Thanks, this looks a lot better.

PR #663, opened 2020-01-21
sprites: new Lavamander (Baron of Hell)

Talon1024:
Lavamander sprites by raymoohawk and HorrorMovieGuy
From this thread on Doomworld Forums: https://www.doomworld.com/forum/post/2045211


boogiebogus:
don't forget to update the build cfg for offsets.

either way, i agree with this change.

ObsidianPlague:
Doesn't seem fitting for a baron replacement.
Would fit more for a Hell Knight replacement, though.

Calinou:
Wouldn't it be difficult to see on certain backgrounds given its colors?

MatthewTheGlutton:
Aesthetically it seems to make more sense for the knight, come to think of it...

chungy:
I'm not really sure the dinosaur aesthetic fits Freedoom. I'm just not feeling it.

StarmantheBlaziken:
> I'm not really sure the dinosaur aesthetic fits Freedoom. I'm just not feeling it.

Even though you are one. :P

StarmantheBlaziken:
But I was going to come in and say if you had not noticed, yeah I think it could be the Pain Bringer replacement if it does land.

MatthewTheGlutton:
tbh I kinda see what Chungy means...

What if we got rid of the tail and it's just, like, this big red goblin guy?

boogiebogus:
hmmm... now that doesn't sound like a half bad idea.

selliott512:
Nice work. Really detailed.

Currently in Doom and Freedoom the Hell Knight and the Barron of Hell are the same monster with different color schemes, which I've gotten used to and like in some ways, but maybe that's not a good reason to not have another monster type.

I can understand being hesitant to introduce a dinosaur like monster, but at the same time monsters in both Doom and Freedoom are somewhat eclectic.

If only there was an existing idea that there was a monster that lived in lava areas then this would be perfect. Still nice though.

MatthewTheGlutton:
Looking at BOSSA1 old and new I'm going to agree more with the opinion that this should be the baron replacement rather than the knight.

The hell knight is more of an elite shock trooper mook while the baron is this crazy roid-raging *thing* that has a huge amount of heft and takes a lot more power to stop.

Meanwhile the lavamander looks about twice the size of the pain verbers...

NickZ:
I do like the lavamander; he looks appropriately "evil alien" for freedoom. He has a certain personality to his walk cycle, like he's a lumbering oaf who is very angry that you woke him from his slumber. I think this would make a good addition to the project.

doomfighter33456:
i agree with you

HorroMovieGuy:
I feel like I should post this here, just as documented proof of permission from Raymoo that we are allowed to use this monster, should it be accepted in the base branch:
opera_qNhb73jn7w

mc776:
Changing my initial :+1: to a :-1: because the more I look at the gold recolour of the pain lord the more I'd want to see that rather than this.

mc776:
I was curious about how much has changed since the most recent release and I just noticed the lavamander had been committed and then reverted previously.

For the record, that seems to be inadvertent and not the result of previous discussions: https://github.com/freedoom/freedoom/pull/699

shinobody:
I oppose the change on two counts.

One, Lavamander and Nukeptyle look cool, but they do not really tell an interesting story beyond 'lizard monsters'. Pain Lord and Pain Bringer show the idea of this reptile/bird people in an organic power armor. Their cannons are clearly separate organic beings who breathe plasma/polaric bolts out of their mouth (they have entire faces and even emote!). In their death animation you can even see that the mouth-arm cannons have _their own separate blood color_ which is a such a cool lore detail. It even compounds storytelling since Berserk Pack is the same kind of mouth-arm entity, so it implies that whatever it is, it's similar to the thing that the Pain Lords are using.
And we can go further, as we can warrant a theory that the alien creature inside of the Assault Tripod is what the Pain Lord aliens look like without any armor.

Second count is colorblind accessiblity - Lavamander and Nukeptyle are red and green, and red-green colorblindness is the most common type of colorblindness out there. The only really visible difference is crest on their heads, which I think is too little for difference between a Hellknight type and a Baron type.
In contrast, the new Pain Lord and Pain Bringer sprites (https://github.com/freedoom/freedoom/pull/834) are silver and gold - white and yellow. Difference between white and yellow will be visible under almost every type of colorblindness (and there's some difference so even people with achromatopsia, completely no color vision can distinguish them).

So in interest of storytelling and accessibility I vote no and recommend instead the gold Pain Lord recolor.

mc776:
I agree in principle, though I think the PR so far is *only* intended to use the lavamander for the baron, not both lavamander and nukeptile for baron and knight - which of course is even worse from an implied worldbuilding/UI signposting point of view since we lose the indication that they're the "same" and thus have the same projectiles and don't infight.

Xindage:
seems like this sprite was already repurposed for zauberer, so we're closing this.

PR #664, opened 2020-01-21
New DSSECRET

ObsidianPlague:
Something that could fit in more than the original doom secret sound.


MatthewTheGlutton:
What's the source of the sound?

ObsidianPlague:
As in what "instruments" the sound uses or where the sound is stored?

MatthewTheGlutton:
How it was made, who recorded it, etc.

There've been some rather unpleasant controversies in the past, so it's better to be safe than sorry.

ObsidianPlague:
I gathered up the original audio sources from Freesound, which were under CC0. I merged them together by pitching them both down, making the tempo of the church organ part faster, adding some reverb, some echo, resampled it to a lower frequency, and exported it to 8 bit.

The original audio sources are from...
https://freesound.org/people/giddster/sounds/437337/
https://freesound.org/people/liezen3/sounds/477373/

selliott512:
Cool sound.

I wasn't familiar with this lump, but a quick Google revealed "If a PWAD contains a sound lump called DSSECRET, it will now be played along with a message when the player finds a secret area.". Do you know if it works with IWADs too?

For the new files to be used I think buildcfg.txt will need to be modified as well. I only end up with a DSSECRET lump in freedoom2.wad if I modify buildcfg.txt in the following way:
```
--- a/buildcfg.txt
+++ b/buildcfg.txt
@@ -286,6 +286,7 @@ DSCYBDTH
 DSSPIDTH
 DSHOOF
 DSMETAL
+DSSECRET
 
 #ifdef PHASE2
 DSDSHTGN
```
The commit message is just "Dssecret". Ideally it should be a bit more.

fabiangreffrath:
There is a bit of noise towards the end when the sound fades out. Maybe this could be cut out and the sound a bit shortened,

fabiangreffrath:
Now this PR mixes up a new DSSECRET sound with a completely unrelated music change for MAP22. :-/

ObsidianPlague:
Yeah unfortunately I have no idea how to fix that, sorry
I can't have more than two forks. (i did another PR for a music change)

fabiangreffrath:
You don't need to create more than one fork, you can create several branches for a single fork.

ObsidianPlague:
...Now i look silly. Is there a way to update a PR to pull from a branch instead?

PR #667, opened 2020-02-15
New Map07

Mortrixs19:
A new map07 made by me, to replace the current dead-simple clone map.

[MAP07.zip](https://github.com/freedoom/freedoom/files/4250845/MAP07.zip)

_(I'm still not sure how Pull Requests works, ok?)_


NickZ:
Hey, nice map!

One thing that needs fixing:
On both my playthroughs, I was unable to get 100% kills. It seems like this monster box doesn't ever activate? 
Screenshot_Doom_20200220_140414

Suggestions:
* Add a secret here? Perhaps a soulsphere, since health is scarce in the level
Screenshot_Doom_20200220_140718

* Since it's map07, I suggest also doing something with the technospiders, as is tradition. Perhaps the current exit switch could transport you to a pit that requires you to kill them all before exiting?

Mortrixs19:
Oh, my mistake with the teleporter trap. I forgot to re-add the tag to make the monsters teleport.

I never had problems with the health in the map, but I guess for no-so experienced players a bit more of health could work, I added a soulsphere secret.

[MAP07.zip](https://github.com/freedoom/freedoom/files/4245157/MAP07.zip)

NickZ:
This thing is blocking access to the secret? I can't seem to get access to the soulsphere without jumping over it or noclipping.
image

Mortrixs19:
Fixed the soulsphere secret, and a few minor things.

I guess the only thing left is to change the name of the level and change a few things in the Text screen before the map.

selliott512:
Nice map! My thoughts:

There were times when it was not obvious to me where to go next. That's not necessarily a criticism. Some maps are less linear than others. Figuring them out is part of the challenge.

I was able to play it on UV-MAX and get 100% kills without too much trouble. I only found 1 of 6 secret areas (the soul sphere in the last picture in the comments), but I didn't look very hard. I just happened to notice that switch. Getting the polaric energy cannon that I saw in the secret area that I  didn't figure out would've been quite helpful.

As you probably know there's another MAP07 contender written by GooseJelly with fixes from DOGB01. You can find it by searching for "map07.wad" in the general Discord channel. I'll avoid picking a favorite, but I will say you get points for submitting a pull request. Things get lost and forgotten otherwise.

IMHO this is a nice improvement of the existing MAP07. The existing MAP07 is too hard, and too similar to Doom 2.

As to the mechanics of the pull request - it may be helpful to squash the commits into one, but the person who applies it will tell you what matters to them.

selliott512:
Some technical (as opposed to game play) feedback:

1) Did you mean to add file "levels/D_MESSG2.mid"? It seems out of place. Would it be better to overwrite "musics/d_map07.mid"?
2) The file mentioned in 1) is also in lump "D_SHAWN". Usually FD maps only have all 11 lumps (see the "Node builder" section of "BUILD-SYSTEM.adoc" for details). Also, usually the first lump is the map name.

I don't recall what the music was, but I'd move things around to address the above however you think makes sense, run "make" and then see if the music is as you expect just by running the built "freedoom2.wad" (without passing "map07.wad"). That's assuming you have a build setup, and doing so works for you.

Sorry if that's pedantic, but maybe it helps others.

Mortrixs19:
My intention was to change the current MIDI for something more fitting (The old D_MESSG2 from 0.11.3). I wasn't sure If I had to change the d_map07.mid file in the ''musics'' folder, but now it's changed with the old Map26 midi.

I added the music as D_SHAWN in the wad file since that's the common way you replace music in levels, just so people can play and review the wad with the intended music.

Not sure who the author is though, I took the Midi from an old version of freedoom, so I assumed is free to use.

chungy:
I need to also run this by Blastfrog who is presently in charge of early Phase 2 levels

chungy:
In general you've done a good job, but for full clarity, there are some issues to keep in mind:

* Your commit history is extremely messy, with multiple commits to delete a file and add a new one back in separate commits, and generally the long stream made here as you've edited and re-edited. This can all be collapsed into a single commit (this is easy for us as maintainers to do too, granted)
* Surprise music lumps either in the musics folder or in a level WAD for a pull request only about a level is not the best idea.
* Yes, in normal PWADs placing music lumps in the same *.wad as a level is standard practice, but Freedoom needs to be more strict and careful: level WADs must only contain the level, nothing else.
* Reintroducing music removed from older versions of Freedoom requires a lot more discretion and discussion; it has often been removed for a reason.

NickZ:
For reference, D_MAP26 (aka D_MESSG2) was added in this commit: https://github.com/freedoom/freedoom/commit/d4136ca71165f887182a3ce0db0c1123905553e1#diff-7d7854dbe7f5a77f30fa4622af94197d
And was replaced in this commit: https://github.com/freedoom/freedoom/commit/7aa680ea051ac6ce1f4a72dcf2e952fe9ad62b1b#diff-7d7854dbe7f5a77f30fa4622af94197d

So it looks like it was originally introduced to get rid of music that had potential rights issues, and was just replaced with something that fit better for that map. I don't see any issues with this music.

For reference sake, it was originally made by Rakohus.

Mortrixs19:
Yeah, I'm sorry if the PR was a pain to review, I'm still figuring out how GitHub works :/ 

Anyway, I finally removed the midi from the wad file and also removed the midi from my branch, I guess that should be enough to clean this mess, but I will still request for the music to be changed in the future, since the current track for map07 doesn't fit the map.

NickZ:
this was merged in 2de1ff1bce864523dfc6a1f8a526db6a76f5fea1.

PR #671, opened 2020-02-27
Add Continuous Integration

NickZ:
Uses Github Actions to run `make` and upload the build artifacts

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -9,21 +9,48 @@ jobs:
     
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Get full repository history
+      run: |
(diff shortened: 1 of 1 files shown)


chungy:
Two questions:

Why not install deutex via apt instead of compiling it? Potentially building the master branch could be broken and look like a Freedoom failure.

Also, where exactly are build files being uploaded? And why not use `make dist` to make uploaded files?

NickZ:
the deutex in bionic segfaults when running --help, which causes the build to crash. I'm not sure if the fix made it into bionic-updates, but its not available in the base bionic image that github has. I've pinned the build from github to v5.2.1.

I went ahead and I changed it so it uploads the build artifacts from `make dist`. They're uploaded to github itself, you have to go into the details of the action and look at the artifacts button at the top right.

PR #672, opened 2020-02-28
Add snap packaging

NickZ:
This adds snap packaging to freedoom.

The snap package that this builds is bundled with gzdoom with some sane defaults (texture filtering off, WASD controls, no jump, etc.) to provide a good "out-of-box" experience. It's designed to act as a stand-alone app.

To build it, make sure you have snapcraft installed by running `sudo snap install snapcraft`, then run `make snap`.

To install the snap package after its finished building, run `sudo snap install --dangerous freedoom*.snap`


Makefile
```diff
@@ -270,3 +270,6 @@ uninstall-freedoom:
 install: install-freedm install-freedoom
 
 uninstall: uninstall-freedm uninstall-freedoom
+
+snap: force
+	snapcraft
```
chungy: Make sure the file ends with a terminating newline.

snap/local/config/gzdoom.ini
```diff
@@ -0,0 +1,88 @@
...
+DPadDown=invuse
+DPadLeft=invprev
+DPadRight=invnext
+Pad_Start=pause
+Pad_Back=menu_main
+LShoulder=weapprev
+RShoulder=weapnext
+LTrigger=+altattack
+RTrigger=+attack
+Pad_A=+use
```
chungy: There are a lot of keys in this file that seem like they are better left out, keeping only the ones changed from GZDoom's default. I assume the port can handle a partial config file?

snap/local/config/freedoom.wrapper
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# copy default settings if they don't currently exist
+if [[ ! -f "$SNAP_USER_DATA/.config/gzdoom/gzdoom.ini" ]]; then
+    mkdir $SNAP_USER_DATA/.config 2>/dev/null
+    mkdir $SNAP_USER_DATA/.config/gzdoom 2>/dev/null
```
chungy: Combine these two lines by using `mkdir -p`, and don't redirect stderr.

NickZ: Can't do `mkdir -p` here, permissions issues with snap containment.

snap/local/config/gzdoom.ini
NickZ: Unfortunately, it doesn't work like that with binds; if you provide a binds section and you don't have stuff bound, it will not use any of the defaults

Calinou:
Those default configuration changes are interesting. We could port them to the Flatpak if we can get GZDoom to work well with a "partial" configuration file :slightly_smiling_face:

snap/local/config/gzdoom.ini
chungy: That's too bad. Sounds like a bug to me, but oh well.

snap/local/config/freedoom.wrapper
chungy: Do you mind elaborating on this?

Also, I thought of something else: is `$SNAP_USER_DATA` the same as `$HOME` or is it some kind of special sandboxing environment? I'm thinking it is possible that ~/.config doesn't actually apply universally (since `$XDG_CONFIG_HOME` is what should be used...)

NickZ: Basically, because of the apparmor profile that's applied to the snap, attempting to run mkdir -p with an absolute path will fail to run because apparmor will not allow it to even stat certain directories in its path. Trust me, I've torn my hair out over this enough times :P 
 
`$SNAP_USER_DATA` is a special sandboxing environment, yes. It's where all the snap user data gets put instead of `$HOME` (because it doesn't have permission to write to hidden files in the home directory).

NickZ:
GZdoom works fine with partial GENERAL config entries and partial CONSOLE
entries, but not binds.

NickZ:
I've decided to switch to using [LZDoom](https://zdoom.org/downloads) instead of GZDoom. LZDoom is based on an older version of GZDoom, but it is being updated with patches from upstream and maintains support for older hardware that does not support OpenGL >=3.3, which includes the raspi3 and raspi4. I've tested the snap out on both raspi3 and 4, and they run well when the GL drivers are enabled.

After testing the functionality of the snap store auto-builds, I've decided that the snap packaging should just be in its own repository. We will want to use the snap store auto-builds because trying to compile a snap ourselves for multiple arches would be a herculean task (I've attempted this myself with other snaps, there's no cross-building to be had and building on ARM is a pain). Unfortunately, the snap store autobuilds will only build from the latest `master` branch, and will not allow you to specify any tags, so anything pushed to master will be built. So, if we just move the snap packaging into a new repository, we can just push to master every time we want to tag a new release.

@chungy Can you please create a "freedoom-snap" repository in the freedoom org?

chungy:
Well, if it's going to be a separate repository, I assume this PR is safe to be closed. I'll create the other repository.

PR #677, opened 2020-04-14
Less annoying Map22 music

ObsidianPlague:
I made map22's music sound less horrid to listen to. Hopefully this works as a replacement?


Xindage:
i've heard the music and it's definitely better, gonna commit them

PR #680, opened 2020-04-19
New C1M9

Mortrixs19:
New C1M9 made by me to replace the old one for being too plain and boring for a secret level.

[C1M9.zip](https://github.com/freedoom/freedoom/files/4500283/C1M9.zip)


fabiangreffrath:
This pull request does also replace `map07.wad` and its music `d_map07.mid`. Is this intended?

selliott512:
Fun map. It took me a little over 10 minutes to play. I didn't find any secret areas, but I could've looked harder.

It looks like you've used your master branch again which is why the MAP07 change is still here. One way to fix this is to create a new branch, but you want it based off of latest Freedoom master, not your master, submit a new pull request and close this one. Or you could fix up your master branch so it only has the change you want.

Eureka's Check / All says "1 unclosed sector" and "1 mismatched sector". Probably no big deal, but maybe easy to fix.

Calinou:
It's a fun map! It took me about 8 minutes to complete it on UV skill. I have a couple of comments:

___

Floor texturing looks a bit too basic and repetitive here. Not sure exactly what can be done about it.

Screenshot_Doom_20200424_225405

___

Not sure if this is a secret, but the wall texture here appears to be misaligned:

Screenshot_Doom_20200424_224939

___

This area can't be escaped (or at least, I didn't find a way to), yet you can have a whole minute of radiation suit when you reach it. This means the player can be stuck for a while before dying.

Screenshot_Doom_20200424_225009

___

The wall textures are set to be unpegged, which causes them to not scroll while the traps are being revealed. This looks a bit strange in action, is this intentional?

Screenshot_Doom_20200424_225039
___

There should be an Exit sign above this door:

Screenshot_Doom_20200424_225204

selliott512:
After playing few more times I was able to get my time down to 6:45, but only with saves, and only after studying the map carefully in a level editor to understand all of the secret areas, and to figure out the best path.

I think the secret areas are relatively hard to understand particularly by the standards of C1 maps, in part because they seem to require some doubling back, but if the reasoning is that secret maps should have more challenging secret areas I can understand that.

If you want to fix your master branch so that it only contains the latest commit one way is to drop your fork, recreate it, and then recreate the commit. But I'm not sure what effect that would have on this pull request. If you want to preserve your fork you can probably do so with the following steps:

1) Make sure you are on your local master branch, and that your local master branch is up-to-date: with respect to your remote master branch in your fork (in "origin"):
```
    git checkout master
    git status -sb
```

2) Add a remote to reference the original Freedoom as "upstream", and then fetch the latest remote branches for it:
```
    git remote add upstream git@github.com:freedoom/freedoom.git
    git fetch upstream
```

3) Make sure you only care about the latest change in your local master branch (your map07.wad change was already merged):
```
    git log --name-status upstream/master..
```

4) Reset your local master branch so it matches latest Freedoom master.
```
    git reset --hard upstream/master
```

5) Cherry pick just the lastest commit from your remote master branch (the one on "origin" in your fork):
```
    git cherry-pick origin/master
```

6) Make sure you have just one new commit that Freedoom doesn't have, and that it is the correct commit:
```
    git log --name-status upstream/master..
```

7) Push force to make your remote master branch like your local master branch:
```
    git push -f
```

If you don't want to do any of that the Freedoom maintainers will probably figure it out.

In general it's easier to create a separate branch for each pull request so you don't have to worry about them interfering with each other, so you can do that next time if you want.

mc776:
Should this be closed given #703?

Issue #685, opened 2020-05-07
Is it worth turning the boss brain texture into a full height lava-fall?


Blastfrog:
While I am generally very adverse to anything that could theoretically break compatibility in even the strangest, rarest and most unusual use-case scenarios, I'm wondering if this is one instance where we can make the sacrifice for compatibility in order to gain a huge benefit.

As far as I know, the only thing that changing a wall texture's height could break is stair-building, as it refers to the texture's height when determining how far to raise the floor. However, who is honestly going to use the DBRAINx texture to make self-building stairs in a PWAD? They wouldn't even be useful to climb, since the max climbing height is 24, but DBRAINx is 32 tall.

Since the texture already depicts lava, and there is no actual lava fall texture, this would be a huge gain for level designers. I'm not trying to open a pandora's box here to further theoretical compatibility breakage, but I think this is one case where we can make the sacrifice.

fragglet:
Sounds good to me. 

> As far as I know, the only thing that changing a wall texture's height could break is stair-building, as it refers to the texture's height when determining how far to raise the floor

I don't think this is true - step size is determined by the linedef type, and different linedef types raise different stair heights. So even this concern shouldn't be an issue.

Blastfrog:
> Sounds good to me.
Cool, glad you approve. I figured since there are so few hardcoded animated texture definitions, and this one already depicted lava, that it would be perfect to make it more useful as a full size lava fall, and any traditional use of the boss brain texture would still look fine.

I think I'll get started on a lava fall texture immediately, and when finished will submit it. I figure then it'd be a good idea to make an announcement on the forums so mappers know it's available.

I think it'd probably be a good idea to use new patch names, however, so that any PWAD with custom texture definitions that may use the boss brain patches won't have their intended appearance ruined. This admittedly would be a rare edge-case, but I figure better safe than sorry.

> I don't think this is true - step size is determined by the linedef type, and different linedef types raise different stair heights. So even this concern shouldn't be an issue.
Oh, whoops. I swear I read something like this on Doomworld forums years back, in regard to differing step texture heights between Doom 1 and 2. To be honest, for all of my map editing over the years, one thing I have never done was make self-building stairs.

And on that note, I notice that Freedoom phases 1 and 2 reflect Doom 1 having 8 tall steps and Doom 2 having 16 tall steps. If gameplay behavior isn't actually an issue, then perhaps we should consider just using the Phase 2 versions in both phases?

[...]

PR #686, opened 2020-05-09
Changed DBRAINx textures to a lava-fall

Blastfrog:
Height has been extended to 128, and new patches have been added (LFALLx) and are used instead of RWDM11x. I am deliberately leaving the original patches as-is, for the sake of PWAD compatibility.


fragglet:
Great! This fixes #685.

PR #689, opened 2020-06-22
New Phase 2 MAP26

guynamederick:
A new MAP26 by DOGB01, went through a month of changes and it appears to be final now. It has been tested with Chocolate Doom 3.0.0 and 100% stats should be possible on all difficulties. ZokumBSP 1.1 was used to rebuild the nodes.


MatthewTheGlutton:
Definitely liking the first 30 seconds or so of this better than the current map26...

Calinou:
I just played it through on skill 3. It's pretty good! It took me about 15 minutes to complete from a blind run.

Here are some comments:

___

The ceiling texturing seems a bit basic/out of place here. Maybe it should use a darker texture?

Screenshot_Doom_20200704_140438

___

These barriers are nonsolid, which mean you can run into them and block your view completely. It's a bit strange. Maybe these should be solid?

Screenshot_Doom_20200704_141049

Screenshot_Doom_20200704_141056

___

Nice lighting effects, courtesy of GZDoom :slightly_smiling_face:

Screenshot_Doom_20200704_141340

MatthewTheGlutton:
I'm in favour of leaving the final encounter as is - for most of us "properly" defeating the enemy before moving on is its own reward.

guynamederick:
Small update to MAP26, addresses the ceiling texture and cages not being impassable as pointed out by Calinou. I changed the flying skull enemies to serpents at the east blood room as Blastfrog pointed out how they are not able to reach the player because of impassable lines. As for the final encounter, I am not changing that as there is nothing I can do about it in vanilla and like MatthewTheGlutton said, let the player feel the satisfaction of doing it.

PR #690, opened 2020-07-15
Map11: Properly mark lines as not on map

perryprog:
Since it's possible to get a map revealer near the end of Map11, it was possible to see a few lines that were erroneously marked as part of the map.

This is my first time modifiying a WAD, so I'm not sure if I've missed any steps. I don't believe I needed to rebuild the nodes considering what the change consisted of, but I could be wrong.


perryprog:
Update: Oops, just realizes SLADE will remove the NODES lump even if your changes wouldn't have affected it. I'll have to set up a node builder, and I'll then update this PR sometime later today.

Xindage:
we have conflict in this PR can you do a checkout of your file perrypog?

PR #691, opened 2020-07-28
patches: improve Plutonia brick textures

Talon1024:
These should look much better, now that they are less rectangular, and have more depth


MatthewTheGlutton:
Does this affect the tiling and alignment of existing maps? Some of the brick lines are quite different.

Talon1024:
It shouldn't be too different. The biggest difference I see is with the 8-row patches, which are not in the Plutonia IWAD, and are only used in FreeDM, and possibly other Freedoom maps.

Xindage:
they indeed look differently as it has depth now. Some screenshots of freedm map18 that use some of them:
image
image

Looks good to be merged.

Xindage:
So far this one have been in experimental for a time, and i played this map over some times and the textures looks good, going to merge this.

PR #692, opened 2020-07-29
E4M2: Unmark teleport destination as multiplayer

### LINKED ISSUE #652: C4M2: Thing 683 (Teleport Destination) is marked Multiplayer-Only
This teleporter that opens doesn't work as a result:

DOOM0000

perryprog:
This gets rid of a multiplayer flag that was set on a teleport destination which was supposed to be accessible via Line #4047. When previously nothing happened on a WR, it now teleports the player as expected.

Two things of note, however: there are currently two teleport destinations near each other, each one serving a similar purpose. I'm not sure why they aren't the same ones, but in case there was something I was missing, I did not merge them.
Secondly, I'm wondering why have the teleports at all. Why not just have a WR trigger that would reset the "trap"? The trigger for the bridge lowering is already W1, so I imagine it would be fine. As I have very little experience (read: a couple hours) with Doom map editing, I'm not sure if there's some sort of logistical challenge that the teleports get around.

Regardless, this PR is just a simple bug fix, and I do doubt that this particular part of the map really warrants any more change.

Screen Shot 2020-07-28 at 9 42 57 PM
Screen Shot 2020-07-28 at 9 43 33 PM
(The second image is the similar teleport I mentioned that uses a different, but nearby, destination as the one fixed in this PR.)

P.S. Seems like my branch name was late to the chapter -> episode party :)

Edit: Closes #652


Catoptromancy:
It does not seem to break routing or anything else and looks like it would only enhance deathmatch games. It changes speed of player movement in area. Especially if bridge is down.

That bottom teleport was added so players could come back to area later without being trapped. 

Floor there raises for some reason, that might be an actual bug. And map has 17 deathmatch starts, vanilla needs 10.

Xindage:
we have conflict in this PR can you do a checkout of your file perrypog?

perryprog:
@Xindage resolved conflicts for both PRs. Thanks for the reminder, I kept meaning to get around to fixing those.

Xindage:
are you sure that you checked what changed in the master e4m2 and your version of e4m2? just asking man because i thing that catoptromancy inserted changes in e4m2 that's why it was conflicting to begin with.

perryprog:
I just checked out the master version and redid my changes according to my original comment, so it'll be up to date with any changes. I also checked with bspdiff and the BSP trees of both PRs are identical to that of the master versions.

PR #697, opened 2020-08-10
Give MAP31 normal and secret exits

kitchen-ace:
This changes the old exit to be the normal exit, and the old fake exit
to be the proper secret exit. Access to the secret exit requires
getting all the secrets in the map (i.e. telefragging all the Assault
Tripods/Cyberdemons).

I also fixed most of the switches not visually switching in vanilla.


kitchen-ace:
I couldn't find any guidelines on making changes to maps like this, so I just went ahead and did it. Hope that's OK. 

If it's better that the fake exit is kept, I can rework it.

PR #703, opened 2020-09-22
New e1m9 by mort.

Xindage:
This new map is to replace the actual e1m9 becuase it's too bland and this one is better in matter of gameplay.
It's beatable but still require feedbacks.
There's some know bugs: Misalined walls and inescapable pits.

This PR obsoletes #680
Edit: Mort is unable to take care of the map right now since his computer is broken, so i'll lead this pr and map for a while (with mort) until it's good to go in freedoom.

Post bugs and/or suggestions here, i'll be fixing them.


mc776:
I've just done an extremely quick godmode playthrough of this map, and an *attempt* at the current E1M9.

The new map has lots of alternate routes, lots of fun stuff to find, overall ammo/armour balance seems a bit generous but that's *rewarding the player for finding the secret* so it's all good. Monster and item placement feels natural but still challenging. Really nice flow, only felt a couple times I needed to backtrack but even those seemed avoidable.

In contrast, the current map starts you off with an arbitrary sudden appearance of monsters as soon as you move, has multiple unmarked key doors right by the start, unintuitive player-blocking lines,... I gave up. These are all issues that can be solved but the overall flow already felt like constantly running into dead ends and going back.

This PR is an unmitigated improvement.

PR #705, opened 2020-10-11
MAP12: More switch offset fixes

rueter37:
Some switches still have wrong offset values. This PR fixes these:

- Switch on the back side of the yellow key bars
- Switch on the back side of the small blue key bars
- Switches on both sides of the large blue key bars
- Switch on the back wall of the small storage room in the southeast of the map


rueter37:
This PR is obsolete now because MAP12 has been officially updated by #746.

Catoptromancy:
The map was only made vanilla, those issues are probably still there.

rueter37:
Can confirm that the issues are still there, except of the spot where the large blue key bars were, because this room has been rebuilt.

PR #706, opened 2020-10-19
Improved Switchs

Xindage:
DPstudio improved the switchs adding depth and glowing to them.

PR #711, opened 2020-12-06
DPstudio switchs, revision 2

Xindage:
Sorry for opening 2 PR, the problem is that I'm redoing my OS and things got break, please close the #706 on merge of this.


rueter37:
I really like the new switches! But I noticed two oddities:
- The recent changes are not applied because of the filenames being uppercase. You still see the switches with the lights "leaking".
- On the recent variant of the switch with the red and green light, the switch handle always points down now. The older variant of the switch has the handle pointing up if the switch is set to green.

Xindage:
Thanks i'll revise the commit.

Xindage:
It's working as intended now.

AXDOOMER:
Looks great!

PR #716, opened 2021-01-16
sprites: center trilobite death alignment on body mass.

mc776:
The sprites have this big spurt of blood going off to one side which meant that centering on the sprite centre made the entire body shift the opposite direction.


mc776:
Confirmed no conflict with the newly added imp gib sprites.

Xindage:
Thanks for the confirmation.

Issue #721, opened 2021-03-17
Problem with C3M7 solo-net and multiplayer


Hector577:
I was playing freedoom phase 1 co-op. When my buddy and I reached the blue door in C3M7 and climbed up the stairs we found four vases that block our path. And it's a problem because there's a switch there that is necessary to proceed. 
![Captura de pantalla de 2021-03-15 23-43-09](https://user-images.githubusercontent.com/80479550/111409368-70c65b80-869c-11eb-9eba-da5b80fc2752.png)

Xindage:
confirmed, it happens in 2 other places as well, i'm going to fix them after cheking an older version of this map.

PR #724, opened 2021-03-20
fix for #721

Xindage:
I just removed the pillar that appear in multiplayer, still this map has other bugs that need a better editing of the map to fix.


Xindage:
incompatible anymore, closing

PR #725, opened 2021-03-20
New midis for freedm pack.

Xindage:
Every new mid will be put there on this PR, wait confirmation to commit.
Note: Every mid here was made by korp.


AXDOOMER:
In my option, rather than replacing dm06, dm07 and dm08, they should replace musics with a lower bpm because they feel too relax for fast-paced deathmatch maps. Good candidates to be replaced by them are dm22, dm26 and dm30. 

dm22 <--- would be a perfect fit for single-player missions, but not fast paced deathmatch (still, it kinda fits with the current map, but it's on the list of maps to be replaced)
dm26 <--- this music is not suited at all for deathmatch (music is way too relax, doesn't convey a sense of action)
dm30 <--- bit too relax also, bpm not high enough

Replacement musics should also be matched to a map if the map ambiance is a good fit with the replacement music. For example, I don't think the suggested replacement of dm08 in this PR is a great replacement for the current dm08, as it's too slow bpm for a small map that's designed for highly fast-paced deathmatch, and it also would fit better with a map that would be in space, because that's what the music conveys. Such music with a low bpm are better suited for bigger maps.

A good example of a music that fits well with its map would be dm17. The music really fits with the ambiance of the map (a cave) and there's a really high bpm section, which is a good fit for a small map like it. 

Other potential candidates that could be replaced are: dm20, dm21, dm23. Music slots that are duplicate of others should be replaced first because it's better to have original music for each map (I didn't check for them, so I haven't made a list).

PR #728, opened 2021-03-30
Finished raymoohawk Octaminator by Horrormovierei replacing revenant.

Xindage:
This commit follow up with a new sounds for the punching.

image


northivanastan:
A couple of first impressions:

1. These sprites are really big in comparison to the Dark Soldier and even the Revenant. It's hard to get used to.
2. I think some new sounds are needed, at least for idle. Something more robotic/mechanical than the demonic chanting that the Dark Soldier does.

Xindage:
i have my own complain about the height only, because the hitbox of him is ok

mc776:
SKELA1D1
Doom2: 41x79
Darksoldier: 40x60
Octaminator: 54x82

Actual hitbox: 40x56

Octaminator is way too wide

Issue #499, opened 2017-09-23
Add new E1M5 music?


Voros2:
http://s000.tinyupload.com/index.php?file_id=84107328017836047279

This is a cleaned up version of a track jupiter_ex had done before. It was decided that it would go into slot e1m5, since e1m5 and e2m6 share the same track.

How about adding it now?

ElonSatoshi:
Put it in a pull request Contributing to Freedoom is a good way to learn how to use Git.

I can't listen to the music, it is showing as a blank web page.

Voros2:
I would, but I can't right now. That's why I'm suggesting it here, so that someone else can see it or the maintainers can directly add it the repo themselves.

Voros2:
Here's a more stable link: http://www.geocities.ws/v/o/voros/musics/freedoom_v2_t0.mid

Litrivin:
Files differ, the second one seems truncated:

http://s000.tinyupload.com/index.php?file_id=84107328017836047279
```
freedoom_v2_t0.mid
Size: 15.99KB (16370B)
MD5: 545d2331213870664183f6884bd7c5b
```

http://www.geocities.ws/v/o/voros/musics/freedoom_v2_t0.mid
```
freedoom_v2_t0.mid
Size: 8.57KB (8771B)
MD5: 9d590e7e9b4f8ce5fda2f55682add3bc
```
Also, I think this MIDI is way too loud and its volume should be normalized.

Proluiz:
Voros link is missing it's possible to you guys repost the original version, otherwise it's lost...

Xindage:
closed, by #730

PR #730, opened 2021-03-30
New MIDI for E1M5

northivanastan:
I always found it kind of annoying that this same MIDI was used _four times_ in Phase 1; D_E1M5 and D_E2M6 are identical and both are reused in E4. So here's a new D_E1M5. The E1M5 map feels dark and empty to me, despite the relatively bright lighting in places, so I decided that my MIDI should emphasize that emptiness. 

Please do provide any feedback!


northivanastan:
Looks like this would solve issue #499, someone else apparently had a MIDI for this but it was lost.

northivanastan:
Also released the MIDI on [OpenGameArt](https://opengameart.org/content/derelict), for easier access.

rueter37:
Really like this one, I fully support this being the track for E1M5.

northivanastan:
Someone on Doomworld pointed out a technical issue with this MIDI - the drumset track is set to MIDI OutputPort 2, which will try to use a second MIDI driver preventing some users from hearing it. Help from MIDI wizards on this would be appreciated!

northivanastan:
The same person on Doomworld (CammyBanana) fixed the MIDI, so thanks to them!

Xindage:
Approved these changes.

PR #734, opened 2021-04-12
Add new SLAD patches by Korp and MissLavender

Talon1024:
Korp submitted the original on the Freedoom Discord server, and MissLavender made the SLADRIP edits


mc776:
Refreshingly less Doom-like!

I do like how dangerously glowing bright the slime on the old ones looks though. Could that be done here too?

fragglet:
Thanks!

PR #735, opened 2021-04-12
New Combat Slug muzzle flash by Craneo

Talon1024:


Xindage:
This definitive an improvement.

mc776:
Looks a lot better, but while you're at it can you make it so that the flash is the same size for all angles? Right now in the H frame it gets way smaller as you get closer to a side view.

(And if you do this please also remember to update the BUILDCFG file so the sprites are given the correct alignments)

Xindage:
talon can you check if this is really ready to commit?

Xindage:
ok i tested the experimental wad to see the difference between the 2 fatsos and well so far it's an improvement.
Bus still this sprites need some rework.

PR #741, opened 2021-06-09
New boss brain sounds for Phase 2

guynamederick:
Updated dsbosdth.wav, dsbospn.wav, and dsbossit.wav files with new sounds created by me. I thought of trying to have a deep, menacing voice similar to the sounds by Fredrik Johansson in older versions of Freedoom. I don't find the current one to be all that threatening or appropriate for a giant monster leader.

The sounds I made are amplified by 4 DBs so they do not sound so quiet in-game, I also hope they are not too hard on the ears either on higher volume levels. Would like to know if they could be raised more or lowered, maybe even sped up if needed.

All sounds are exported as unsigned 8-bit mono WAV files at 22050 Hz, and the line for dsbossit.wav is a direct quote by Richard Stallman that is often linked to a misquote. The quote in question is: "All the so-called 'Linux' distributions are really distributions of GNU/Linux."

"Linux and GNU" page here: https://www.gnu.org/gnu/linux-and-gnu.html
"Incorrect Quotation" page here: https://www.gnu.org/gnu/incorrect-quotation.en.html

Suggestions and feedback is welcomed on this.


guynamederick:
Sounds have been updated after some feedback, I have a much more threatening voice in all audio files and to add to the machinery aspect of the boss, the pain and death sounds have some "glitchy" audio to make it sound like the machine that is keeping the boss alive is dying out. If this is not so wanted, I could go with the non-"glitchy" audio if that is prefered.

PR #745, opened 2021-07-17
sounds: add some missing sounds.

mc776:
**dsflame**
Recorded by me using a plastic bag.

**dsbspwlk**
Based on a recording from me with a file cabinet, with some edits by AxelMoon.

**dspedth**
Mixed from megalyth's growl1 and nthompson's dsbarexp from the attic.

**dsflamst**
Mixed from megalyth's growl3 and the "fwoosh" sound from Hideous Destructor which had started off as just me blowing into the microphone.


Calinou:
Remember to apply this diff before running `make` to build IWADs:

```diff
diff --git a/buildcfg.txt b/buildcfg.txt
index abe4ee45..61fc4f1b 100644
--- a/buildcfg.txt
+++ b/buildcfg.txt
@@ -304,9 +304,9 @@ DSKNTDTH
 DSPEDTH  = dummy
 DSSKEDTH
 DSBSPACT
-DSBSPWLK = dummy
+DSBSPWLK
 DSVILACT
-DSFLAME  = dummy
+DSFLAME
 DSFLAMST = dummy
 DSBOSPIT
 DSBOSCUB
```

It makes sense to integrate this change directly in this PR.

mc776:
There was never any delay in the sound. What are you even talking about with that?

mc776:
There *is* an audible delay between when the necromancer's fire is first visible and when the sound begins, but that's actually the fire actor itself delaying the start of the sound. In Doom this is not noticeable because its DSVILATK sound is designed to sound like the flame appearing so it usually masks the silence of the flame, but in Freedoom DSVILATK is a vocalization by the necromancer itself and totally unmistakable with any sound the flame could plausibly generate.

In fact, at first I though we didn't even *have* a DSVILATK and I'd just misremembered the archvile's see/alert sound or something - hence the attempt to make a new one which I quickly undid when I went to BUILDCFG and realized one already existed.

Calinou (approved):

mc776:
Might as well go whole hog here.

This PR now fills in all missing sounds previously linked to "dummy" in BUILDCFG.

mc776:
[TIL why I can't test DSFLAMST in normal play.](https://doomwiki.org/wiki/Sound_cutoffs#DSFLAMST) Anyway it sounds acceptable enough as a substitute, I personally like the new DSPEDTH, and the other sounds have been discussed, so I think this one's ready to go.

PR #755, opened 2021-08-17
colorblind-friendly keys, door indicators and HUD icons

slashdevslashurandom:
This pull request will change the keycards to be more accessible for colorblind users.

* The red key now has a horizontal line on it, the yellow key a vertical line, and the blue key has a cross shape. The HUD icons for these keys have been redrawn.
* The skull key HUD icons now look more like the actual skull keys, with different horns and shapes on each.
* The key-colored door indicators now also have a horizontal/vertical/X-shaped pattern drawn on them.


fragglet:
This is really cool work and I strongly support making Freedoom more accessible!

Do we have all the key indicator wall textures covered? We should make sure we aren't missing any.

For those who aren't colorblind (I'm not) there are colorblindness [simulator websites](https://www.color-blindness.com/coblis-color-blindness-simulator/). Here are some screenshots of the new keys and door indicators with Deuteranopia (most common form of colorblindness) and Achromatopsia (total colorblindness):

Screen Shot 2021-08-17 at 11 36 08 AMScreen Shot 2021-08-17 at 11 39 53 AM

Screen Shot 2021-08-17 at 11 36 29 AMScreen Shot 2021-08-17 at 11 40 01 AM

Should we try to make the indicator lines brighter for the red and blue cases?

slashdevslashurandom:
1. In later commits, I made it so that the inside-the-circle lines on the keys are already gray (and bright enough), so they're easier to differentiate. @fragglet 's comparison image uses the older versions.

2. The skull key indicator textures are not changed, but that's because these seem to be derived from an actual photo of a skull and colorized, and I don't think I can make the necessary changes. (If I had the choice, I'd make it so that the skulls instead look like the skull keys themselves, which all have different shapes and horns.)

Xindage:
The vertical bar need some touch up, for me it's a little off

slashdevslashurandom:
The vertical bar on the doors uses a color ramp where there's just not a lot of colors. I wish I could do a better bar, but I didn't want it to look too primitive...

slashdevslashurandom:
Update: made the door key indicators more distinct. Now each of them have a noticable lighting pattern matching the lines on the keys (horizontal bar for red, vertical line for yellow, cross shape for green).

Xindage:
it's ready to be commited?

slashdevslashurandom:
I personally would say yes, but it all depends on the project's maintainers, not me.

chungy:
Looks like a pretty great addition :)

PR #762, opened 2021-09-12
New Music for E1M6

Ajanddino:
New midi for e1m6 replaces the old very repetitive midi, I also updated the credits. also sorry about all the other comments didn't know how to get rid of them but the only changed file should be the midi for e1m6. Hope you enjoy the new track


Calinou:
https://github.com/freedoom/freedoom/pull/549 was included in this pull request. You'll want to use a different branch than your fork's `master` branch as a base to avoid this.

In general, I recommend not using your fork's `master` branch for pull requests. Instead, I recommend *always* creating a separate branch to use as a base for every pull request you open. This ensures you can make independent changes in multiple open pull requests against the same repository.

Ajanddino:
Okay thank you I'll keep this in mind for the future. Is this one fine though? or what do I need to do to get this done?

Ajanddino:
> The `c2m8.mid` change that was merged in #549 was mistakenly included in this pull request. You'll want to use a different branch than your fork's `master` branch as a base to avoid this.
> 
> In general, I recommend not using your fork's `master` branch for pull requests. Instead, I recommend _always_ creating a separate branch to use as a base for every pull request you open. This ensures you can make independent changes in multiple open pull requests against the same repository, and also avoids a host of other issues that happens if your fork's `master` branch gets out of sync with upstream's `master` branch.
> 
> This new track is pretty good, it's definitely an upgrade from the current E1M6 soundtrack.
> 
> **Edit:** If anyone wants to listen to the track without building a WAD, I exported it to a FLAC here: https://0x0.st/-xeR.flac
> It might not sound the same as your source port or OS configuration (I recorded it in GZDoom with FluidSynth on Linux using the SC-55 soundfont).

do I need to do this before this can be merged?

Calinou:
> do I need to do this before this can be merged?

I can't merge pull requests here, but I would recommend creating a new branch on your fork and making a *single* commit with the `d_e1m6.mid` track replaced.

Calinou:
Superseded by https://github.com/freedoom/freedoom/pull/763.

PR #764, opened 2021-09-15
New midi for e2m9

Ajanddino:
This is my last song I was working on for freedoom. completely new track, hope you enjoy


Calinou (approved):
I think this one is also a sharp upgrade from the one we have currently – it's much less repetitive.

PR #767, opened 2021-10-06
sprites: new Revenant (Octaminator) by Raymoohawk and @HorroMovieGuy

Talon1024:
The sprites in #728 were too big, so this one uses @HorroMovieGuy's smaller Octaminator sprites.


HorroMovieGuy:
> > Looks good! I notice that you also changed sounds/dsskepch.wav - was that intentional?
> 
> Yes, that was intentional. It came in the WAD @HorroMovieGuy submitted to the Discord channel.

I've included that because the new monster uses a punch as the melee attack. The old scratching sound would feel weird in that context.

mc776:
Should the manual, dehacked and manual illustration changes also be added to this PR?

HorroMovieGuy:
> Should the manual, dehacked and manual illustration changes also be added to this PR?

My previous PR with the matribite had the due changes made to the dehacked, manual picture and also the descriptions. I think putting everything together into a single PR is a good idea.

mc776:
Nice!

Don't forget the obituaries though...

selliott512:
This is a nice change. The new monster is lighter and easier to see. The tentacles seem consistent with the flesh worm and the matribite.

The conflict is in `buildcfg.txt`. Let me know if you need help resolving it.

selliott512:
I see you fixed the conflict, which is great.

Since there are two authors how do you see this being merged? The first commit with the monster graphics and sounds is by `HorrorMovieRei <>` (no email). The remaining commits for the manual, obituary, etc. are by `Kevin Caccamo <kevin@ciinet.org>`. Given that, the two authors, do you want to squash down to two commits by author, and that will be the commits actually added?

I think I can do that to your fork, if you want me to.

selliott512:
Since I'm not sure everyone here saw it I thought I'd summarize the feedback I solicited in the sprites Discord channel. A lot of people (a slight majority?) like this monsters, but there was concern about how strong the monster appears. One comment suggested "Maybe just removing the plates already work". Another was "I actually like the design of the Octerminator, but I think it'd be more suited as a hell knight replacement than, say, the revenant or any imp.". In other words, a slot where a stronger monster is expected.

So, if there is a way you could make the monster look somehow thinner or weaker that may address those concerns.

mc776:
I just tried building this (or rather master with the sprites and offsets added manually).

At first I was skeptical of the "looks too strong" thing, but after a melee with a knight where I was *viscerally surprised* at how quickly it went down I kinda see the point.

I love how *visible* it is, though, as someone who has as hard a time seeing the dark soldier in choco as I do the spectre.

One other problem: the plating makes it look like a painbro variant, if not from a distance actually a knight. Or sometimes a minigunner.

....so basically all my problems lie in the shiny silver metal bits.

I tried translating the sprites and `"80:105=152:159", "103:111=9:12", "4:4=96:96", "16:47=48:79"` actually looks really good - instantly recognizable, nasty and aggressive but not Colossus-esque:

Issue #753, opened 2021-08-07
Lower spawn cube sprite Z offset


mc776:
[I'd submit a PR but I've still got a thing outstanding so I'd figured I'd post this before I forgot yet again.]

[These](https://github.com/freedoom/freedoom/blob/master/buildcfg.txt#L2230-L2233) should be something more like **50** instead of the current ~~29~~.

Here's the putto from Hideous Destructor in Doom 2 and Freedoom, using a modified version of Nash's radius debug mod to show the hitbox relative to the sprite. (I had to crouch for the Freedoom shot to better show where the bottom of the box was relative to the sprite. Note that this actor is 32 units tall.)

![_](https://user-images.githubusercontent.com/24984517/128589874-f777ca3c-3bba-4b12-be0e-c5df76c5fd9a.jpg)

selliott512:
It looks like the one on the left aligns the lower surface of the cube with the lower surface of the hitbox whereas the one on the right aligns the center of the cube with the lower surface of the hitbox. I see from git that it's been that way since 2015:
```
commit c48c5b9f040f74891990fc08f511e320e630adbb
Author: Eric Ou <eriance@gmail.com>
Date:   Mon Aug 3 09:32:55 2015 -0700

    sprites: new cube spawner
```

I don't know which is correct, but "29" seems like a specific and intentional number. Hopefully others will comment.

You mentioned that "I'd submit a PR but I've still got a thing outstanding". I noticed that your [open missing sounds pull request](https://github.com/freedoom/freedoom/pull/745) merges from the master branch in your Freedoom fork. If you instead create a branch in your Freedoom fork and create the merge request to Freedoom master (the branches don't have to match) then you don't have to worry about your master branch being tied up with a pending pull request.

I like the `#L2230-L2233` trick to highlight the relevant lines in a URL. I'll have to remember that.

mc776:
All actor sprites are supposed to be offset so that the point represents the bottom, but for projectiles and projectile-like things this is really weird and counterintuitive to remember (and it's extremely difficult to see the difference in normal play) so I can see how it could have initially be placed in the middle like this.

PR #770, opened 2021-10-14
sprites: correct projectile offsets.

mc776:
Turns out #753 applies to all projectiles except the rocket pre-explosion frames.

(Yes, I am aware that this is how it's done in Doom. It's wrong there too and we're under no obligation to perpetuate that error.)

Also fixes a PLYC gibbing sprite copypaste error.


Xindage:
could you post some screenshots for comparison?

mc776:
Here's what they currently look like using [Nash's Radius Debug](https://forum.zdoom.org/viewtopic.php?f=105&t=60804
radebug_projectiles_before
)

Magenta shows where the collision box is. All monster fireballs and the plasma ball clearly extend far below the box, while the rocket extends a little above it.

BFG ball might be properly centered, it's hard to tell.

mc776:
resizeafter
And here's after. Mancubus offset might need to be raised back up a bit...

mc776:
Screenshot_Doom_20211102_183756

Here's what the mancubus ball looks like with the shift.

mc776:
On further reflection maybe it's better to just center everything for now based on the canonical sizes - anything that gets that precise can probably just work around this anyway.

PR #778, opened 2021-11-21
E4M8: Add Soulsphere in starting room

jadenquinn:
Double Impact, distributed through the Doom Unity Port as an addon has several discrepancies to the Freedoom version, and I am adding one back in, which is the supercharge/soulsphere that is in the starting room of E1M8/C4M8, since C4M8 is so dang difficult, thought this would take the edge off the start, although it is still really hard.

Screenshot_Doom_20211121_051103
sphere


Xindage:
did you edited this map whit slade? i'm noticing that the map turned out lighter than it was supposed to be.

jadenquinn:
> did you edited this map whit slade? i'm noticing that the map turned out lighter than it was supposed to be.

I did. Should I be using UDB?

Xindage:
i dunno why when slade edits a map it turn out lighter, like if slade wont generated some stuff, no idea really, but committing now.

jadenquinn:
Alright, thanks.

selliott512:
FWIW I tried to recreate the brightness issue with Slade v3.1.13 and I couldn't. I even checked a few of the sectors and they had the same light level.

In any case I think this is a good change.

Xindage:
I meant lighter in file weight not brightness.

selliott512:
Ok, that makes sense, and I see that now. By default Slade doesn't rebuild nodes, but it can be configured to run an external node builder. If the nodes are not rebuilt then you loose the generated nodes, but you can rebuild them with `make rebuild-nodes`, which is just a wrapper for ZenNode.

ElonSatoshi:
If Slade makes smaller files, why isn't that a good thing?

selliott512:
The WAD files are not smaller because Slade is more efficient, but because, by default, Slade does not add generated lumps to the WAD files. Some other level editors add those generated lump files resulting in a larger file size, but in any case regardless of which level editor is used eventually when someone runs `make rebuild-nodes` those lumps will be replaced/added, so it doesn't matter very much. 

Also, these generated lump files aren't used by some modern Doom engines, but since Freedoom aims to be vanilla compatible it will have them.

PR #779, opened 2022-01-16
MAP01 revision

Blastfrog:
I've meant for this revision to get in for quite a while now, hadn't gotten around to creating a pull request for it myself until now. Better late than never, I suppose.

Some tweaks to item/monster placement and some cosmetic polish.


Calinou (approved):
I gave it a quick run and it seems good to me.

mc776:
I might suggest putting the second serpentipede back on the second platform in medium? It's a good prompt to get the player to think about that area as an additional hint for the soulsphere secret.

Blastfrog:
> I might suggest putting the second serpentipede back on the second platform in medium? It's a good prompt to get the player to think about that area as an additional hint for the soulsphere secret.

I'm not really sure how the serpentipedes on the pedestals would have any effect on players figuring out how to get to the soulsphere secret. The intended hint is that the left/right windows open up more (letting you know that you can now run back outside), and then the sound of a door opening to your left when you go into the middle window.

mc776:
I think I was thinking it makes the player engage more with that area after the windowsills have lowered, where otherwise they might just think of it as a shortcut.

But then again I suppose the physical appearance of the secret combined with the sound of the lowering platform is enough.

That said, that imp's presence also adds to the player's decisionmaking whether to engage it immediately while it's still mostly harmless, or wait until later at the risk of being attacked while having forgotten about it...

PR #781, opened 2022-02-02
Some suggested monster names

mc776:
Figured I'd take a stab at the clunkiest ones here. Leaving "dark soldier" alone since we've got an Octaminator PR already while I don't think there's anything close to ready for replacing any of these.

Combat Slug: **Chimera**
A chimera has multiple heads and breathes fire, and is a combination of multiple different organisms that don't really fit well together.
The combat slug design is this vaguely disturbing blobby mass that's implied to have full-sized limbs and bones *inside* it, instead of being either a proper blob or something with articulated limbs. It has a tiny bitey true head, but also two mechanical "heads" that both fulfill the "oddly fitting parts" element of the chimera but also spit fire like one.

Technospider: **Manticore**
A manticore has a human head on an animal body in a way that's been depicted in a much more intentionally disturbing way than your typical sphinx or erinyes. It is famous for its deadly stings that it launches at its enemies.
The technospider's head is just this really weird... mess... sitting on top of its body. It shoots a perfectly precise stream of fast, high-damage projectiles.

Pain Bringer: **Paladin**
Pain Lord: **Warlord**
Elite tank units with some vague medieval nobility connotations. I wanted to go with a more holy-knight angle but unfortunately "Templar" and "Crusader" are already actor names in GZDoom for the corresponding Strife mobs and that would just be too confusing.

Assault Tripod: **Throne**
I tried to make some kind of War of the Worlds Martian connection at first (and the reference to the etymology of Ares is tempting), but that led to a sidetrack about milkstools, and hence seats, and the fact that it really does look like a squat wizened thing sitting in a high fancy chair. Add a demonic/destroying angel theme, and we get this proposed name, paired with...

Large Technospider: **Dominion**
The master and lord of this hive of scum and villainy. A second destroying angel.


mc776:
There seems to be no demand for this...

PR #789, opened 2022-05-01
New music!

KorpKat:
made some new tunes


Calinou:
I gave a listen to all tracks in this PR.

- Pretty good work overall :slightly_smiling_face: 
- Some instruments in DM06, DM07 and DM08 are a bit quiet (especially compared to the next track, DM09, which is probably a bit too loud actually).
- Some instruments in Phase 2 MAP26 are a bit too quiet as well.

KorpKat:
> * Pretty good work overall 🙂
Thanks!

> * Some instruments in DM06, DM07 and DM08 are a bit quiet (especially compared to the next track, DM09, which is probably a bit too loud actually).
> * Some instruments in Phase 2 MAP26 are a bit too quiet as well.

Did you use any soundfont? This was all composed using GS Wavetable (the standard midi soundfont)

Calinou:
> Did you use any soundfont? This was all composed using GS Wavetable (the standard midi soundfont)

I used whatever soundfont comes with GZDoom, which seems fairly accurate to SC-55 in my experience.

axredneck:
GZDoom soundfont plays some MIDIs (overall, not only from this PR) with major flaws.

Xindage:
so everyone who tested it, it's a sure go or not? it's like several midis at once, what you guys think?

PR #792, opened 2022-05-19
patches: new PIVY3

Talon1024:
I made this one from scratch, using The GIMP and [Material Maker](https://www.materialmaker.org/)


selliott512:
This looks nice to me. I think it be helpful if you briefly stated what problem you're trying to solve. You wanted to the ivy to be a bit larger and more distinct?

Talon1024:
Yes, but also the older PIVY3 has some consistency issues due to shoddy editing, and also that it was sourced from [this Minetest texture pack](https://forum.minetest.net/viewtopic.php?f=4&t=1583).

mc776:
The new one looks a lot more like vines than the old one which just kinda looks more like indistinct green goop.

Indistinct green goop might be more useful for some mappers but I think so far the general intention has been vines (I could be wrong on this).

Talon1024:
Squashed and force-pushed

Xindage:
its already a pelleted in doom format?

PR #793, opened 2022-06-05
A map for freedoom

Kyleminecraft:
(Ignore name of wad.  The name is just a left over from previous ideas before I decided to do a map for freedoom's IWAD)


Xindage:
played your wad, the map has a good layout but still lack theme consistency.

What i mean it, everything is textured a little random, when you want to theme your map you should make it consistent, and some of the textures choices you did is a little off for where you used them.
Next topic i want to comment on is few hallways had no care at all, they're just plain, it would be interesting if you do something in them.
About the layout it's average, it's good and really can be used, just need few touch ups, can be done as you fix the texturing of the map.
By last the slot that you can use can use range from phase 1 ranging from episode 1 to 2 and maybe somewhere late in cluster 1 or early cluster 2

mc776:
I actually thought the texture design was pretty solid as far as techbase stuff goes. (If anything I was going to compliment your use of the tekwall white concrete textures which don't get nearly the attention they deserve, and the way you used the different textures to demarcate the main areas of the map.) There's some tutti-frutti and alignment issues but those are easy enough to fix.

There are some issues with the flow of the map, though - lots of wandering around wondering what that switch did or where I'm supposed to go (maybe marking the switch-opening things a bit more clearly as such could help), but also getting caught in decorations (the starting area tech lamps 

I kept wishing we got the chaingun much, much earlier but it does add to the tension and ammo scarcity - gotta balance between hitting everything immediately on sight versus grazing a few too many serpentipedes and being forced to take bigger risks with the grossly inferior weapon.

I'm amused at how completely useless the rocket launchers (multiple) are, lmao. A clever trap for the unwary. This actually kinda feels like a DM map in some places, which I think is a good thing.

It's too easy to accidentally exit the map when you actually want to fight the pain bringer. The exit itself is obvious enough but a bit more room to move (or maybe raising the helipad so you have to go up some steps or use a lift) could definitely help.

Or the encounter could be redone so it's a trilobite hovering over the helipad itself.

This really feels like a Phase 1 E1-E2  starting map... and meanwhile [some folks have wanted to replace E2M1](https://github.com/freedoom/freedoom/issues/864#issuecomment-1431824128). Could this maybe see some work in the near future?

mc776:
Closing this for lack of response to feedback.

PR #795, opened 2022-06-10
Change manual to build as A5 paper size.

fragglet:
This seems like a better choice for the manual since it is a "booklet"
size similar to the original Doom manual, rather than a "full page"
size that is rather unusual for a book.

Several changes are needed to allow asciidoctor to successfully produce
a PDF. The main change is to set a bunch of images to 100% width,
especially in tables where this means "100% of cell width". This allows
the tables to scale down successfully to the smaller size. I have also
adjusted the column widths of various tables to accommodate the smaller
size without the text wrapping awkwardly.

### DIFF
--- a/manual/Makefile
+++ b/manual/Makefile
@@ -1,7 +1,8 @@
+ASCIIDOCTOR_FLAGS = -a toc -a pdf-page-size=A5
 all: freedoom-manual.pdf
 
 freedoom-manual.pdf: manual.adoc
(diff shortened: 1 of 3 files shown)


fragglet:
Example built PDF is here: https://soulsphere.org/random/freedoom-manual.pdf

selliott512:
Looks good. I didn't try printing it, but I looked at it in evince before and after your change. The text is relatively larger now, easier to read.

PR #804, opened 2022-07-14
Revision one of freedoom textures.

Xindage:
Added new textures with stock patches, check the comments.
Removed broken glasses textures since it was unused.
added a new light patche, t14_4, by korp.


Xindage:
i'll be accepting suggestions from you guys to proceed with the rework, so fell free to comment.

axredneck:
Are you going to remove more textures?
I ask because i'm working on a pwad that targets Freedoom instead of Doom, and i used broken glass texture in one place.

Xindage:
I don't plan removing more textures, just the glass. Because It was completely unused and has wrong offsets in its definition.

PR #809, opened 2022-10-05
levels: DM cleanup for Eureka warnings

selliott512:
This is to fix Eureka warnings so that F9 (Check / ALL) passes cleanly. This pull request is separated into two commits where everything other than reducing the number of DM spawns to ten is in the first commit, and reducing the number of DM spawns to ten is in the second commit.


selliott512:
Thanks Xindage. I was worried this would stay open forever and or be controversial (particularly my picking DM starts to delete). I'd like to eventually work my way through the maps, probably in alphabetical order with `e1m*.wad` next, with a similar cleanup. Maybe not right away, but at least some over the next few weeks.

PR #815, opened 2022-10-30
levels: Cleanup Eureka errors in E3MN levels

selliott512:
```
In levels/e3m*.wad files fix errors flagged by Eureka when F9 (Check / ALL) is pressed. This is mostly missing textures and similar errors.
  E3M2  Remove unused texture on an animated switch
  E3M3  Split stuck easy medium worm into two worms
        Move a few stuck zombies apart
        Fix Medusa by changing WOOD1 to WOOD8
        Remove crusher linedef that did nothing
        Fix transparent wall by changing MIDGRATE to METAL
        Fix transparent wall by changing AQTRIM03 to AQRUST10
        Fix Missing textures
  E3M4  Fix ceiling below floor
        Delete doom cameras (#32000)
        Fix Medusa secret area with GSTGARG
        Fix Medusa red sound tunnel with ROCKRED2
  E3M5  Separated two pairs of stuck monsters
        Delete two DM starts to get down to ten
  E3M6  Fix mismatched sector by deleting extra linedef
  E3M7  Clear flag 25 on linedef behind player start
        Remove trilobite in the void
        Clear hovering Medusa texture in green walled lava pit
        Fix Medusa exit sign
        Clear unused tag in on linedef
        Clear unused tags in sectors
  E3M8  Set impassable on all outside walls
  E3M9  Clear unknown sector type 72
        Clear unknown linedef type 181 multiple places
        Clear unknown linedef type 160 in small triangular sector
        Fix Medusa by changing SKSPINE2 to SKSNAKE2
        Delete doom camera (#32000)
        Clear unused tags in sectors
```
----

These changes should be minimally noticeable in the game, but I wanted to draw attention to a Medusa texture change that I made.

E3M9 has a room with spine textured walls that has a hidden exit:
e3m9-medusa-before

which is now less hidden:
e3m9-medusa-after

I haven't seen other places in the game where we caused the player to doubt themselves as they  wonder how they got into the room. Using this texture seems like a possible compromise to me. The stuck player will investigate this linedef before too much frustration.

If we really want to preserve this behavior I could add a texture that contains a single patch, making it Medusa safe:
```
SPINE3_1        128     96
*   SPINE3_1    0       0
```
And use that for this wall.

Or I could make the sector behind this linedef into an open door that closes when a linedef, somewhere in this room, is crossed.

Or I could change the texture of all of the walls in the room to `SKSNAKE2` which would make the floor. walls and ceiling consistently look like something like a giant stomach.


selliott512:
There are two other case where my Medusa fix made a secret area a bit more obvious, but I think it may be an improvement.

For E3M3 there was a secret two sided secret area in a wooden wall that surrounds a lava pit. The passes over it on a raised walkway, so the user might not notice the secret area at the base. Now with `WOOD8` instead of `WOOD1` the user is more likely to notice the secret area. I think that's reasonable since the secret area could easily be overlooked.

For E3M4 there is now a gargoyle (`GSTGARG` texture) on the wall of the two sided secret area. Previously there was no texture misalignment hinting at all, so I think this gets the map closer to a reasonable amount of hinting.

Admittedly some of these Medusa warnings may not an actual problem in game, but having the Eureka check pass cleanly is valuable.

To elaborate on my texture comment I can preserve any of these secret areas exactly by adding single patch textures - they are all small enough that they fit in a single patch. If so let's add a new section to `lumps/textures/textures.cfg`:
```
; Add single patch textures here to resolve Medusa effect problems or warnings

SPINE3_2        128     96
*   SPINE3_2    0       0
```
where each single patch texture is named after its patch.

PR #816, opened 2022-10-31
levels: Cleanup Eureka errors in E4MN levels

selliott512:
In levels/e4m*.wad files fix errors flagged by Eureka when F9 (Check / ALL) is pressed. This is mostly missing textures and similar errors.
```
  E4M1  Reduce deathmatch starts from 11 to 10
        Remove unused Strife tag #999
        Add missing textures
  E4M3  Reduce deathmatch starts from 12 to 10
        Add missing textures
  E4M4  Reduce deathmatch starts from 11 to 10
        Add missing textures
  E4M5  Reduce deathmatch starts from 11 to 10
        Remove unused tag #8 from sector
        Add missing textures
  E4M6  Reduce deathmatch starts from 15 to 10
        Add missing textures
  E4M7  Reduce deathmatch starts from 17 to 10
        Remove unused Strife tag #999
        Remove unused tag #2 from sectors
        Add missing textures
  E4M8  Reduce deathmatch starts from 12 to 10
        Remove transparent unused MIDGRATE texture on solid
        Add missing textures
  E4M9  Reduce deathmatch starts from 16 to 10
        Remove unused tag #13 from sector
        Change transparent MIDBRN1 textures on solid to METAL
        Remove unused transparent BRNBIGC textures on solid
        Add missing textures
```

This is should be straight forward and not controversial other than possibly the loss of deathmatch spawns, but will they be used anyway? Ten spawns is still perfectly playable.

Just in case anyone is concerned about the "Add missing texture" changes - the texture field for each sidedef is fixed width, so adding missing textures does not increase the size of the WAD file.


selliott512:
There was recently a discussion on the development Discord channel about whether the 999 tags could be helpful for ZokumBSP, which Freedoom does not currently use. I don't think they are for a few reasons.

For both E4M1 and E4M7 the 999 tags are applied to *sectors*, not linedefs, which ZokumBSP cares about. Also, the user does interact with the lower portion of the linedefs, since they are stairs, so I don't think we can safely exclude them from the BLOCKMAP lump.

For MAP13 (a different Eureka cleanup PR) the 999 is applied to a tiny linedef, but I don't think it makes sense in that case either - the user does interact with that linedef.

PR #823, opened 2022-11-16
Made the DM only section in E2M1 playable

Snoucher:
I have modified the map so that you can play through the DM only section of e2m1. This should help with players feeling like they've not fully explored the map.

P.S. I have never used github like this before, please don't bully me.


selliott512:
Congratulations on submitting your first Freedoom pull request. I don't feel strongly about whether this should be done either way, but I'll give you some tips that may be helpful for this or future pull requests, but don't worry about it too much - people can fixup what you do before applying it to the repository.

Generally git commit messages should only describe the code change since it lives on  forever in the git repository. The "P.S. I have never used github like this before, please don't bully me." would probably be better off as a comment. Also, most of the folks here are pretty nice.

I like the idea of having an extra secret button, but you've made the existing button more visible but putting it on top. The existing button makes the box of energy cells accessible. The sector that contains the box of energy cells should probably be a secret area, so maybe we should treat it like a secret area even if we don't change it.

It looks like you've made some additional changes:
```
wad2image.py -v -c yadex --iwad freedoom2.wad -d gif -s --show-cmd animate /tmp/e2m1-old.wad /tmp/e2m1-new.wad
```
e2m1

You should probably describe the changes briefly in the commit message, or in more detail in a comment, one or the other. Not in graphic detail, just the idea.

Your level WAD file contains some additional lumps namely `__EUREKA` which causes Eureka to assume that `doom.wad` is the correct WAD file when opened directly. The contents of that lump:
```
# Eureka project info
game doom
port vanilla
```
If you're using `doom.wad` to as the base resource to edit Freedoom levels it will work, but you may miss out on seeing some Freedoom specific textures.

This isn't a big deal because eventually (before release) we'll remove all extra lumps leaving only fundamental ones. You can read about it in `BUILD-SYSTEM.adoc`.

Snoucher:
Hey, thanks for the feedback! I've already marked the room before it a secret and it should be trip as intended. Plus, the energy cells are still DM only, as you shouldn't obtain a plasma rifle in this map, it's only the rocket launcher that you get to play with.

And no, I do not use doom.wad to edit the map. Maybe I exported it wrong or there's some default settings I forgot to change.

I understand if the map doesn't get merged. I just wanted to show a fundamental problem with the map when played in singleplayer, and to propose a change that for (better or worse) makes the map more enjoyable when played in singleplayer.

Xindage:
i dont agree much with these changes, first these things are no DM only, multiplayers actor are like it says "Multiplayer" if you open this map in coop these objects is going to be visible and public to anyone pick it up, these closets and "unused areas" exist just to help deathmatch, otherwise we'll have the same issue that we have when player doom2 in coop. (here's this free bfg,ssg and rocket launcher in map01)

Snoucher:
Thanks for enlightening me about the co-op factor of the multiplayer items Xindage, it would've never came to my mind if you've never mentioned the possibility. I'm sorry, however now that you've mentioned it, I conclude that the alterations of the map are not currently acceptable since I exclusively focused on making the single-player factor of the map better.

You still have the problem with people thinking that they are missing a huge portion of the map, so I plead for a better alternative that isn't going to arouse players that they aren't getting the full experience. It's very important that people don't feel cheated out of a considerable (and visible) portion of the map just because they aren't into deathmatching. Maybe make it an unappealing ledge, that way it will be more of a surprise than finally getting the full experience and knowledge of a map.

PR #830, opened 2022-11-30
levels: Fix GRAY1 walls around the end trilobites

selliott512:
This is similar to the previous fix. Prior to this fix the player would have to be pressed up against the closest corner of the balcony across from the exit to notice. Now it's all but impossible to notice.

The trilobites are raised before the player enters the room with the exit portal, so it doesn't matter how visible the hole is in that room.

---

I'm going over all of the "missing" textures that I set with GRAY1. Hopefully not too many more problems.


selliott512:
As per the recent discussion in #development I'm going fix this by reverting the `GRAY1` texture change.

PR #831, opened 2022-11-30
levels: Fix gray walls around the blue key for E2M7

selliott512:
This pillar used to not have a texture presumably to act like an invisible force field around the blue key. However, they only way the player can get to the key is to pass through the door, which triggers W1 Floor up LIC (5), which drops the sector, so the player can never interact with the force field. The solution is to drop the gray walls to the surrounding sector. It should be possible to clean this up more, but this seems like the safest solution for now.


selliott512:
As per the recent discussion in #development I'm going fix this by reverting the `GRAY1` texture change.

Issue #806, opened 2022-09-11
Map27 needs deathmatch starts


mc776:
Lest anyone say "you should play freedm": sometimes it's fun to have a deathmatch in a known singleplayer/coop campaign level, to see how differently it plays and try neat setups where you're in the monsters' position. Some mods also have pvp last person standing modes where the monsters are actually meaningful.

Besides, it's a bit jarring and unpolished to have the other maps have the starts but then suddenly have the game abort because map12 has no DM starts even though 13-26 would have played just fine.

mc776:
Editing the title now that #812 takes care of Map12

selliott512:
I added the minimal 4 DM player starts as part of my Eureka cleanup in order to make it happy. It was the only map that did not have them. Feel free to add more, but maybe that's a lower priority.

selliott512:
We can close this?

mc776:
Can we hold off a bit? I was hoping to see if I could do a few things this weekend. This is a fairly large map and only 4 starts is kinda awkward.

Also one of the DM starts is facing the wall...

selliott512:
Hold off with regard to the upcoming release? I'm totally fine with that. There's some last minute cleanup I want to do as well.

mc776:
Oh, right, that.

We definitely don't have enough time to actually test this but at least the PR I've added now shows a bit more of the map in a busy match.

selliott512:
I could only get the door to work from inside the beserk room for gzDoom, but not PrBoom+ and Chocolate Doom. Maybe plain old DR Open door instead of S1 Open and stay /fast ?

You have 13 DM starts exceeding the vanilla limit of 10. It's on the [release checklist](https://github.com/freedoom/freedoom/issues/400). I've reduced the number of DM starts to 10 for all other maps recently. Can you spare 3 DM starts?

PR #833, opened 2022-12-01
levels: add more deathmatch starts to Map27.

mc776:
Also create a door out of the berserk pack room.

Addresses #806.


selliott512:
It looks like the Ubuntu build server had trouble pulling a dependency.

mc776:
Do I need to re-do the entire PR? Since I need to make changes anyway it's fine though I should finalize the changes first.

Re: https://github.com/freedoom/freedoom/issues/806#issuecomment-1335267823

I can get rid of 3 DM starts but the door thing I don't understand at all - changing the special to 1 does not fix it and I have no idea why it only works in GZDoom but not in Chocolate.

selliott512:
You should not have to re-do the entire PR. You can either add another commit with the newer version of `map27.wad`, or amend the commit with the newer version. I usually use git from command line, but I assume the web interface will allow you to upload the same file again (but I haven't tried it).

It looks like you have no tag (tag of 0) for the linedef in question. The sector tag is 4. It's like GZDoom is able to guess that you want it to apply to the sector behind it, but other ports are not.

selliott512:
This may have been overlooked in the most recent batch of merges. We should get it in before the release. I don't normally merge level changes, but I can tomorrow if no one else does since this one has been approved, as far as I know.

mc776:
Running a botmatch on Hideous Destructor instagib for a few minutes didn't bring up any obvious game-stopping stuck moments or offputtingly awkward starts.

I am unable to squash any commits on my end.

I don't have anything further to do on this so I think it's ready to go.

Issue #798, opened 2022-06-22
Rename Map27


mc776:
I was skimming the issues and I saw the typo in #787 "Warped Elementary" and the implications of shooting up a place with "Elementary" in the name suddenly jumped out at me.

This probably won't be the last time someone makes that mistake.

Meanwhile, the actual name "Warped Elementality" is completely meaningless and has nothing to do with the actual map. (I've wanted to suggest a name change for a long time even before the current map stood in this place (see #693 etc.) but this "Elementary" connection was a last straw.)

I think a new name would be good, but it's not appropriate for me to submit a PR unilaterally. (yet)

The map right now is a fairly dark-ish map with a lot of stone and wood and candle-lighting, kinda fancy.

...and has a *lot* of blood.

...and some very blatant cross motifs in significant areas, like how you'd build a cathedral.

~~Some ideas I can brainstorm:
~~Umbral Halls~~
~~Temple of Blood~~
~~Altar of Sacrifice~~
~~Halls of Sacrifice~~
~~Sacerdotum~~
~~Foundation of the World~~
~~Bloodhalls~~
~~Blood Church~~ - yeah no, churches are a pretty common target for spree shooters too~~
EDIT: Best candidate so far: **Palace of Red** as suggested by @Talon1024 

Link to the DEHACKED lump so people can see the adjacent names for context: https://github.com/freedoom/freedoom/blob/master/lumps/dehacked/dehacked.txt

Thoughts?

mc776:
Reading this list again with fresh eyes I think **Halls of Sacrifice** fits best with the surrounding map names...

MBYlt:
So because someone can misread it and get offended, it needs to be changed?

None of the proposed alternatives are any better:

**Umbral Halls** - Could be misread to suggest umbilical cords - that's a yikes.
**Temple of Blood** - Oh, so it's okay to shoot up Jewish places of worship?
**Altar of Sacrifice** - Using the most sacred place in Christian houses of worship for your shooting, eep.
**Halls of Sacrifice** - Wow, what do you have against the creator of the DOOM Bible?
**Sacerdotum** - Could be misread as "Santorum", the name of a gross politician and the gross stuff that comes out after anal sex.
**Foundation of the World** - Are we suggesting that the world is founded upon slaughter? Pretty dark and grim, why must everything be so negative?
**Bloodhalls** - A woman's reproductive organs are no place for games!

[...]

PR #838, opened 2022-12-07
dehacked: change Map27 name.

mc776:
Addresses #798.


Talon1024:
The level title graphic needs to be changed as well.

mc776:
Does it? I thought these were auto-generated when the wad is built.

Talon1024:
Ok, it doesn't. I just thought I remembered a discrepancy between the level title graphics and the level names in Freedoom.

Issue #842, opened 2022-12-11
Map12 blue door has gross texture alignment


mc776:
Discovered while checking #800 

Probably just needs to expand the thing by a few pixels

![badbluedoor2 png](https://user-images.githubusercontent.com/24984517/206928139-c972a9d7-2196-44de-929f-e27eadc86cad.jpg)
![badbluedoor1 png](https://user-images.githubusercontent.com/24984517/206928141-c27b75f3-6c49-4676-b15c-132c5999fa3a.jpg)

Issue #800, opened 2022-06-25
MAP12, switch offsets


PROPHESSOR:
Hello
Is it OK?

![image](https://user-images.githubusercontent.com/19251320/175792315-83ec4699-17d6-432a-a749-3f9860fc8c20.png)
![image](https://user-images.githubusercontent.com/19251320/175792326-e2833ce1-d7f0-43ae-afbf-c2182fc4cd84.png)

![image](https://user-images.githubusercontent.com/19251320/175793022-299c13fc-4f93-4988-a6b6-3809d4bd1203.png)
![image](https://user-images.githubusercontent.com/19251320/175793024-2ea97a2e-1a07-4375-9fbc-91ade2440e88.png)

![image](https://user-images.githubusercontent.com/19251320/175793074-9865c4a1-e80c-4760-9140-4148ad15fc21.png)
![image](https://user-images.githubusercontent.com/19251320/175793077-b4577a73-95e8-45ac-b778-f4411f6b1d69.png)

![image](https://user-images.githubusercontent.com/19251320/175793222-2fe2110c-0cb8-475a-a53f-41ab37004f18.png)
![image](https://user-images.githubusercontent.com/19251320/175793224-dedd7272-c317-47a6-8ca0-8120709b46f4.png)

And other switches like these

Catoptromancy:
Current map is a hastily made vanilla version. Map really should be preserved in its boom version elsewhere and replaced in Freedoom.

Xindage:
requires confirmation now.

mc776:
Lift switch is OK.
![1_lift_ok png](https://user-images.githubusercontent.com/24984517/206928045-2d89250e-513e-4caf-befc-2ec700d2ee89.jpg)

Other-side yellow door switch is not OK.
![2_yellowswitch_bad png](https://user-images.githubusercontent.com/24984517/206928054-7a373145-884a-443f-a9ff-98b569d2d35d.jpg)

Blue door switch is OK. (Assuming I'm looking at the right one - the actual architecture seems to have changed, not just the texture alignment.)
![3_blueswitch_ok png](https://user-images.githubusercontent.com/24984517/206928074-e5592869-b87b-471c-a1d3-2c9e00917a32.jpg)

Issue #405, opened 2017-03-01
misaligned button texture in MAP12


jonathan-gramain:
Hi,
One of the wall button textures in MAP12 (railroads) is misaligned everywhere it appears in version 0.11 (I think 0.10 was ok).
Thanks otherwise for this awesome level!

![button_texture_misaligned_1](https://cloud.githubusercontent.com/assets/3418241/23447785/1c6ea690-fe02-11e6-83c7-0707f8834114.png)
![button_texture_misaligned_2](https://cloud.githubusercontent.com/assets/3418241/23447786/1c70ec66-fe02-11e6-94bd-da04a50ca94a.png)

Voros2:
It's the only Zok (Zok is the level author) map that suffers from this. Map15 also has misaligned switches.

Catoptromancy:
Those maps are being converted to vanilla, will eventually take care of this.

mc776:
Both of these have been addressed (see my reply to #800) and I think this issue can be closed.

PR #844, opened 2022-12-15
levels: fix Map12 switch and door.

mc776:
Addresses #800 (#405) and #842.

Also fixes some texture alignment issues in that grey room in #842.


mc776:
comparison of the grey room bits, top row is before.

mc776:
Fixed some more stuff around that room.


The wall on the other side just wasn't aligned to match the details at all, and should have that door trim thingy to match the other side of the window.

The continuous floor strip wasn't matching anything so I moved it to the top.

Lowered the door so that at least one side didn't have the top bit cut off like that.

Got rid of that awkward seam in the little screen walls.

Added a little "watch your step!" just because.

mc776:
Fixing that cut-off bullet hole before it kills me.

PR #846, opened 2022-12-18
levels: rebuild nodes

selliott512:
Ran "make rebuild-nodes"


selliott512:
I refreshed this just now with latest master so the maps changed I was last involved in late 2022 are now rebuilt. I would merge this before my next PR, the one with the Eureka fix for E1M9, since it will depend on this.

selliott512:
I've fixed the merge conflict by redoing it with latest master. This should be ready again.

PR #848, opened 2022-12-19
sprites: new torches.

mc776:
Add visible stand at bottom; make tall torches the same for more similar proportions and aesthetics to id.

Trilobite head courtesy of Craneo.


Xindage:
just to make sure are they already pelleted in doom mode? or are they truecolor png yet?

mc776:
All images were converted to Doom paletted images in Slade, then re-exported and edited by hand using only the colours found in the image. They shouldn't contain anything not found in the Doom palette.

PR #851, opened 2022-12-24
Edit to E1M2 (phase 1)

shinobody:
Made changes to E1M2. Improved encounters to be more interesting, give the map a clear theme (fighting Pinkies in different settings), improved texture use for easier navigations, signposted progression better, and added some visual flair (big standing AGM letters outside the lobby, some outdoor areas, and shadows in the final room).


mc776:
I appreciate that this no longer looks like a door, which has bothered me forever about this map:
cr_2022-12-24-171630_1600x900_scrot

I'll withhold judgment on the rest until other people can give feedback, but... *why*:
cr_2022-12-24-171701_1600x900_scrot
cr_2022-12-24-171512_1600x900_scrot
There's no meaningful justification to forcibly breaking the flow for the player. If they've got the skill to get out of there to evade enemies let them. And the wall is hideous.

And why arbitrarily force the player to be unable to turn back? That people *aren't* regularly trapped in arena fights is one of the things people have long appreciated about the original Doom's level design.

mc776:
The same comment here https://github.com/freedoom/freedoom/pull/852#issuecomment-1364730915

applies to this PR, mutatis mutandis.

shinobody:
I wouldn't normally do this, but the clear emerging theme of this level is 'encounters with Pinkies'. This level is chock full of Pinkies. And becuase of Doom's poor pathfinding, if you can get away from Pinkies even a couple meters, they completely stop being threatening.
The keycard encounter is the worst offender because I noticed that if you manage to ride the elevator down, the pinkie will literally just sit on the edge of the cliff, waiting to be picked off harmlessly. I hate when people mindlessly place Pinkie encounters in such a way that allows to kill them while being in no threat at all.

I guess an easy fix would be to box in player but much subtler - rather than locking player in an arena, I could simply spawn extra enemies where the exit to the arena is. If player manages to kill or avoid them, they deserve the win. (The enemies don't have to be particularly tough or threatening, as they would only exist to complicate the combat puzzle -1 or 2 imps would be enough).

Texture changes were done primarily to make it easy to navigate - I thought original used textures kind of haphazardly, and was sometimes hard to navigate because of that. In this version, there's a clear 'theme' to textures used in each room - one room is 'green', one is 'brown', on is 'grey startan', area before red keycard uses computer textures, one has industrial grating for floors... For example, you have to admit this passage:
image
Is not much harder to accidentally not pay attention to than it was before. Or more specifically, notice that it was opened by a switch.
image

mc776:
> rather than locking player in an arena, I could simply spawn extra enemies where the exit to the arena is

I'd much rather see this, or even make the exit a bit harder to reach while under attack (a couple crates forming a path back up, etc.).

shinobody:
Okay, implemented the feedback by changing those two encounters to discourage player from escaping rather than forcibly boxing them in! Thanks!

shinobody:
Okay, so mc776 made some edits on Discord (including a very pretty outdoors area), I made some tweaks, and then they made some more edits. This is the resulting collab edit that I believe is the final form of this unless someone has more feedback.

shinobody:
New edits! People in discord commented the last outdoor area is lackluster, so mc776 redesigned it and it looks really cool now. I also did some minor tweaking to some linedefs.

Xindage:
seems like its final so? can someone approve this?

mc776:
I have no credentials or authority but I approve.

PR #853, opened 2022-12-24
levels: Address UX issues in E1M1; some texture alignments.

mc776:
The big things I see that need fixing in terms of visual design:

- The door at (800,1447) is annoying because you don't see the blue keycard trim until you're right up to it.
- The door at (940,0) is too hard to notice.
- It's not clear where the player had come from - there's no teleport pad and behind the start is just a dead end.
- You can easily see the water floor textures on the pillars as they descend and it looks awful.

The doors to the prison cells could be more obvious doors but that's best addressed by modifying that texture since it's clearly labeled as a door in the lump name.

So to address those, I have:

- Moved the pixels around that blue doorframe so the blue strips are wider and face the player more as they approach.
- Moved that entire right-path door to face the player's start position.
- Added a broken-looking fake door behind the player start, with a hole that's theoretically big enough that one could imagine a player going through but not actually big enough to do so even while crouched in GZDoom.
- Changed the platform-lowerer into a walk-over line so that the floor texture can change.

Some consequential/incidental changes:

- Changed the wall texture on the other side of that first door to something that lines up with that entrance.
- Moved the shotgun shells by that wall so they're visible to the player, allowing the shotgunguys a second's worth of ambush if the player isn't careful.
- Fixed up that door leading to the upper outside, making some textures match better and raising the entire structure so the top of the doorframe lines up with the surrounding textures.
- Made that former platform-lowering switch into a switch to open that door, just to have something to make the player walk in there.
- Moved Player 1 to face the lift dead centre since it looks better that way.


mc776:
Before and after shots plus a shot of the new switch.
e1m1fix

mc776:
Before/after of the area behind the first door.

shinobody:
Okay I think we should have a talk instead of an edit war. That's just silly.

mc776:
This isn't something "we" "should" have a "talk" about. This is just another competing proposal presented to the community for review. Either can be accepted or both might be rejected in favour of some third thing or the status quo.

mc776:
Chocolate Doom screenshots of fixes in latest commit:

Issue #791, opened 2022-05-14
Phase 2 Map06 staircase lighting looks bad in hardware-accelerated truecolor


mc776:
![Screenshot_Doom_20220514_115839](https://user-images.githubusercontent.com/24984517/168444969-4f332223-6d80-4f2c-86bd-fd9a44a18b25.jpg)

It seems like the light levels on the bottom two stairs should be the other way around...

fawtytoo:
Would it though? You do have the gamma turned up in that screenshot, but the 2nd to last step has a "shadow" which is consistent with the walls either side, and the last step has lighting consistent with outside as if that's where it's lit from. Maybe?

mc776:
I have no idea what happened to the gamma on that shot.

I did have similar thoughts when I posted this but ultimately it really makes no sense:

- the closest apparent light source should be that yellow strip.
- that sector is under a solid ceiling, but so is the one right next to it closer to the exit. If the light is coming straight down that it would create a shadow like that, surely that other sector would be similarly affected, but it isn't.
- the rationale for that other sector being brighter is that it's closer to the big open area by the exit. If that's the case, then the light must be coming in at an angle, in which case the suddenly-dark sector wouldn't be the one that's dark but rather one or two of the sectors higher up the stairs; if instead it's being lit by ambient light, then this darker sector wouldn't be that dark to begin with.

I'll post a PR at some point now that I'm a bit more confident about editing maps without breaking everything.

mc776:
forgot to attach these
![whereisthesky](https://user-images.githubusercontent.com/24984517/214116181-faeddc36-54d1-4347-b25b-c1136f08195d.jpg)
![brightersector](https://user-images.githubusercontent.com/24984517/214116186-afbe760f-e71b-4985-a8f7-93e67ebdf16c.jpg)

mc776:
On the other hand, the effect looks *incommensurably* better in vanilla
![_](https://user-images.githubusercontent.com/24984517/214116699-064f3335-0a4a-444b-9964-d84bdaa6836c.png)

mc776:
compared to GZDoom
![_](https://user-images.githubusercontent.com/24984517/214116968-09c55560-d22d-43fd-ad67-f6b0ace17c6a.jpg)

It seems the problem is that the sector closest to the outside is too bright which ruins the effect...

fawtytoo:
[...]

Issue #714, opened 2020-12-18
Map 10 has shadow on start point


crabcrabcam:
![image](https://user-images.githubusercontent.com/13933264/102617406-0b141300-4131-11eb-9602-e1243e552a52.png)

Tested in prboom-plus (dsda-doom fork), crispy-doom, and gzdoom. 

The shadow flickers in and out, even while staying still. (I'm firing as ctrl is part of my screenshot bind, and is also shoot)

Xindage:
yes i remember this spawn point it's was based in plutonia start, prob it's still or it's a leftover from the previous versions.

mc776:
I always thought this was intentional, just... weird.

Should it just have the lighting changed to match the surrounding area, or do we even want to keep the Plutonia entrances and exits at all?

Calinou:
I think it's a neat effect, but I wouldn't mind going for a different spawn design where flickering would be placed *around* the initial player position, rather than exactly on it.

mc776:
Changing the sector special to 17 (light flicker) gives us a nice subtle not-entirely-perfect-teleport-pad effect.

mc776:
Since we're making some changes here, does the way that landing pad fails to tile properly bother anyone else? For all other aesthetic considerations it seems to be the best flat to use but... oof.

mc776:
Sorry, just had to deal with this one other little aesthetic thing that's bugged me forever about this map - the way the water tunnel just kinda *ends* like that.

Here's a viny waterfall instead:
![_](https://user-images.githubusercontent.com/24984517/215050524-ce4b64f3-dfac-4d6d-806b-5e67d0914e0f.jpg)

mc776:
I think this can be closed.

PR #865, opened 2023-01-24
levels: adjust lighting on Map06 stairs to courtyard.

mc776:
Addresses #791.

Also fix one texture alignment at the start of the map.


mc776:
Almost forgot to mention: that weirdly dark bit of wall now has an overhang casting a shadow on it.

mc776:
Before (current Map06) and after for this new update to the PR:

mc776:
Just noticed this map has the same issue as #714 ...

PR #868, opened 2023-02-14
dehacked: revise intermission texts.

mc776:
The current ones have lots of sentence fragments that are somehow austere in tone while still having lots of unnecessary words - a combination that doesn't exactly flow too well. This is an attempt to rewrite a lot of them while trying to keep all the critical elements of each.

A lot of the "it doesn't matter, who cares" / "oh. a trap. AGAIN." talk also felt like it would pull the reader out of the game, however in-character it was clearly intended. (I understand it's inspired by "where's your fat reward and ticket home" from Doom E1 but if you look at all the other id intermission texts things are much more triumphant and hopeful in tone - D2's "There must be a way to close it on the other side. What do you care if you've got to go through Hell to get to it?" is more bravado than despair.)

Replaced some things that create expectations of specific things to be found in the very next map but aren't there - the Map06-7 lift (you start on a teleporter pad), the Map11-12 platform bolts (you start just inside the door), the "massive doors" of the Military Labs (you start on the end teleporter immediately after E1M8), etc..

I was tempted to add some details about the small town at the end, but I think restraint is called for here - each reader should imagine their own nostalgic little ~~Arcadia~~ pastoral idyll.

Also "dissipitates" might've been a bit slangy still...


mc776:
It does seem some of the old lines were overly cautious about the length limit so some sentences are reformatted but otherwise unchanged.

All messages were tested on Chocolate Doom which I assume would be the most consistent and conservative, but if there are any line-too-long errors please let me know which port and settings.

mc776:
So something came up in a discussion on a fanfic discord that prompted me to re-read the proposed and current texts in a significantly different "character voice" than my default (or at least the one I gravitated towards when I wrote this) and there is definitely something to be said for the conversational tone of the current.

This might need some tweaking.

fragglet:
Okay, let me know when you've revised the text to your satisfaction and we can take another look.

mc776:
I think it's as good as I'm going to be able to get it.

Issue #867, opened 2023-02-05
Map starts and exits with that zero-light effect


mc776:
Starting this issue as a continuation of #714 (which I'd kinda hijacked for general Map10 improvements, sorry)

So IIRC this is an effect inherited from Plutonia, so it's got 2 things against it:
1. clearly copied from one of the proprietary IWADs
2. looks goofy enough that it gets reported as a bug

In Phase 2 I've seen them in the following maps:
(P=uses that pad-and-diamond Plutonia start/exit marker; P*=uses this style of exit marker *but does not actually have this lighting effect*)

4 end  (but maybe keep it cuz it's actually cool-looking? Still looks like a rendering error to modern eyes though)
~~6 start P, end P~~
~~10 start~~
~~11 start, end~~
~~13 start P, end P~~
~~14 start P, end P~~
31 end P (both exits)
~~32 end P*~~
17 start P, end P
18 start P, end P
~~19 start P, end P~~
~~24 start P, end P~~

Phase 1 does not have anything resembling this kind of map start/end marker.

Xindage:
seems like it was fixed by #949

PR #871, opened 2023-02-17
levels: fix various bugs.

mc776:
Thanks to Goji!, Inuk and rednakhla on Discord for pointing these out.

E1M3: Northern lift simplified to address texture alignment problems.

E1M5: Door near (-205,1336) (leading out into open ceiling area with the big strip of lights down the middle) door tracks needed to be lower unpegged.

E1M9: Lift near (-2328,120) was split into 2 sectors, causing HOMs when they went out of sync. There's nothing that relies on this split (contrast the neat lighting stuff from Map22) so the lift is just merged into one sector.

E2M2: Shellbox near (-486,192) is right on the line between two stairs, causing it to rest on the bottom step which causes ports like GZDoom to have the sprite clip *very* visibly into the upper stair. Moved it slightly so it rests on the upper of those two stairs.

E2M3: Door leading to red key and "door" leading to soulsphere: former should be lower unpegged but latter should not, but were reversed. Two exit-door-textured doorframes also given more conventional DOORTRAK and lower unpegged treatment. The teleporter representing the hatch going down into the nukage is now fully repeatable.

Map07: Infinite height in vanilla would cause the spectres in the red key courtyard to trap the player on the entrance ledge from below in a way that could not be seen or diegetically explained. Those three spectres now warp in only after you cross the ledge. (Setting them to "ambush" would do nothing since you're in LOS with them from the top of the ledge.)

Map11: Lights above red keycard weren't aligned; moved that entire sector and added a few lines to round the corner. Removed a strobe effect on the exit teleporter to compensate for a GZDoom issue where the light would go to absolute zero during the blink.

Map12: Room to the south with the 2 stimpacks, ammo boxes, 2 chaingunners and berserk would sometimes cause some of the items to be "levitated" to the highest sectors they touch. Moved them away from said higher sectors - it looks a bit sloppier but this is a backroom not a storefront lol.

Map13: The easternmost archvile platform had the archvile stuck in the seam, preventing it from lowering in vanilla. (Worked fine in GZDoom) Moved it a little further in.

Map19: The combat slugs teleport in from a W1 teleporter which could sometimes be spent while one of the pinkies is blocking the destination, permanently preventing that slug from teleporting in. These are now WRs like the other teleporting enemies.

Map22: More W1 monster teleports that should be WR. Also filled in some missing textures in the multi-sector lift connecting the cavern to the hall in the southwest, which parts are clearly not meant to be seen moving separately but can - it still looks fucked up if you manage to desync them, but it's a diegetic fucked up now.

Map24: Another W1 spawn. This one is impossible to screw up in vanilla, but there are some mods that could end up spawning something there that could block the archviles from teleporting.

Map25: More W1 problems. The spawn source room now also has a small barrier to make sure each pinkie only goes to its own teleporter unless the initial teleport fails.

Map27: Lizardbaby dropping too far meant that the bracket was falling along with it in a visibly unnatural way.

Map29: Broke up all the long linedefs on the perimeter of the map to get around the invisible hitscan barrier bug: https://doomwiki.org/wiki/Hitscan_attacks_hit_invisible_barriers_in_large_open_areas (Ideally this entire perimeter should be redone to break up the box in favour of more natural-looking formations, but that's a bit outside the scope of a fix like this.)

Also got rid of the Plutonia-style start/end teleports on the fixed Phase 2 maps, to address #867.


mc776:
oof that automatic linking

mc776:
I'm adding stuff as I receive news of it - as long as nothing's being made worse this PR can be committed at any time, and if anything new comes up I can just post a new one.

Until then I intend to post miscellaneous fixes like these in this PR.

PR #875, opened 2023-02-22
sprites: upgrade health pickups.

mc776:
Korp's medikit and stimpack as shared on the Discord, recoloured to use the Doom palette.

The health bonus originally started out as my recolour of Korp's green bottle to give it the same yellow colour coding as the hyperdrive sphere. (If it looks too much like the medical pickups the health going over 100% looks like a bug.) The design I ended up with is inspired by Pac-Man's dots and Unreal's health plants.

The berserk is reshaded and given a similar cylindrical look to the stimpack. The face is no longer a human skull but something more alien.


Talon1024:
I think the health bonus plants should be in little flower pots. Not all health bonuses are placed on top of soil, and it doesn't make sense to have plants growing out of the floor of a UAC/AGM base that isn't overgrown.

mc776:
It's just stuck there, like a barnacle.

fragglet:
Merged. We can iterate on the design of some of the other powerups if necessary but the medikit changes alone make this a big improvement.

PR #877, opened 2023-02-23
Added old C2M3 music back into Freedoom as C1M3!

jadenquinn:
I loved this track so much, I felt that it couldn't go to waste, so I added it back into Freedoom: Phase 1 as C1M3 instead of C2M3, since C1M3's midi was always... not as high regarded as some of the others in my honest opinion.


Calinou:
- Superseded by https://github.com/freedoom/freedoom/pull/878.

mc776:
Since a recent comment elsewhere brought it up: just a heads up we're calling them episodes again now...

PR #879, opened 2023-02-25
levels: E1M9 Fix mismatched sector

selliott512:
Eureka flagged a "mismatched sector" at the bottom of sector 255,
which is the bottom door jamb in the upper right hand area of the map.

---

There is really just one commit in this merge request, but it depend on PR #846 to rebuild the nodes. You probably want to merge that first, and then this PR will look like it contains one commit.


Xindage:
conflicting changes... hmmm. can you take a look in that selliot?

selliott512:
Sure. This is my usual Eureka cleanup. It was just one file this time. I'll wait for things to calm down and try again.

PR #515, opened 2017-11-22
textures: add some custom textures

Talon1024:
flats: add FCGRATE1 and FCGRATE2
patches: add CEILVINE and CGRATE1
buildcfg: add new flats to build config
textures.cfg: add new textures to build config

FCGRATE1, FCGRATE2, and CGRATE1 are my own work.
t.v. gave permission to use CEILVINE in Freedoom:
https://www.doomworld.com/forum/topic/45052-tnt-support/?do=findComment&comment=765129

Issue #202, opened 2015-12-17
Get rid of symbolic links


fragglet:
Freedoom has a slightly weird setup where we store resources (graphics, sounds, etc.) in subdirectories by the creator of the resource, and then symbolically link them into the parent directory for deutex to read.

This made some kind of sense years ago when there was no version control and I was trying to divide up the repository in a way that I thought made sense. But I don't think it's a good idea any more. It causes several problems:

* Symbolic links don't exist on Windows, so it is difficult to submit changes to Freedoom. Example is #195 where @Blastfrog encountered problems because of this.

* The repository is harder to browse in GitHub because you can't just find a resource and click it - you just get a "symbolic link" display telling you that your resource is in another directory.

* Resources often have more than one creator. This causes problems because if someone edits a resource originally made by someone else, it's no longer clear where that resource belongs. For example I took the "policy" that it should resources go in the directory of their original creator, but I've also seen resources moved between directories when making a change. Lack of consistency means that the directories become meaningless or even misleading.

* Subdirectories duplicate the functionality of a version control repository. You can already see who made a resource by checking the revision history of a file in Git, and it's likely to be more detailed and accurate than the subdirectory name.

* Dividing up by directory can imply some kind of ownership or territory, which is counter to the nature of the project.

I propose that we get rid of the subdirectories and move them into the parent directories as a more flat list of files. There are a few exceptions where this won't work; for example, lumps/genmidi, lumps/dmxgus, graphics/text are examples of functional subdirectories where we build things; they are not simply there to show authorship. In these cases we can still get rid of symlinks by changing their Makefiles to copy generated files into the parent directories.

fragglet:
By the way, if we do this, I'd recommend splitting it into multiple commits by author directory. That way if you want to see the old subdirectory information, you can look at the revision history to see "this file used to be in the fraggle subdirectory", etc.

Blastfrog:
I'm fully in favor of this. I was intending to suggest this myself at some point.

chungy:
All sounds good, should be fairly simple to implement as well without breaking the build system. Feel free to do so :P

rjy:
See also

https://www.doomworld.com/vb/freedoom/42768-handling-maps-in-svn/

[...]

Issue #130, opened 2014-10-19
Punching walls makes sparks


AXDOOMER:
![sparks](https://cloud.githubusercontent.com/assets/6194072/4691619/80581d78-5727-11e4-9393-cb6b66b1cb6f.gif)

At the same time, I'd like to say that I find the bullet puffs too bright. Compare them with Doom's bullet puffs, which only have a few (or one) full bright yellow pixel(s) in the middle. FreeDoom's bullet puffs (no offense meant to their author) just looks a big pile of bright yellow pixels. A full SSG blast at a wall looks like there are lights shining on the wall (they look like small suns). They need to be redrawn.  (That's my opinion, what's yours?)

![chaingunning_a_wall](https://cloud.githubusercontent.com/assets/6194072/4691677/7df8323e-572c-11e4-8c43-4c62ec44bedc.gif)
![doom65](https://cloud.githubusercontent.com/assets/6194072/4691711/4f5ce1ca-572e-11e4-9c96-82719440d8ea.gif)

NickZ:
resolved with sodaholic's new bullet puffs.

PR #882, opened 2023-02-26
levels: more misc. fixes and improvements.

mc776:
This one's a bit more ambitious than the other one, sorry.

**E2M8**
Missiles were disappearing into the outer walls. It's clearly not desired behaviour since they only disappear when fired at ground level targets - when you have plenty of space to dodge - and not when fired at you while you're on the ledge where it might have been handy. It also wasn't happening consistently. Flipping the linedefs didn't solve the problem so the sky above the slime has been lowered to compensate.

Some more ornamentation added to the corners of the slime to cover up the resulting sky oddities.

**E3M6**
The monster closet in the far west-southwest was having some render issues so I moved some vertices around.

The skull pole in the eastern blood room interrupted my flow so I moved it.

Added a bit of stuff to the exits. Both E3M7 and E3M9 start off with a pretty strong techbase/hell mix so the additions should reflect that.

Changed up the start. The back wall now has some skylights, the players start just in front of the door and instead of one serpentipede standing with its back to you completely oblivious it's standing around the corner waiting to ambush (along with 1-2 buddies)

Rocket launcher #53 (near the star thing, which has also been given a bit of a makeover) is now marked multi-only.

Fixed various texture issues and merged various identical sectors.

Added more health and a way out of the pit with the zigzag bridge.

Moved the bars around to make it a little more obviously impossible to squeeze through.

Marked the teleporter to the red door as needing a red key. (Not marking the door itself just for aesthetics.)

**E3M7**
Fixed up some texture alignments in the first fighting area and the lift leading into the intestine area.

Fixed up some texture alignments in the intestine area and some minor remodelling of the wooden exit gate out of it.

Added an anti-pinkie-block trench (see Map13, Map20 and Map29) in front of that first drop in the red spiral chamber.

Moved some stuff around in the intestine/stomach area to get rid of jarring fake contrast effects.

Made the teeth at the start of the digestive tract white.

Added more health and armour in the areas leading up to the spiral.

Added some skylights on that long marble corridor.

Moved everything in the spawn chambers so that the last weapon you autoswitch to is the shotgun. There is no skill level that places anything you can immediately melee right behind that first door so it's just "haha we gave you all this prep but you forgot to switch" as the player takes a bunch of fireballs while switching to a proper weapon.

Made the cyberdemon fight more interesting and chaotic. All four corners now (can eventually) open back up into the hall, and monsters warp into the hall and just outside the front door.

**Map06**
Continuing from #865.

**Map10**
Addressed #867 and #714.

Softlock if you could run into (-2696,736). Moved the tree and rock to allow passage out, and also added a medkit there as a little reward (or compensation if you get cornered in there) given how long it took for this to be reported.

Marked the blue door leading to the yellow key as needing the blue key.

With jump enabled you can easily grab the blue key early and skip the entire left twisty area of the map. I don't want to fence this area off because it looks and feels like shit that way, but the player should still have a reason to go the long route. Right now there's a rocket launcher and chaingun that way.

So I've made the following changes:
- Added a tiny step in the eastern corner of that ledge that lets you get up to that level without jumping.
- Changed thing #130 from a chaingun to an ammo box, but added another chaingun that's multi-only. You will still get a chaingun in this room in single, but you'll have to pry it out of someone's cold dead hands.
- Flagged RL #202 as multi only and adding 2 more rockets in that spot.
- Flagged shotgun #365 as multi only, and added another shotgun to (-3648,400).

The super shotgun secret #1 is also trivially easy to jump onto. I've moved the gun into a secret area that can only be opened if you specifically go down that waterfall.

**Map12**
Restored staircrate access to berserk pack #510.

Changed sector 740 into a door and tagged it 56 to match the other stuff that opens in that area.

Shrank sector 437 and neighbours to align the ceiling lights correctly. Also added an overhang.

Made the monsters in sector 575 teleport into the teleport sectors inside the accessible part of that sector.

Made chaingunner #515 visible by using chainlink instead of a tinted window. (how would it even shoot through the glass without leaving holes???) It is also an imp in easy and medium.

Fixed some texture alignments/matching in the northeast tower and stairs in that big western arena.

**Map13**
Added the rising-trench trick to keep the spectres from blocking the very first jump into the big slime pit.

Gave the top of the blue key pillar a more appropriate texture.

**Map14**
Addressed #867.

Secret sector #107 expanded to keep the player from picking up the chaingun inside it even when the secret is not made accessible.

Got rid of the cage trap for the yellow key, in favour of a baron (knight in easy) warping onto the pad when you hit the switch, which must be killed before you can leave. Reasons: the cage trap can be too easily ruined by a player or zombie, even accidentally; these kinds of traps are not good for replays because they disrespect player agency and punish efficient players with a longer dead time; the trap is impossible to guarantee without either an extremely contrived funnelling the player into the trigger line, or having multiple trigger lines thus running the risk of forcing the player to wait it out on two separate occasions. An additional shell box is provided with the yellow key.

**Map15**
Northwest secret involved a multi-sector lift that clearly goes out of sync. Added a new simple lift behind it instead, and fixed up various textures there.

Moved 333 and 334 a bit over so they couldn't be picked up from the other side of the wall by the switch.

Made overlapping enemies 295 and 255 alternatives that only appear in different skill levels.

Window overlooking blue computer switch now has chain link to mark it as impassable.

The rising red key platform (and all its various components) could be triggered twice, resulting in unwanted effects. The second of these switches is inside the pit and clearly designed to prevent a softlock in case the player reached that area before the switch was hit. The second switch has now been removed; the only path leading to the area is now only accessible if the first switch is hit. As this disables the lift function of the final bridge, safety lifts are added on both sides.

**Map16**
Broke up the linedefs in the outdoor visual so everything should be under 2048.

Added fences, railings, tall lamps, etc. to better mark where the player shouldn't be going.

There's this gap between two buildings that you can slowly squeeze into and find nothing at the end. It's clearly not meant to go through but it's not quite small enough to communicate that reliably, which keeps getting some players misled. I've expanded this area and made it end with the shittiest tiny little fenced-off alley I can do in that few linedefs. The stimpack is now back there.

Spruced up the blue key building a bit, and made that curvy gap actually navigable.

Widened the little steppy things on the edges of the thick box things because navingating that was really annoying.

Miscellaneous other breakings-up of monotonous surfaces.

**Map20**
Broke up a lot of >2048 linedefs.

P1 now starts in the middle, right on that pillar that looks like it'd be a great slightly-unusual starting teleport pad, instead of having an easily opened door right behind them with no explanation how they got there without noticing the shotgun guy. The blocking pillars now have a skull so you know the skull switches are related. Said switches are now part of the fence - it looks vastly better this way and fully jump-proofing this section was always a bit of a nonconcern.

The starting building outer catwalk tends to snag the player in vanilla if they run along the side (which now has a fence to demarcate it). As far as I can tell this is completely unavoidable, so the entire area has been widened to avoid accidental contact while avoiding getting caught in the more obvious geometry of the pillars.

There are three more shotgun guys behind the first door in multiplayer. That's four shotguns!

Windows in that first red building you have to get into are NO LONGER MARKED IMPASSABLE. OUR LONG NATIONAL NIGHTMARE IS OVER. I've also added thin deep invisible grooves in the lava that disappear on your first jump attempt, to keep the spectres from blocking you.

The switch granting access to the blue keycard room is now behind a cage and cannot be punched, getting around some mods that stop the fist from alerting monsters. The ceiling has also been raised slightly to address texture alignment.

The lift that you take into the big red building if you fail the jump now visually matches the one inside.

The blue keycard lift monster spawns are now staggered so you can't clear through all of them at once and stand around awkwardly to wait for the lift to finish. (If you didn't pistol start this map you might just clear out one wave, then stand awkwardly for a much shorter and non-deterministic time until the next.) There's also a lot more of them, just so you don't clear them out *too* quickly. The monsters will teleport into one of the next areas if they try to follow you out of the room.

Figured out why I've always ragequitted this map: there was NOTHING in the marble temple that would tell the player that the skull pillar in front of the cyberdemon is a usable lift. No vanilla-compatible switch texture works here aesthetically. So I just finished the implied temple and added an altar, in the middle of which is a clearly visible switch.

Added more health (of which there had been nearly none, even what I've added is stingy af compared to even Map21) and armour and replaced a multi-only (GOD WHY) soulsphere in the southeast with an also-singles medikit.

Made the square room less utterly dreadful by adding some sky.

Added a lift out of the damage sector in the end area.

Added some secrets.

**Map23**
The two doors in the G-shaped spiral corridor just to the east of the helipad are now single-use stay-open. The problem is that with all the corpses and archviles the closing doors were creating a lot of crushed corpses that were being raised as ghosts in vanilla. (Made even worse by the fast doors which could reduce the opportunity for more incoming monsters to block the descending door)

**Map27**
Spaced out the candle display in front of the devil icons on the east with the shellboxes, to prevent more floating-ledge movements. Honestly I think it looks better this way especially with the bigger Freedoom candle sprite.

Merged/deleted a few redundant sectors with no change to map geometry.

**Map28**
Fixed a misaligned marble skull face near (-4159,1348).

**Map29**
Starting room large devil icons now have staticky trim instead of appearing to tile. Each face is now only two linedefs instead of three, as the perfectly orthogonal middle lines were subject to the fake contrast and didn't look right.

Widened the shaft on the southwest arrow and the flagellum on the northeast sperm to let the serpentipedes move more freely.

Moved armour away from the yellow key platform, guaranteeing its availability from the ground.

Bottom room with spectres and archvile now blocks the spectres from getting right up below the ledge until you've crossed it. The rising platform in that room is also a bit more visible - with the same textures and dim lighting it was almost totally invisible in vanilla unless you already knew where it was.


mc776:
screenshots showing the changes (the ones that can be screenshotted anyway): https://riotouseaterofflesh.tumblr.com/post/710393900331859968/riotouseaterofflesh-trying-to-find-the-least

PR #888, opened 2023-03-04
flats: add bloody skulls and bodies.

mc776:
Every time I need to do something with gross hell textures I'm surprised again that the clearly already 64x64-tiled gore in WALL76_1 and BODIES (SKULWAL3 and SP_FACE2) aren't available as flats.

If we've reached some technical limit for flats this can be closed, but it would be nice to have this stuff look consistent from above and the side on certain structures.


axredneck:
I think these vertical lines of blood are a bit unsuitable for a flood texture.

Xindage:
can you make the name of the first flat something more lined with the actual flat names?, one is named fbodies, why not flat5_6c or 7?

mc776:
I'll change it but I would've thought the objection would be going the other way, that it'd be better for new stuff to have more descriptive names!

mc776:
fwiw here's what the skull flat looks like right now, tiled and rotated 90 degrees:

fragglet (approved):

PR #890, opened 2023-03-08
Eureka fixes and rebuild-nodes

selliott512:
This PR contains two commits that can remain separate, but don't need to:

1. levels: Cleanup Eureka errors in five levels
2. levels: rebuild nodes

This PR depends on PR #886 , so please apply that before this.

As far as I know this is the last set of level changes for 0.13.0.


Xindage:
conflicts, and i did exactly what was told in the pr.

selliott512:
Hmm. It seems I based my changes on an old version of the "more fixes" PR that I had locally. I'll redo it today. Sorry about that.

selliott512:
Ok, so I wasn't as foolish as I thought. What happened is that the "more fixes" consisted of several commits that got squashed into one when they were merged. When that happened this branch still had the multiple commits. From git's perspective it didn't know how to apply those commits, and prior to the the two newest commits I intended, onto the master branch.

However, the files that acted as a basis for my two commits, and what is now in master, are exactly the same. I was able to fix the problem just by cherry picking. The files in the current head of this branch are exactly the same as they were before, which you can verify, if you want, (from my previous head to the new one):
```git diff da4c85c0d3b5799b0ee10f44e76ab08d7186f631 f09d19ff1c6f3ef07b2ad9998c2168e473ffac87```

If when merging you instead pick "Create a merge commit" the odds of this kind of conflict (when applying PRs that depends on other PRs) is lower, but I can understand why that's not done. Not wanting to clutter the git log with extra unsquashed commits is totally understandable. I sometimes check other branches in order to squash their commits, but that may be too cumbersome to do in all most cases.

Incidentally github's phrasing for the three merge options is very confusing. Only the first option creates a merge commit. The other two do not in spite of the word "merge". They should instead user some other term such as "append". The smaller text is more accurate.

All of this is just hopefully interesting. It's ok to keep doing what you are doing.

Xindage:
map26 :(

selliott512:
Thanks for trying. I'll take another shot at it tonight or this weekend this time.

mc776:
e1m2 :sweat_smile:

selliott512:
Ok, one more time, if you don't mind.

mc776:
Still got the bugged E1M2... (see #898)

selliott512:
> Still got the bugged E1M2... (see #898)

Do you mind if I just cherry-pick your commit in #898 into this branch before my Eureka fix commit, and then update my Eureka fix commit to fix whatever is left to fix in E1M2, and then close #898? I feel like that would be a painless way of solving the problem.

mc776:
Sounds like a plan! (and a good one)

selliott512:
> Sounds like a plan! (and a good one)

Done. This should PR be ready to go in (again).

PR #891, opened 2023-03-08
lumps: Fix phase 2 demos

selliott512:
This commit fixes the obviously broken demos which happened to be the
phase 2 demos.

---

The demos at least seem to be working now for 0.13.0, but feel free to rerecord demos that I didn't - it's possible that level changes have introduced errors that are not obvious.

This is based on PR #890 (Eureka fixes and rebuild nodes), so I would recommend applying that first.


selliott512:
This is a similar situation as PR #890. I rebased it onto current master just now.

PR #892, opened 2023-03-09
levels: improve Map27.

mc776:
It turns out DOGB01 had posted an improved version of the map on Discord back in April 2022 and it slipped past everyone. I've taken it and re-implemented the DM starts (and additional door triggers to let people into the central room from the other direction), fixed the lizard baby and candle rooms and merged some identical non-contiguous nearby sectors as before. (EDIT: and made the cavern in the southwest something a bit more bearable - and possible - to look at.)


mc776:
"because they just seemed to be regular pillars" i.e., columns extending fully from floor to ceiling with no additional horizontal ornamentation, which wasn't doing anything that some 1-sided linedefs could do

mc776:
It occurs to me that brightening the cavern reduces the incentive to get the liteamp secret, but I think that's largely only theoretical - by the time you can fight your way into the lizardbaby room *and back out to the start*, you've probably already killed all the spectres in the cavern along with anything else that doesn't either go fullbright during its attack or launch a visible fullbright projectile, so the liteamp's only real use relative to the cavern is helping you navigate the area at all in order to figure out where everything even is.

Meanwhile, if you're the sort who managed to even see the secret, chances are you never needed that aid.

As for making the entire encounter much easier, with this change it's about as easy for me to see spectres in Chocolate as it is to see them before the change in GZDoom using "Bright" sector light mode, while with the current in Chocolate I'm basically telling where the spectres are by sound and blocked movement.

selliott512:
Were you hoping to get this into 0.13.0? I know it's last minute, but if it's safe I don't see why not. It's unfortunate to miss intended fixes.

PR #894, opened 2023-03-10
levels: improve Map26.

mc776:
Continuation of #893 as the general mapfix resulted in a conflict.


mc776:
Confirmed this version can 100% everything on hard in Chocolate without noclip or IDKFA.

PR #898, opened 2023-03-11
levels: bugfix: remove E1M2 wall midtex.

mc776:
The level had a decorative door-the-player-had-come-in-from that was obscured by a wall texture.

Thanks to Smokey on Discord for pointing this out.

I've also removed the Doom Builder camera thing.

As this seems to be an isolated oversight I'll just PR now instead of saving up a batch of fixes.


mc776:
will be implemented in #890

PR #899, opened 2023-03-11
sprites: fix minigunner loop.

mc776:
The looping sprite different size vanilla bug strikes again!

This one's unnoticeable if you look at the offsets or test on even ground, but *painfully* obvious when the minigunner is firing at something to the side while standing on a ledge - it'll just shift back and forth drastically every frame.


mc776:
This is distinct from what's going on [here](https://doomwiki.org/wiki/Sprites_flickering_across_ledges_or_lifts) - the problem this PR fixes happens with just the minigunner standing on a steppable ledge (<=24 diff) on its own volition without anything pushing or moving it, seen by a stationary viewer.

PR #911, opened 2023-03-17
dehacked: improve pickup msg & nightmare warning.

mc776:
Makes the verbs consistent, gets rid of all references to spheres, gives terms to all powerups that line up with their corresponding IDBEHOLD cheats*, and hopefully reduces the bathos of the health-adding spheres a bit.

Also makes the nightmare skill level warning a little more accurate and explicit, instead of sounding like it's just the standard advanced user mode.

*It turns out the options on the idbehold cheat message also need to be abbreviated as the current one is cut off, so we can't really do the "letter = " stuff.


lumps/dehacked/dehacked.txt
```diff
@@ -144,25 +144,25 @@ STARTUP5========================================================================
...
-GOTHTHBONUS = Got a 1% health bonus.
-GOTARMBONUS = Got a 1% armor bonus.
+GOTHTHBONUS = Picked up a life charge.
+GOTARMBONUS = Picked up an armor charge.
 GOTSTIM = Picked up a small health pack.
-GOTMEDINEED = Found a large health pack, just in time!
+GOTMEDINEED = Picked up a large health pack, just in time!
 GOTMEDIKIT = Picked up a large health pack.
-GOTSUPER = You got the overdrive sphere!
+GOTSUPER = Hundredfold life!
```
Calinou, line 152: Doesn't "hundredfold" mean "multiplied by 100"?

mc776: It used to be "Hundredfold life charge" but that got too bleh so I cut out the last word.

mc776:
Since it's been bumped to oblivion on Discord here's a pair of screenshots for what this fixes:

selliott512:
Looks good. I gave it a quick try.

PR #912, opened 2023-03-18
levels: more misc. fixes.

mc776:
**Map02**
The yellow door appears to open by splitting itself into 3 parts, then moving only the middle part leaving the sides behind. This is now fixed so that the entire door moves.

Touched up the texture alignment above the exit door and replaced the floor on top of the monitor thingy blocking direct fire from the serpentipedes in front of the yellow door.

The little drops into the water area near the start can be cheesed by staying in the lit areas behind the monster blocking lines. This seems to be unintended behaviour so I've added new blocking lines that are exactly the size of the player.

The big drop past the red and blue doors needs something to stop the pinkies from blocking the player's initial drop. They're all set to +ambush now.

The exit sign has been given the flat-bleed treatment as Map01 and Map26, but this time I've actually kept the lower floor within monster step height. (I'd fix it for Map01 but it turns out the monsters step around it without issue.)

**Map03**
The green corridor had some texture alignment and mismatched flat issues; the switch door was overcomplicated for what it was doing and didn't even look better than just using a simple wall switch texture. The corridor itself had a shadow across the middle that had nothing casting it; it has been removed.

Gave the curved terminal wall light grey flats to not stick out as much.

**Map21**
The cage bars along the walls of the lava border corridor should be impassable.

**Map26**
It's possible if you're careful to grab the super shotgun without triggering the lowering panel, causing the player to be potentially lost without any further indication they should move towards the alcove. Since I want to keep this as a foreshadowing of what is possible with stealing the key from the necromancer, there's now also an armour bonus a little further in.

The rising pit was too easy to jump into before it was finished, ruining the effect. I've packed everything closer together and moved one of the hard spectres into the corner outside the pit. The northern bit of the room is also expanded to let the worms move that way.


mc776:
before and after for the main map02/map03 aesthetic tweaks:

PR #914, opened 2023-03-19
dehacked: rewrite intermission texts.

mc776:
Reposting #868 because that was way too many commits.

The current ones have lots of sentence fragments that are somehow austere in tone while still having lots of unnecessary words - a combination that doesn't exactly flow too well. This is an attempt to rewrite a lot of them while trying to keep all the critical elements of each.

A lot of the "it doesn't matter, who cares" / "oh. a trap. AGAIN." talk also felt like it would pull the reader out of the game, however in-character it was clearly intended. (I understand it's inspired by "where's your fat reward and ticket home" from Doom E1 but id was never quite this relentless - if you look at all the other id intermission texts things are much more triumphant and hopeful in tone: D2's "There must be a way to close it on the other side. What do you care if you've got to go through Hell to get to it?" is more bravado than despair.)

Replaced some things that create expectations of specific things to be found in the very next map but aren't there - the Map06-7 lift (you start on a teleporter pad), the Map11-12 platform bolts (you start just inside the door), the "massive doors" of the Military Labs (you start on the end teleporter immediately after E1M8), etc..

"Dissipitates" isn't exactly release-quality standard diction.

Some lines were rearranged so that natural sentence and phrase breaks coincide with line breaks - everything has been tested on Chocolate, it seems we do have more room to work with than the previous writers seem to have thought. (If someone out there *specifically intended* to make sure the sentences spill over into the next line in the absence of a paragraph break I'm willing to listen to reasoned argument about it; my reasoning is that the way the text gradually appears on the screen it makes conversational sense for these linguistic boundaries to coincide with the inevitable slight timing break compared to when you can skim an entire block of prose at once.)

I have, of course, left the iconic final three lines of Phase Two untouched.

### DIFF
--- a/lumps/dehacked/dehacked.txt
+++ b/lumps/dehacked/dehacked.txt
@@ -535,14 +535,14 @@ C3TEXT = You'll find no AGM here. No humans.\n\
          And you.\n\n\
          You've been tracking their movements\n\
          and comparing notes with AGM's records.\n\
-         Everything seems to lead to one central\n\
-         entity sending them out, worming its way\n\
-         to Earth through AGM's teleporter network.\n\n\
-         All you need to do now is find this entity\n\
-         and destroy it, if you ever want to see\n\
-         the constant fighting come to an end.\n\n\
-         This could be the beginning of\n\
-         your freedom - or your doom.
+         Hunting down the interdimensional being\n\
+         that has been sending its nightmare army\n\
+         to Earth through AGM's teleporter network.\n\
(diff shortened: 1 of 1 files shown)


shinobody:
Hi, I want to take a crack at tweaking these.

mc776:
Okay now that we've clarified what needs to be added - really just more emotional impact, actual themes, motivations - yeah let's talk about this...

fragglet (approved):

Issue #916, opened 2023-03-21
levels: abolish hot starts!


mc776:
Quoting Xindage on Discord:
> i dont think we should support ambush start maps
> these work horrible in some multiplayer ports
> zandronum for example load the map server side and then tell to the clients to load the map
> there's a big delay when doing that
> if you're playing survival for example, in ambush starts, there's a high chace of you die or something similar, its unfair

Personally I've always disliked them and thought they felt like the mapper was sloppy, but now that there's a legit technical issue I'm happy to literally make an issue out ofit.

Let's have a list of maps where we have them, and I can cross them out in this post once they're dealt with (with my suggestion for the simplest fix):

E1M2 (all 3 baddies in starting room, just make them turn away)
E1M5 (all zombies in starting room, maybe move the player out of sight)
E2M9 (serps in starting room, turn away)
E3M3 (starting room, maybe just have the player start in a different room and teleport in)
E3M5 (2 serps to the east, turn away or move out of LOS)
E4M1 (one zombie in the distance)
E4M3 (whole bunch of guys all over the place)
E4M4 (2 guys down to the south)
E4M5 (starting room worm)
E4M6 (one zombie to ENE)

Map10 (serpentipedes in the big field, just make them turn away)
Map11 (zombies in starting room, just make them turn away)
Map29 (the arachnotrons, the imps on the raised area, a couple pinkies - tbh I'm not sure how this could be fixed, and given the torches fully protect the player anyway I'm not sure it's worth it)
Map32 (shotgun guys in starting room, ~~can make turn away - or remove them entirely in favour of a gibbed shotgunguy right where P1 starts and a shotgun nearby; if this changes we need to change the intermission text as well~~ maybe let's keep this one because the intermission text introing it is so iconic...)

(The above parentheticals underestimate the complexity of some of the situations. #917 does more than this.)

mc776:
So far I've only checked hard - it's possible there are more monsters that appear on easy as alternatives to harder monsters and these do start sighting the player while the hard equivalents don't.

mc776:
#917 should have gotten all of the ones mentioned in the OP... if any more pop up they can be dealt with then.

PR #917, opened 2023-03-21
levels: yet even more minor fixes.

mc776:
**E1M2**
* Addressed #916. Instead of having those three goons face you differently on hard, you just get more enemies (also facing away from you) in that room.

**Map10**
* Addressed #916. I'm doing these two maps just because they were already updated recently so we might as well go all the way.

**Map11**
* Addressed #916.

* Made the colour door stripes go all the way down, because we can't fucking look up in vanilla and the things are invisible when you're right up at the door in question. Each door now also has an arrangement of lights in front of it corresponding to (when you're facing it) the shape associated with that colour keycard.

* Lowered the switches just beyond the red door to eye level.

* Made the central gazebo skylight thing properly centered in that starting room.

**Map26**
* The bridge switch was a bit hard to find. It's been moved to the panel next to the pantry.

* At some point I'd somehow forgotten to deal with ceilings, resulting in misaligned and mismatched stuff everywhere. Hopefully it's all been dealt with now.

* Added some details to the grey tunnel and made the green torch in the wood tunnel present in single/coop as a light source.


selliott512:
I think you may have this figured out from looking at issue #916 , but here are the remaining hot starts where one or more monsters are active when player 1 spawns on UV:
```
Freedoom 1:
  E2M9 E3M3 E3M5 E4M1 E4M3 E4M4 E4M5 E4M6
Freedoom 2:
  MAP29 MAP32
```

mc776:
Thanks, I'll add it to the issue.

Or is this a hint that they should all be done before this PR is merged?

selliott512:
Whatever you want to do is fine. I offered to do it in #development at one point, so I can add it to the things I check in addition to my usual Eureka checks. If I do it I'll just rotate monsters slightly since that feels like the safest change to me. In any case I want to make sure I don't cause conflicts with maps with my work, so I can wait if need be.

mc776:
I did them anyway.

Unless I somehow seriously broke some *really* obscure secret I think this is ready to go.

mc776:
...I just realized I forgot to look at the Phase 2 ones...

mc776:
And there we go.

selliott512:
Thanks. I checked that you got them all except for E2M9.

mc776:
And there's that.

mc776:
Just to keep this from turning into yet another round of feature creep I'm restricting myself to undoing flat bleed tricks that I introduced, plus those that I just happened to find on the affected maps in the course of doing mine. Should be ready to merge again absent any major objection to the changes to E3M1, E3M3 and Map02.

fragglet (approved):

PR #919, opened 2023-03-23
New basic armor and edited armor bonuses

shinobody:
Two things:
- I noticed that with all recent changes, all health items are green, while armor items are mostly red. However, basic armor is the exception because for some reason, it's green.

So I propose a recolor of basic armor to yellow. It kind of makes sense if you think about it - basic default player outfit is green, then you get the yellow armor, and later you get the big red armor. It's a progression of colors.

- In addition to that, I also edited the newly red armor shards/bonuses to look a bit nicer.


mc776:
I approve in principle, but:

1. Could you run the images through PNGCRUSH or something? GitHub is showing a 200% increase in file size.
2. How do the bonuses look in vanilla? The current ones already feel just a tiny bit dark.

mc776:
If I import into SLADE, convert to Doom graphics, convert back to truecolor PNG and export, I manage to get the tiny file sizes from before.

Not sure why... are not all of the images on here restricted to the Doom palette, or is there a bunch of metadata somewhere?

mc776:
Disregard #2, they look great!
DOOM01
DOOM00

I do suggest (which I did here) moving the offset to the absolute bottom of the sprite, though - there isn't really anything to gain for vanilla aesthetics with something this small and low on the ground and setting it at the very bottom makes sure it won't get cut off in GL ports.

shinobody:
Okay, used your re-compression and changed the offset.

axredneck:
Default player's color is green, so maybe make armor green after all?

mc776:
Was the yellow because of Quake's progression? That could make sense - player starts out already wearing green, then move up from there.

fragglet (approved):

PR #922, opened 2023-03-28
Changed copyright year plus added myself to the credits

DynTylluan:
> Changed copyright year

I think it's kind of silly that we need to remember to keep on updating the copyright year every time the real-life year changes, so I think this would be easier, although I'm not going to die on this hill if this part gets rejected

> Added myself to the credits

I have not done work _for_ freedoom/freedoom, [but I have done stuff for the website](https://github.com/freedoom/freedoom.github.io/commits?author=DynTylluan), so giving me credit would be nice

Cheers

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -964,4 +964,10 @@ E: jadenquinn8@hotmail.com
 W: http://iredmc.us.to/
 D: levels, musics
 
+N: Cass Python
+S: Owly
+E: owlgal69@protonmail.com
+W: https://owly.fans
+D: Freedoom website
+
 =======


selliott512:
Although updating the copying right is a pain it may be that the specific date range has some legal value. I know we update it each year at work even though it would be easier to do what you did. Given that my inclination is to leave the copyright be.

A trivial bit of feedback with your credits update - previously there was a blank line before the final `=======` line. I don't know that it matters, but add one back, if you don't mind.

Calinou:
Copyright years can probably be removed or replaced with "Copyright 2001-present". See https://github.com/godotengine/godot/pull/70885.

DynTylluan:
> A trivial bit of feedback with your credits update...

I didn't even realize that I did that (whoops), although I'm not sure why this is even there. [I did ask on the Discord guild](https://discord.com/channels/616148111076884520/616148111076884524/1090412882279288953), but no-one has replied as of this post

I'm not sure if me accidentally removing one would even be an issue, as I think it might just be for line breaks, but if anyone knows why it was there, please correct me

Xindage (approved):
i agree with the -present system as well, while i approve it i'll wait a while before commiting it, just in case if fraggle has some commentary about it.

selliott512:
Let's abandon the "to the present" for now since we haven't gotten anyone with sufficient authority to ok it.

The `CREDITS` change is conflicted because I added someone else, but that's easy to fix.

I'm happy to make both changes and merge if that's ok.

DynTylluan:
Sounds all good to me 👍

selliott512:
Done. Here's what the lost commit looks like so you have a record of it:
```
commit 846f70e50d4d85446d96112e3449bc72be0cc854
Author: Cass "Owly" Python <52046695+DynTylluan@users.noreply.github.com>
Date:   Tue Mar 28 22:48:37 2023 +0100

    Use "to the present" for copyright end year
    
    Use "to the present" in place of a specific end year for copyrights so
    that we no longer need to update it each year.

diff --git a/COPYING.adoc b/COPYING.adoc
index 4b4a3257..f86df5a7 100644
--- a/COPYING.adoc
+++ b/COPYING.adoc
@@ -1,4 +1,4 @@
-Copyright © 2001-2023
+Copyright © 2001 to the present
 Contributors to the Freedoom project.  All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
```

PR #924, opened 2023-03-30
dehacked: fix quit messages.

mc776:
The messages hadn't been checked for length. Fortunately caught relatively early because the only string that shows up in vanilla (and thus would cause a crash) is also one of the longest ones.

I've also amended three things in the text itself:

1. big burly violent guy who's all about freedom and blaming scientists for things is...... not something that's aged well. And when a corporation is involved you know it's the investors pushing for the bad thing and not the actual researchers on the ground.

2. that other quit message literally reads like you're goading someone into suicide IRL, in the current environment where everybody's depressed and a lot of us have lost important people to plague or war or medical neglect or workplace accidents or violent bigotry or any combination thereof. I've taken the liberty to rephrase it so the emphasis is on what I think was intended.

3. the "hey buddy" really wasn't working with the "much" and "truly".

Threw in a replacement for "exit to DOS" as well.


selliott512 (approved):
The English text changes look good to me. I don't know exactly what the impact of the added newlines is (the `\n`s), but I assume that's reasonable.

I can probably merge this soon if there's no objections.

mc776:
Sweet!

The newlines are actually the most important thing - if the line gets longer than the screen Doom crashes.

(And in my defence for not insisting on the *wording* changes before saying the original PR should be merged, things hit different when they suddenly pop out at you and you're not looking at them in a big list of similar things...)

selliott512:
This fixes a crash? I'll just merge it right now then.

PR #925, opened 2023-03-31
Update GitHub Actions "make" script

andOlga:
This script currently throws these warnings:

image

Reading the relevant blog posts, I can see that the CI will just end up being completely broken by June, and this PR is meant to prevent that from happening.

It makes the following changes:
- Switch from `actions/checkout@v2` to `actions/checkout@v3`. There are no real changes in how the action itself is used (at least none relevant to our use case), so it's a simple version bump.
- Changes the `::set-output` command to use Environment Files instead. Other than how the command itself is written, there are no changes to its use, so it amounts to just a small change in syntax.

### DIFF
--- a/.github/workflows/make.yml
+++ b/.github/workflows/make.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         fetch-depth: 0
     - name: Get full repository history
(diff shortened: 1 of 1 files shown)


selliott512 (approved):
As far as I can tell what you've done here is completely reasonable, but haven't worked on `make.yml` before. I came up with the same solution as you by looking at the documentation, it seems to fix the problem, and the artifacts produced look correct. I'll merge this soon if there are no objection.

Issue #732, opened 2021-04-05
MAP11 too similar to Doom 2?


milasudril:
MAP11 contains some locations that are very similar to MAP11 from Doom 2. It is like the original, but larger and somewhat shuffled. In particular, the section behind the yellow door. It also uses the same monsters for the first part. The second part in the original MAP has been moved in FreeDoom to the blue key section. The Arch-Vile appearance is identical to the original map. What do you think, should these parts of the map be reworked?

mc776:
Screenshots?

Calinou:
It's quite similar (and definitely "rings a bell" when playing Freedoom), but it doesn't feel so close as to be problematic.

### Freedoom MAP11 (behind the yellow key door)

![freedoom webp](https://user-images.githubusercontent.com/180032/214812838-9c0e1de4-f164-49c9-9e03-0243b9ff34a9.png)

### Doom 2 MAP11 (room on your right after spawning)

![doom2 webp](https://user-images.githubusercontent.com/180032/214812857-05e35f56-2f8b-44bc-9feb-fa897162e054.png)

selliott512:
I tend to agree. If it were up to me there would not be any particular similarity between any FD map and the corresponding (or any, actually) Doom 1/2 map. This is both a matter of principal (that FD wishes to demonstrate that it is its own independent game) and to minimize risk of intellectual property trouble.

mc776:
Now that I'm taking a closer look... yeah, both the twin towers and the bi-directional bars feel way too much like D2 homages now.

If anyone's got a really good mid-tier-difficulty map in the works, designed from the ground up to be vanilla compatible, this might be high priority to replace.

milasudril:
@mc776 Perhaps you do not need to redo the entire map. I would simply redraw this area. Also, the map has a nearly identical Arch-Vile encounter to that in doom 2. You find it behind the blue door (Sector 144).

mc776:
I'm working on a massive revision to this map. The id-style twin tower room is being overhauled into a fortified entry gate and you do not open the entrance within by going inside, accessing the symmetrical lifts and hitting two switches. That combined with the colour should hopefully mitigate any remaining idness.

The sliding bars beyond that have also been replaced with random machinery to still trip the player up without providing a hard barrier.

The stuff behind the blue key door is replaced entirely (thanks to @Xindage for this re-do almost entirely from scratch) - there's still an archvile but it's definitely not that box reveal.

[...]

PR #934, opened 2023-04-07
levels: revisit Map11.

mc776:
In this I try to address two issues:
1. All the stuff, other than mere difficulty, that makes me always want to immediately idclev/idclip out of this map.
2. The Doom 2-ishness of the blue key branch, as per #732.

One common theme in 1 is to ditch or mitigate a lot of gotchas that are basically beaten by dying and memorizing the sequence on restart (or savescum), in favour of things that remain challenging even when you know it's coming. A lot of mandatory direct backtracking has also been mitigated by opening up optional alternate routes.

The name has been changed since everything in this cluster is named with some diegetic description. I always like to imagine the first three doors as actually being fixed seamless interdimensional portals (with the rest of the "real" facility being the final branch) as an explanation for the seemingly unrelated appearance of each branch.

Note that this adds enough content that it creates a savegame buffer overrun. I am proceeding on the basis that [these are not worth caring about](https://www.doomworld.com/forum/topic/92282-savegame-buffer-overruns/).

Detailed notes on changes in the PR.

**Central hub**

The first three branches may now be completed in any order. What used to be the red door now requires all three keys to open. Accordingly, the SSG (and, outside of easy, minigun) is now available in the starting room and each branch (other than yellow) gives the same colour secret skull as its main key.

The four panels around the entry pad are a bit thinner to keep the player from getting stuck. The entire area is also given some aesthetic tweaks.

The soulsphere secret teleporter is now almost completely hidden until it is opened. The soulsphere is already clearly visible and inaccessible so you know there's a secret there, so there's no point in showing the door at the start.

**Northern branch**

The southeast nukage room crates no longer form dead ends.

Because you already have the SSG at this point, and to prevent situations where the new necromancer from the east branch can be led in here and raise up ghosts, the crusher is now gone. The shotgunners are also placed in a more sensible position (I'm not 100% sure about this one, if anyone wants them oblivious again I'm happy to revert this).

The switch that opens that first big door now makes it stay open. There was nothing of value in forcing that time-sensitive switch-hunt - basically a mandatory sound-based secret - on the way back.

The first backtrack switch is also made accessible by hitting a switch by the skullcube door, instead of simply running towards it. This gives the player a better sense that something has happened and they should check behind them.

That switch on the way back now opens the actual secret. Which solves the problem of there having been *absolutely no visual indication* of that secret door even being there...

The yellow key room is now accessed by a fast lift going down instead of a proper door. This allows the blocking geometry to be something that doesn't leave the user immediately trying to use it on sight, and also provides a more natural-looking way for the player to get stuck in the room until they hit the switch, with less delay.

The first yellow key room switch encounter, the one that releases the combat slugs, now also spawns a monster in the northern gargoyle switch room. Hopefully its presence should alert players to go check it out.

The second gargoyle switch opens up a one-way shortcut leading into the yellow key room. It is also now visible from the beginning and obstructed by bars and distance, so the player knows that something needs to be done.

The slug and spider warp-ins are now missing the escape route blocker in easy mode.

**Western branch**

Lots of little aesthetic adjustments to try to get rid of the *Doom II: Hell On Earth MAP11: Circle of Death (named "'O' of Destruction!" on the automap)* influences.

All the big trees can be fully moved around, instead of having two that have an invisibly impassable gap. They are arranged fully symmetrically, which with the warmer colours and emphasis on the face-like design hopefully should reduce the Doomishness a bit.

The corridor between the towers does NOT branch out into lifts leading up to them. You'll have to find another way inside.

The big skull switch room and the two sewer-like corridors beyond it have been substantially overhauled into a single big lobby. Getting on top of that panel is intentionally marked as a secret since it's just really hard to get up there without getting snagged on the pillar.

The minigunner switch room now has panels separating the minigunners until you enter the inner room. The secret room the less obvious switch opens has been moved. The serps are also moved so you're not just shooting a bunched-up fishbarrel of them when you open the door.

That trick with the invulnerability sphere fake secret crusher has worn out its welcome. It has now been replaced with more monsters, which you can escape with a new teleporter that takes you somewhere less than ideal.

Turned the combat slug blue key monster closet into a second passage back up to just inside the fortification. Nothing being bypassed has subsequent monster warp-ins, nor is any route that opens up accessed by backtracking that way, so making the player go back the way they came is completely pointless.

Replaced some of the illusory walls with actual interactive secret doors. My rationale for this is that in vanilla you snag and bounce a lot when pushing up against walls, so most players (like me) will try to avoid touching the wall and just spam the use key instead, as opposed to a GZDoom player who could freely do both at once.

Made the skull trigger lines actually around the skull pickup, to make it more likely that the player actually picks it up in the course of enabling the door.

The first secret door leading to the yellow skull is now opened by freeing all five lizard babies.

**Eastern branch**

Replaced entirely with my edit of Xindage's remake. The creepy abandoned lab with the lizardbaby has been remade from scratch in the blue key branch; there is still some change in architecture after you pick up the red card but it is more limited.

**Southern branch**

The lowering floor panels are now all 8-above-HAF fast switches. On paper it's actually a little *more* inconvenient than it used to be since you now have to press one more button, but this relentless "fake door trick you into moving towards it before doing something else entirely behind you" bullshit is frankly giving me clinical depression when I look at this map.

Realigned all the SUPPORT3 textures on  the pillars on the walls of the first arena.

Aligned textures near the entrance into the final sniper arena and the minigunner platforms behind it. (Said platforms have been lowered to 64 from 80.)

The final arena lighting is now under 255 to allow the darken-with-distance effect to show so that everything doesn't look so flat *in that specific part of the map where gauging distance is especially vital*, with minimal interference with the [overall aesthetic](https://www.youtube.com/watch?v=Fb1JXEK_Q5M).

There is literally no possible way that a person can figure out, given only what can be seen ingame without cheating, that walking over those lowered pillars opens the route to the final skull key. It's now a Map01-esque switch behind the exit sign. That entire final skull key area has also been redone from scratch.

The exit pillar switch/BFG secret wing has also been significantly (>90%) overhauled.

Exit pillar tops are now brown, use bronze textures for better tiling, and are gathered tighter around the exit. Side effect: combine this with the above and now you can access the super skull secret before having hit the switch, resulting in less back-and-forth. This is a feature.

The three super-secret doors now totally fail to open if you haven't done the thing, instead of getting stuck part way through.

All barriers to normal progress can now be opened from the other side. You should now be able to get back to the start starting from any valid point in this branch not part of the BFG secret.

The exit lines are now monster-blocking, to prevent visual oddities where a critter is standing on the teleport pad but not going anywhere.


mc776:
just realized I posted this on Good Friday and wondering if I really should have gone with the original name idea "Rift Station Golgotha" for all the skulls...

selliott512:
Nice. You've put a lot of work into it. I'm going to play through it and try to give you some feedback tomorrow.

mc776:
Summary of visual changes since last screenshot post (new METAL2 textures not included in this PR):
doom00
doom01
doom02
doom03

selliott512:
I've played through this a bit. Nice. You've added a of detail. The new parts are fun to explore. I like it aesthetically. Some observations that may or may not be helpful:

- Using the lift in sector 340 it's possible got get passed the blue door in sector 327 without having the blue key.
- The lift in sector 57 immediately lowers when the player rides it up, presumably because of the lack of headroom at the top. If you mean for the player to jump off at the correct moment then that's fine.
- The final door South of the player start was not obvious to me. I thought the yellow, blue and red stripes were just static marking indicating that it should open when I have those keys. After I looked at it in Eureka I understood it was a switch requiring all three keys.
- It seem significantly harder to me than it used to be with lots of hard monsters and not so much ammo, but that may be because I'm not as familiar with it as the previous MAP11, so I only found 5 of 15 secret areas.
- At the end I had found all three keys, the blue skull key, and I got all but one kill.

Xindage:
I'm a above average skilled player, and the difficult of this map for me was honestly ok, I don't even find it hard at all.

mc776:
People's thoughts on this?

Instead of the gargoyle being the switch for the lift itself, have it be used only once to make the lift accessible from the "normal" side and the real lift switch in a sane location?

DOOM00
DOOM01
DOOM02
DOOM03

mc776:
Disregard the above suggestion. I've just been reminded that a similar remote down lift switch exists in the west branch and the way it's set up now is really part of the map's UI language.

Besides, adding that extra stage really makes things way more complicated.

Talon1024:
I was only able to find 15/16 secrets. One of the secret sectors is Sector 58, which cannot be registered because it is a monitor within one of the northern secret passages (Sector 675).

mc776:
Confirmed this branch builds correctly and I can now 100% the secrets on vanilla.

Now that I've finally settled for myself the issue of what to do with those two switches by the big gate fortifications, unless someone can spot a new issue I think this one's ready to merge.

mc776:
I have no idea what the hell I've just done.

Someone please squash this properly before merging.

Xindage (approved):

PR #936, opened 2023-04-12
sprites: remove redundant necromancer angles.

mc776:
Also fix the offset in frame X.

This can be revisited if and when someone actually does the new angles for the necromancer - until then we can save a bit of bandwidth and space, and have a minor QoL improvement for anyone who's trying to browse the sprites.


Ferk:
Would this cause trouble when loading a PWAD with an archvile replacement?

Xindage (approved):

PR #938, opened 2023-04-14
levels: address map29 artificial difficulty.

mc776:
If you pistol start this map, on the teleport into the tower you will be forced to get helplessly shredded by serpentipedes while you're switching to the polaric that's right at the destination. A second one has been added on the platform with the backpack that leads to the teleporter, and the current one has been flagged as multiplayer-only.

Right after this is the blue key, which cannot be retrieved without taking damage from the hurtfloor below. The way the pit is set up, having two teleport pads is unnecessary except to increase the chances of a spider warping into the hub which never made that much sense anyway. To address all these issues, there is now only one pad, right next to the rocket launcher switch, which lines are monster blockers, and the blue key platform is now *higher* than the ledge you reach it from so you can see it moving, retreat back into the tower, and go back to the RL platform and reach the telepad without touching the lava.

The medikit with the blue key has been moved into the main ground area in the northwest.

The jump from the wood tower onto the teleport to get into the stone tower is now surrounded by monster blocking lines.

Got rid of two things that seemed intended only to spite the player:
- the sub-shotgun in the green room, which would cause a weapon-switch just as you face the painlord at the top of the lift, is now just two packs of shells.
- all four doors of the minigunner shack are usable. (They are also equally lit, unlike previously when one of them was slightly darker and it wasn't the one that was usable.)

Added a radsuit and medikit to the back of the green building, so that exploring that area isn't merely unmitigatedly punished.

Added a visible light source to each of the indoor areas.


mc776:
I just realized that you can run between the serpentipedes when you warp in there.

But by that point you'll have taken extraordinary amounts of damage from the hurtfloors and pinkies and serps and getting blocked by movement at the wrong times, so it does seem like one of those "die and retry" gotchas i've been trying to fix in #934.

Ignore the 10 seconds that this PR had been marked as draft, unless anyone has any really good ideas or urgent bug reports it's good to go.

Xindage (commented):
I've tested but i still find it kinda too hard for my way of playing, i would also consider at least one suit to spawn in multiplayer for the sake of it, 3 or more players fighting in this small area is going to make this extra tought.

Xindage (approved):

Issue #109, opened 2014-10-02
Python 3 support


chungy:
Eventually Python 2.7 is going to become less common or decreasingly used as the default Python interpreter for many installations and distributions (Arch Linux already uses python3 by default), we really ought to upgrade all of the programs part of the build system to be compatible with Python 3.x, hopefully while still supporting Python 2.7.

Right now, this isn't too critical as it is not hard to just use virtualenv-2.7 to force Python 2.7 to be used by default in your environment, but it's an extra step that ideally needs to not exist.

chungy:
This was fixed a long time ago. :)

jmtd:
> This was fixed a long time ago. :)

Except, all your shebangs are still `/usr/bin/env python`, so on a pure Python 3 system, they will not work, with the exception of Arch, which installs `python3` with the name  `python`, against upstream's advice.

You probably want to move the shebangs to `python3`.

chungy:
Our scripts have been made to be compatible with both 2.7 and 3.x, and that's perfectly in accordance with [PEP 394](https://www.python.org/dev/peps/pep-0394/). I wouldn't be bothered much if we want to go 3-only, but I don't think there's any advantage.

fragglet:
Looks like Fedora also just switched `/usr/bin/python` to point to `python3`.

https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/ZXP6BQWXRTVJRVUTRWNMUTLLSQGPHCAS/

PR #941, opened 2023-04-16
dehacked: correct Thing #109's height.

mc776:
Here's what it looks like *without* the fix:
deadsargeoffset
The problem also exists in id, so we can trust that any mod that deals with this object will also be dealing with the height.
DOOM00

### DIFF
--- a/lumps/dehacked/dehacked.txt
+++ b/lumps/dehacked/dehacked.txt
@@ -38,6 +38,15 @@ Duration = 4
 Frame 48
 Duration = 3
 
+# Things 104 and 109 both hang from ceiling and use the same sprite,
+# but 109 (the non-blocking one) is height 52 instead, causing its
+# sprite to clip into the ceiling since the sprite offset is meant
+# for thing 104's correct height of 84.
+# This sets the non-blocker's height to match its counterpart.
+
+Thing 109
+Height = 5505024
+
 # This is a magic comment recognized by Chocolate Doom, so that the
 # BEX [STRINGS] section below will be parsed:
 #


fragglet:
I think this changes game behavior? Assuming so, I'd rather not do this - the other dehacked changes so far have been purely cosmetic. Can we not use sprite offsets to fix this?

mc776:
Both actors use the same sprite, so if you change the offset to fix the non-blocker the blocker will be too low.

Right now the non-blocker's useless as a decoration for anything that isn't carefully tweaked to obscure the mistake, so i wouldn't be surprised if no maps used it (or the only maps that used it did their own thing to change that actor's properties).

One possibility is to actually make the non-blocker use a different sprite...

mc776:
> One possibility is to actually make the non-blocker use a different sprite

From what I can tell there's no way to do this in DeHackEd, at least in vanilla.

I can't define one of them to use a brand new frame pointer, and they both use the same pointer which means I can't just set one to use GOR3B0/frame 1 and make a new sprite with that name.

Changing the non-blocker's size to match its blocking equivalent is the least disruptive way to address this (and IMO simply fixes id's actual honest mistake).

mc776:
As far as gameplay effect is concerned, this is a non-solid, non-shootable actor.

I need to check if this actor is normally capable of blocking a lift or a door or a crusher, in either chocolate, prboom-plus or gzdoom - i don't *think* it will in any of these but I've been wrong before.

Xindage:
fraggle what is your techinical point of view about this issue? i dont want o add something in dehacked that may cause some issue.

mc776:
*Theoretically* this change should not cause any problems, but given the way (e.g.) the floating-ledge-items bug tends to be triggered in vanilla during collision checks that seem to have nothing to do with the floating items it seems best to err strongly on the side of caution.

I'm sick of seeing this so I'm going to close it instead of running the extensive demo/multiplayer tests that would be needed to prove a negative that this is *not* ever going to be a problem.

PR #946, opened 2023-04-27
patches: new WALL25_1 by Korp.

mc776:


Calinou:
This will trigger trypophobia for sure, but I like it nonetheless :smile:

axredneck:
It is horror FPS after all, it's supposed to be creepy.

mc776:
In all seriousness though that was one of the reasons why I hadn't posted this PR much earlier.

So far everyone mentioning it seems to just be bringing this up theoretically but has it actually affected anyone?

Xindage (approved):
Committing

Issue #932, opened 2023-04-04
sprites: polaric gun flash should be reconsidered.


mc776:
Right now the flash frames have an irregular cloud of blue gas coming out of the muzzle. This would look perfectly fine if we were dealing with any of the other guns where these are just brief (and consistently alternating, e.g. minigun) frames that imply movement, but the way the plasma works (each shot displays one of the two frames at random and the frames last long enough that they never disappear or flicker while you're shooting) you'll often end up seeing the same frame unmoving for upwards of half a second at a time.

It would be good to design some kind of glow effect that still looks okay when it's "stuck" in the same frame for a while.

Issue #240, opened 2016-02-06
Weapon firing frames that are not lit correctly


AXDOOMER:
Somehow, the light reaches the back of the pistol. This violates the laws of physics.
![doom00](https://cloud.githubusercontent.com/assets/6194072/12869882/05528de4-ccfa-11e5-8900-91034757f9d7.png)
The SSG is not lit at all.
![doom01](https://cloud.githubusercontent.com/assets/6194072/12869883/055a16f4-ccfa-11e5-8fb7-477d1ba1d727.png)
These two are fully lit:
![doom02](https://cloud.githubusercontent.com/assets/6194072/12869885/0592ce04-ccfa-11e5-9c59-ec11ec3acc16.png)
![doom03](https://cloud.githubusercontent.com/assets/6194072/12869884/0569e14c-ccfa-11e5-9e3c-7921bdc7e73b.png)

Xindage:
actual state of this issue, pistol:
![image](https://github.com/freedoom/freedoom/assets/6380147/e406ff95-56a2-4c9e-ab0e-52294f9b6e24)
Shotgun:
![image](https://github.com/freedoom/freedoom/assets/6380147/1d27a53b-de70-4687-acb3-cd1c0808941f)
![image](https://github.com/freedoom/freedoom/assets/6380147/71d0b68e-2c9d-4a1a-a742-1f304e6ee0e3)
chaingun:
![image](https://github.com/freedoom/freedoom/assets/6380147/c2d74d62-07e3-4bad-b630-c5a79c98696d)
![image](https://github.com/freedoom/freedoom/assets/6380147/66b1edb5-b15a-416c-87b1-42dfd38c2552)
Rocket:
![image](https://github.com/freedoom/freedoom/assets/6380147/bb7d8cd1-614b-4f19-9952-26d47eec6800)
![image](https://github.com/freedoom/freedoom/assets/6380147/490c6bd4-420c-4e17-8799-7f76b0f0337d)
Plasma:
![image](https://github.com/freedoom/freedoom/assets/6380147/3cf7fb7b-9323-48a0-8fc9-ea8ca2ae0552)
BFG:
![image](https://github.com/freedoom/freedoom/assets/6380147/8235b0f8-ac5d-4bd4-9f0b-675b22e57445)

Xindage:
someone still think that we still need to update something about this? bfg and plasma dont follow the rules as well, but i dont know if worth the effort...

mc776:
[...]

PR #947, opened 2023-04-27
sprites: polaric upgrade; sprite offset fix.

mc776:
The current pickup sprite looks pretty much exactly like what one might expect from drawing fanart of the id plasma rifle from memory. I've changed what I could while still remaining consistent with (and hopefully becoming more consistent with) the HUD sprite.

Since the gun shoots straight forwards I've also made the flash symmetrical - along with the projectile itself (but not *too* symmetrical). This does help address #932 a bit, but if anyone's got a better design I'm all eyes.

The blue flare along the top "spine" of the gun looked a bit janky on the left so I've smoothed it out a bit. The blue also lingers a bit in the B frame.

With all the ongoing news a lot of FPS player types will be following about artillery logistics in the IRL war, the use of "cannon" referring to this short handheld PDW is aging even more poorly than Doom's "clip". The best word I can think of as a substitute happens to result in an acronym I think we might as well lean into.

I've also adjusted some other projectile sprite offsets - my last attempt at fixing them actually placed the sprites a little too high.

screenshot preview, before and after:
DOOM01
DOOM00


mc776:
Attempts at #240 did not go well. The problem is that the pure blue glow takes a lot of tweaking to even begin to not look wrong with the palette we've got, and creating a border of lit and unlit pixels exacerbates all the problems in ways we can't really work around without changing the entire art style for this one weapon alone.

I hope I've at least gotten rid of a bit of jank down the barrel in the B flash frame with this update though.

Xindage (approved):

PR #948, opened 2023-04-28
New Gray/ick wall by korp.

Xindage:
As well new ones using stock patches.
image
image
image


Xindage:
waiting someone else approval on how it looks in the maps who already use grays, do you guys like it or not?
Unfortunately there's no map who uses this graytan, its a new collection of textures that will come with this PR.

korp has a test map that showcase them: [doodoomap.zip](https://github.com/freedoom/freedoom/files/11361112/doodoomap.zip)

mc776:
I say let's merge it now and let people eventually start organically using the new textures, worst case scenario they can be deleted later if like years later no one's found a use for them.

Personally I really like how that first screenshot immediately conveys a vibe that is impossible to confuse with id Doom.

Xindage:
ON HOLD FOR 0.13

mc776:
This textures stuff is over my head.

Would it be worthwhile to close this, do a PR with only @KorpKat's new patches, and then post a new PR with just the textures lump?

mc776:
I've manually added the changes from #962, #1012 and #1032 to the textures file.

I've been autoloading a patches-only version of what Korp posted on Discord and I see no reason why we shouldn't add the grey concrete patches at least, and haven't noticed anything off with anything else.

mc776 (approved):

mc776:
THAT SAID:

> ; Note from xindage: "This is a test, if it pass whitout causing any bug
> ; i'll make try to follow up this rule to revise this file"

is there anything still outstanding?

Xindage:
the test aside like i told already 3 times now, i dont want implement something so heavy before a release, i want to apply it after a stable release, several maps use grays walls, i dont want make things looks different sudden before the release.

PR #953, opened 2023-04-30
levels: minor aesthetic changes that didn't make it into #949.

mc776:
**E2M3**

A bunch of door track textures needed lower unpegged flags.

Lines 1179 and 1181 also given DOORTRAK instead of SUPPORT3 because every time I look at it I feel like the door should get snagged on those bumps. (Same with the blue key lift.)

The tekwall trim railings in the northeast have been raised to prevent accidentally walking onto them. I've also given them a texture that does not look like a misplaced door. The lowering wall monster/switch closet has also been touched up a bit.

**Map04**

Changed the northernmost texture in the hatchling closet in the final room so it doesn't contrast so much that it looks like a secret.

Slightly moved two vertices on the exit flame to get rid of the fake contrast.

Fixed up the central hub aesthetics a bit:
- Aligned the pillar textures.
- Expanded the teleport destination pad, gave it some ceiling lighting, and surrounded it with a grate floor. Now it looks like an actual part of the bigger thing.
- Lowered the contrast between the lights and shadows.
- Got rid of the featureless silver above the doors, giving them tekgreen lights instead.


mc776:
DOOM01
DOOM02

Xindage (approved):

PR #956, opened 2023-05-01
levels: Rebuild nodes and Eureka fixes

selliott512:
My usual rebuild-nodes and Eureka fixes.


mc776:
Just checked the 3 changes, things look fine.

What's the difference with the node building though? I'm using ZenNode as well with these and I don't get why they'd all end up different.

selliott512:
> Just checked the 3 changes, things look fine.

Glad to hear it. I'm generally pretty careful and focused with the Eureka cleanups. When it's small, like this, I'll even do it twice to make sure I get exactly the same file when I'm done.

> What's the difference with the node building though? I'm using ZenNode as well with these and I don't get why they'd all end up different.

Good question.

One of my assumptions, based on my experience, which may be incorrect, is that ZenNode always produces the same output for the same input, even when slightly different versions of ZenNode are used by different people on different computers. What I'm doing is just:
`make rebuild-nodes`
which is just:
`ZenNode levels/xxx.wad -o levels/xxx.wad`
for each level file separately. How does that compare to how you are invoking ZenNode in your editor or otherwise? Are you passing any options to it?

If you run `make rebuild-nodes` now with these changes (or current latest master, either is interesting) does it produce any file differences for you?

I was going to suggest that maybe you submitted something on someone's behalf, but I see that's not the case for at least E4M1 since the last change before these changes was you rebuilding the nodes:
```
commit 7e8677783231bb80fe93ea879a8a516bd3c04fad
Author: mc776 <24984517+mc776@users.noreply.github.com>
Date:   Fri Apr 14 17:00:02 2023 -0700

    levels: build e4m1 nodes. (#940)
    
    Addresses #939.

M	levels/e4m1.wad
```

As we get close to a release I think it's ok to rebuild the nodes as long as there are not any levels in flight (as long is at does not cause conflicts) since it's the easiest way of making sure it's done even if some of them were already done correctly, but slightly differently. If it turns out to be sub-optimal then we can focus on who did it last (me in this case), what's up with that person's ZenNode executable and version, and how it can be fixed, rather than trying to figure out how multiple people did it.

Somewhat of a tangent - in a way ideally (but not really) the lumps produced by the node builders would not be checked into git. Instead they would be like executables - something produced by the build. But that would mean having levels files without those lumps, which would break Chocolate Doom, which we don't want to do.

mc776:
I make sure I'm in my working levels folder just so I can forgo typing any options after `ZenNode xxx.wad`.

selliott512:
> I make sure I'm in my working levels folder just so I can forgo typing any options after `ZenNode xxx.wad`.

Well, that's nice and succinct. My `-o levels/xxx.wad` is in the parent directory seems to produce the same output, so it's not that (I tried both ways with `e4m1.wad` and I got the same result).

Would you mind sharing some whatever information about your ZenNode version and OS you are comfortable sharing? For example, in my case ZenNode outputs its version with each WAD processed:
```
ZenNode Version 1.2.1 (c) 1994-2004 Marc Rousseau
```
and my OS:
```
Fedora 33 64 bit
uname -a
Linux cube 5.9.10-200.fc33.x86_64 #1 SMP Mon Nov 23 18:12:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

mc776:
ZenNode 1.2.1
Debian Bullseye
Linux 5.10.0-20-amd64 #1 SMP Debian 5.10.158-2 (2022-12-13) x86_64 GNU/Linux

selliott512:
That is a mystery. I tried grabbing `e4m1.wad` from the parent of your `levels: build e4m1 nodes.` commit, which I would guess is the input for your node rebuild:
```
git checkout 7e8677783231bb80fe93ea879a8a516bd3c04fad~ -- e4m1.wad  # Notice the "~"
sha1sum e4m1.wad 
b189867575273fa0f1ac9898cac9a2195571dfc2  e4m1.wad
ZenNode e4m1.wad 
sha1sum e4m1.wad 
c31dbbd4b13d69e94002c18666d66cc77cbb274d  e4m1.wad
```
That last hash matches what I have in this PR right now, but not what you have in that commit.

I even tried starting a Ubuntu instance in podman, but I couldn't find `ZenNode` in `apt-get`, so I just ran the `ZenNode` executable from my Fedora instance, and that produced the same output that I got from my Fedora system.

So, either the Ubuntu ZenNode produces different output than Fedora ZenNode even though it has the same version, you had input that was different than the parent of `7e8677783231bb80fe93ea879a8a516bd3c04fad`, or ZenNode is not as deterministic as I hoped.

If you feel like looking at this further it would be interesting if you attempted to recreate the `e4m1.wad` you checked in:
```
git checkout 7e8677783231bb80fe93ea879a8a516bd3c04fad~ -- e4m1.wad
ZenNode e4m1.wad 
```
And see if you get the same result as before.

mc776:
Same end result as you now it seems:
```
$ sha1sum e4m1.wad;ZenNode e4m1.wad;sha1sum e4m1.wad
b0a278ea311a63e532f021f3da80987b3cd73fbd  e4m1.wad
ZenNode Version 1.2.1 (c) 1994-2004 Marc Rousseau

Working on: e4m1.wad

 *E4M1    : BLOCKMAP - 13452/13452 (100%)   Compressed:  73%      0.001 secs
            NODES -  968/973  ( 99%)  SEGS -  3141/3165  ( 99%)   0.083 secs
            REJECT - Efficiency:  83.6%/83.6%  Sectors:   327     0.091 secs

Saving to e4m1.wad...

1 Level processed in 0.175 seconds - 1 Level needed updating.
5.714286 levels/sec
c31dbbd4b13d69e94002c18666d66cc77cbb274d  e4m1.wad
```

It's probably the ZenNode I've got on my Windows machine that's the difference - i think that emergency fix was done while I was at work.

Xindage:
Should i commit this now?

mc776:
I think you should.

Xindage (approved):

PR #969, opened 2023-05-10
Change the last colour of playpal, and change build system to allow palette modifications

### LINKED ISSUE #48: Some sprites have speckles when playing with D2TWID

Talon1024:
Should fix #48


mc776:
I still don't understand why it would be preferable to make colour 255 a copy of another colour rather than a whole new one that no current sprite uses. You still can't use that colour slot anyway and it becomes less clear what's going on.

Talon1024:
So that other graphics or future submissions don't use that colour, and so that changes to the palette (like the blue range I showed on Discord) don't use that colour in darker shades of the `COLORMAP`.

mc776:
Got it.

Can you update the title to the PR though? When I first saw the notification I got terribly confused thinking there were two separate PRs.

Talon1024:
I just found out that changing the last colour of the `PLAYPAL` can break some graphics in vanilla mods. So in order to remain compatible with the aforementioned mods, we need two palettes: one which Deutex uses to convert graphics to, and another one to replace the `PLAYPAL` in the Freedoom IWADs after the graphics have been converted.

Doing it this way will also allow the second palette to use new colours (as long as they are "vanilla-compatible") without the need to change any of the existing graphics. A win-win situation, I'd say.

mc776:
I see there's a  whole new python script for this... this doesn't require the user to install any new dependencies to build, does it?

Talon1024:
No. I don't think the CI build would have succeeded otherwise.

Xindage:
so, the pallete base of freedoom is still untouched? just the build process that work differently?

Talon1024:
I changed the build process to use the new python script to replace the `PLAYPAL` inside each IWAD with playpal-new after each IWAD is built. This allows playpal-new to be improved, without having to change all of the existing sprites/patches/flats. However, playpal-new is currently the same as the old playpal-base.

When Deutex builds the IWADs, it uses playpal-base to convert images to Doom's image/flat format. I changed playpal-base by copying another colour to the last colour so that the last colour wouldn't be used.

selliott512:
Have had a chance to look at this more deeply.

First, my impression is that the idea is that `playpal-base.lmp`, which is meant for internal use, has had color 255 (the last color at index 255) changed from its usual value of a76b6b to cb6b6b so that it matches color 24. Later, as a final build step, `PLAYPAL` is replaced with the usual standard Doom `PLAYPAL` so that color 255 is back to its usual a76b6b value.

When a sprite, or any PNG, has color a76b6b (color 255 before this change) Deutex has to pick another color (since at the time it creates the WADs it does not have that color). I think our hope is that it would pick color 24, but based on my testing it picks color 133, which is a tan color. Not ideal, but still better than cyan.

The `replace-lump` script you wrote in pretty neat. It does however have a side effect that may or may not be a problem. The WAD files produced by Deutex align lumps on multiples of four by adding bits of unused space at the end of the proceeding lump, which the `replace-lump` script does not retain. For example, consider the offset of the first `VERTEXES` lump. It's a multiple of four before `replace-lump` script, but not after. I'm not sure how much that alignment matters, but you could have a simple script that just finds the offset of the existing lump and replaces just that portion of the file with the new lump. In this simpler script it would be an error for the new lump to be a different size than the old one (so you don't have to update the directory).

Also, `git status` should return cleanly after a build:
```
$ git status -sb
## pr-969
?? lumps/newpal.lmp
?? lumps/playpal/newpal.lmp
```
which can be fixed like so:
```
--- a/.gitignore
+++ b/.gitignore
@@ -59,12 +59,14 @@ doom.gpl
 /lumps/freedm.lmp
 /lumps/freedoom.lmp
 /lumps/genmidi.lmp
+/lumps/newpal.lmp
 /lumps/playpal.lmp
 
 /lumps/colormap/*.lmp
 
 /lumps/genmidi/*.lmp
 
+/lumps/playpal/newpal.lmp
 /lumps/playpal/playpal.lmp
 ```

It might be possible to come up with different names than "base" and "new". What do you think of "internal" and "external"?

We can do this, but my inclination may still be to have a script that goes through the PNGs and replaces color 255 with 24 that we run periodically, or at least before releases. I can work on it.

selliott512:
@Talon1024 , are you still interested in pursuing this? How do you feel about my previous comment?

The immediate issue of color @255 was fixed by my PR, but it's still possible to do something like this.

Talon1024:
I'm still interested in having a way to improve or change the current palette, but since `map-color-index` was added, I'll need to re-think this whole idea.

Maybe I'll write a script which takes a PLAYPAL lump/.act palette file as an argument, compares the current and new palettes, and replaces the old palette file with the new one, in addition to re-colouring the existing graphics with `map-color-index`.

PR #971, opened 2023-05-11
Incorporate MIDI pack for Double Impact

northivanastan:
The fourth episode of Phase 1 is currently reusing music from the previous three episodes due to how E4's soundtrack worked in The Ultimate Doom. However, in some source ports it is possible to replace those tracks. Hence, for a while I've been stewarding [a community soundtrack replacement for Double Impact](https://www.doomworld.com/forum/topic/135660-release-double-impact-midi-pack-a-community-soundtrack-for-double-impact-and-freedoom-e4/). I have made sure that all contributors dedicate their work to the public domain (and released it with CC0), so it can be included in any project.

Currently this works by adding D_E4M(1-9) lumps to the build, which will be recognized by Crispy Doom. This is not ready to be merged yet, because the pack is in RC status, and there are still some questions that need to be answered:

- Is it okay for Freedoom to include this content at all, when not all source ports will support it?
- Should we include MAPINFO to expand support for other source ports?
  - Which dialects?
  - Should they cover the entire WAD or just the changes to E4?
  - If they will cover the entire WAD, should scripts be written to automatically generate them from/alongside DEHACKED?


mc776:
As far as question 1 is concerned, on the one hand it is nice to keep things consistent between sourceports, but on the other this would hardly be unprecedented given the quit messages and the player crouch sprites and the combo key/skull status bar icons and whatnot.

northivanastan:
Update: I no longer consider this MIDI pack a release candidate, so I'm entirely fine with this being merged now.

Another idea: If this MIDI pack shouldn't be incorporated, perhaps the documentation and website could reference it as an optional add-on, and link to the /idgames or Doomworld page.

mc776 (approved):

PR #972, opened 2023-05-11
levels: redo Map09 warehouse area.

mc776:
I've complained about the texture alignment of the crates before. I tried to fix it up, but then I got to the smaller dark areas and realized the temporally longest parts of the map were a great big case of terminal One-Sector Twisty Dark Maze Syndrome. So I gave up trying to make all the crate textures consistent, in favour of trying to recapture the interesting ideas behind this area with a brand new implementation.

- A big warehouse area that has you take a specific path at first, but then opens up after you get the key.
- The key is on the east end, behind a big door, in a big square thing. (Not quite the key itself now, but necessary to reach the key - or, rather, reach the big open landmark in the east.)
- The way back involves finding something in some narrower, darker areas to the north and south.
- There are four secrets around the darker areas. One berserk, one plasma, one rockets, and one armour+backpack from a guy who died tragically alone in a dark storage area.
- Fairly constant shooting.

The whole place feels a lot less like Doom Containment Area which is a nice accidental plus.

One small change outside of the warehouse: the monster closets by the hallway no longer open while you're in the warehouse area, but only while you're already walking down towards the exit. (I've also moved the fake door in the final room to line up with some textures better, and replaced the textures on the stray crates in that T-corridor, but those have no effect on gameplay.)


mc776:
Sample screenshots before and after for roughly equivalent parts
doom00-1
doom00-2

doom01-1
doom01-2

doom02-1
doom02-2

mc776:
Summary of what I did to the red key room:
DOOM00
DOOM01

Issue #976, opened 2023-06-04
Some issues in episode 1


nful2:
Version used: freedoom-0.12.0-486-g4b002550
Source port used: gzdoom v4.10.0

1. e1m9 - In megarmor area, if you leave some enemies above the lift, you are unable to go up the lift due to the enemies blocking it. You are able to go up when enemies move away a bit. I think the enemies are able to control the lift.
2. e1m6 - Zombieman is stuck near red key
![e1m6](https://github.com/freedoom/freedoom/assets/100768714/0825b98e-9d62-4bc7-a7a0-1cb87aa9051f)
3. e1m7 - Map only shows 8 secrets, but tally shows 9 secrets. On my first attempt on easy, i was not able to tag one secret (e1m7). Though i was able to do it in second run. BUT in the main release (v0.12.1), the tally shows 8 secrets.
![e1m7](https://github.com/freedoom/freedoom/assets/100768714/47e6a699-02e9-4c99-8f0f-59ea8e110992)

mc776:
1. Do you mean here?
![_](https://github.com/freedoom/freedoom/assets/24984517/479414cd-4bef-4ff4-b3ca-6859a8f1d267)

nful2:
@mc776 yeah, in here. You can get stuck here if enemies are above the lift.

nful2:
Would recommend to add this in to primary objectives for 0.13 release

Xindage:
fixed by #978

PR #978, opened 2023-06-04
levels: minor Phase 1 fixes.

mc776:
E1M6
Address #976. The diagonal crate at the bottom of the nukage pool was too small and the zombieman was stuck. It is now a slightly larger crate using the big crate textures.

E1M7
Address #976. The bars around the rocket launcher were wrongly marked as secret.

E1M9
Address #976. The walk-over lift line now blocks monsters, preventing from trying (and almost always failing and just annoying everyone) to chase the player into the dead end with the armour. (I had considered raising this dead end and removing the fence, letting the player jump back into the main courtyard, but we'd get the unseen pinkie jump block problem again.)

E2M2
Make the exit into a more E2-typical FIREBLU gate.

E4M1
Make That Room look less like someone tried to draw E1M3 from memory.


mc776:
setting this to draft pending a look at points 14 and 19 on #979

mc776:
For those who aren't on Discord here's the screenshots of the E1M8 changes again

Xindage (approved):

PR #981, opened 2023-06-11
levels: improve e1m7 signposting.

mc776:
Nearly all remote doors with switches more than a room away are now either the big brown door, the big AGM door, or the small tan door with the tan stripe going down the middle.

The first lift into the maze now has something eye-catching on the side.

Fenced-in area around blue key no longer has any invisible blocklines. You can jump down into the nukage if you want - the sequence break won't trap you, even if you fail to pick up the blue key here, and there's no reason in principle it shouldn't be considered a legit way to get through this 3-key, only-yellow-obligatory-for-exit map.

The bars around the armour near the blue key are now the correct silver bars.

The switch to lower the yellow key is now much, much more visible.

The red carpet switch that opens up the exit room is now turned around so you are facing the general direction of the door that it opens.

The windows around (1240,2320) now use a glass texture instead of being invisible and impassable.

That vent shaft bit is now dark, and its doors are no longer the "this is not a real door" texture.

The lift at the northern centre now has its down button right by the platform instead of some ways down the hall from it.


andOlga:
I've played through this and am quite happy with the changes, the map now definitely feels way less confusing to navigate.

mc776:
oof, just notice a HOM, don't merge this yet

mc776:
Before and after for the first/second door fix:

Calinou (approved):

PR #983, opened 2023-06-17
build: Add fix-pngs make target and scripts

selliott512:
Add scripts:
  fix-deutex-pngs - Apply the output of deutex to the build
  map-color-index - Map from one color index to another

Script fix-deutex-pngs is to standardize PNGs by applying the result of the expanding IWADs produced with deutex to the build. It is invoked by build target fix-deutex-pngs

Script map-color-index can map from one color index to another. Most likely it will be used to map form legacy transparency 255 to similar color 133, which can be done by build target
fix-legacy-transparency-pngs.

---

Here's my attempt at fixing the D2TWID / color 255 transparency issue.

This MR adds make targets and scripts that help with, but does not itself, do the following things:

1. Standardize the PNGs by extracting them from deutex (`fix-deutex-pngs` make target).
2. Map color 255 to 133 (`fix-legacy-transparency-pngs` make target).

Both of the above can both be done by running the `fix-pngs` make target, or even `fix`.

The reason why step 1 is being introduced at this time is that it's hard to safely search and replace colors in the PNGs without first getting them in some sort of standard form since otherwise the palettes can be different. Step 1 gets rid of things like `gAMA` chunk, and other color correction, which only causes confusion without any benefit since such chunks are not supported by deutex or Doom.

There are PNGs that are not indexed at all, some that have a truncated palette, and some that have the wrong palette all of which make it hard to work with those images - it's less obvious what they will look in the game since deutex picks what it thinks is the best color.

One other PNG chunk that we should talk about is the `grAb` chunk, which stores the offsets. However, the offsets specified in `buildcfg.txt` take precedence. When extracting deutex sets the `grAb` chunk to the value actually used, which is the `buildcfg.txt` value. This means that step 1 will make all of the PNGs consistent with `buildcfg.txt`, which means they will be consistent with reality. However, if we know there are cases where there are PNGs where the `grAb` chunk has the correct offset, and not `buildcfg.txt`, then we should run script `fix-gfx-offsets` now to go the other direction (PNG to `buildcfg.txt`) for those PNGs before that information is lost (but will still be in git history).

Base on at least the technospider we definitely want to replace color 255 with 133, not 24. Here's the original problem:

before

Here it is fixed with color 24 where faint orange freckles are visible:

after-24

Here it is fixed with color 133 which is perfect, as far as I can tell, giving it that completely blemish free warrior potato look that I think we are going for.

after-133

The above is based on my simplifying D2TWID down to just the relevant color lumps, which I put in [this ZIP](https://github.com/freedoom/freedoom/files/11531044/color-255-wads.zip), which can be run like this:
```
chocolate-doom -iwad wads/freedoom2.wad -file d2twid-color-only.wad technospider-only.wad -warp 1 -skill 4
```

[This page](http://www.colortools.net/color_matcher.html) also suggests that color 133 best matches 255, which is also what deutex picks by looking for the closest [R, G, B] vector.

I think these scripts (this MR) can be safely applied now, but I understand that people may have some reservations about actually using them since it touches 2,825 PNG files.

With some software projects there's resistance to doing systematic cleanup that touches lots of files since it makes backporting (merging) to old branches harder, but I don't think that's an issue with Freedoom. I think if anything images, or other files, will simply be copied to previous branches when and if there is ever a reason to do so.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ DEUTEX_BASIC_ARGS=-v0 -rate accept
 DEUTEX_ARGS=$(DEUTEX_BASIC_ARGS) -doom2 bootstrap/
 NODE_BUILDER=ZenNode
 NODE_BUILDER_LEVELS=e?m? dm?? map??
+LEGACY_TRANSPARENCY_INDEX=255
+LEGACY_TRANSPARENCY_REPLACEMENT=133
 
 FREEDOOM1=$(WADS)/freedoom1.wad
 FREEDOOM2=$(WADS)/freedoom2.wad
@@ -18,7 +20,7 @@ FREEDM=$(WADS)/freedm.wad
 
 OBJS=$(FREEDM) $(FREEDOOM1) $(FREEDOOM2)
 
-.PHONY: clean dist
+.PHONY: clean dist pngs-modified-check
 
 all: deutex-check $(OBJS)
 
@@ -55,6 +57,12 @@ deutex-check:
 	echo "deutex can be downloaded from https://github.com/Doom-Utils/deutex."; \
 	echo "The full path to duetex can be specified by passing"; \
 	echo "DEUTEX=/the/path/to/deutex to make when building Freedoom."; \
+	exit 1; }
+
+# Make sure that no PNG files are modified if scripts are to modify them.
+pngs-modified-check:
+	@{ ! git status -s | grep -q \\.png$ ; }  || { \
+	echo "PNG fix targets can not be run if there are modified PNGs." ; \
 	exit 1; }
 
 #---------------------------------------------------------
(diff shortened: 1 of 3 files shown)


selliott512:
For those curious I've compared the three WADs produced by the build before and after the two new PNG make targets.

For `fix-deutex-pngs` `freedoom1.png` and `freedoom2.png` are completely unchanged. However, for some reason `freedm.wad` had four PNGs impacted:
```
patches/redtnt2.png
patches/rw18_4.png
patches/w103_1.png
patches/w103_2.png
```
Although they are technically different (`diff` says so) the difference is limited to the `IDAT` chunk (limited to the pixels). All four images look exactly the same to me, before and after, even when looking at them full screen on a desktop, so I don't think it matters.

For `fix-legacy-transparency-pngs` 172 PNGs are changed due to having color 255. Again, I've confirmed that only the `IDAT` chunk is changed. However, for some reason `graphics/endpic.png` is much larger going from 28 KB to 107 KB which accounts for a vast majority of the 100 KB increase in size for all 172 modified PNGs. `pngcrush` completely fixes it (it gets the size back down to 28 KB), but I suspect that the real issue is that `graphics/endpic.png` is something we don't want to modify for the original problem, or is it? Even if that's something that should never be transparent would the `PLAYPAL` and `COLORMAP` lumps be applicable? I'm happy to exclude it, but I'll wait to get some feedback.

selliott512:
One more observation - to further test my theory that color correction chunks do not matter to deutex and Doom engines I removed color chunks `cHRM`, `gAMA`, `bKGD` from three sprites PNGs (`bos2a1.png`, `sposf3f7.png`, `trooe4e6.png`), rebuilt, and it had no effect whatsoever on the three WADs produced.

fragglet (approved):

PR #986, opened 2023-06-19
levels: minor fixes.

mc776:
E1M3
- Nukage level in the bridge before the blue key was  too log, resulting in not only a texture misalignment on the safety lift but also making a secret inaccessible.
- Fixed the tutti-frutti and misaligned terminals near sector 255.

E1M4
- Potential soft-lock if jumping out the window in the starting area. That outside area now has a one-way route back to the map start. (It is still totally inaccessible in vanilla and thus there are no new pickups or monsters there, and the door is "secret" on the map even though it doesn't actually award any secret to open it.)

Map09
- Moved the rocket launcher secret in the southeast to connect with the path leading up to it, preventing the need for an extremely precise jump to get in even when you know where it is.


Xindage (approved):
wanted to make a joke about e1m3 and the logm but better not.

PR #987, opened 2023-06-20
CREDITS: Update GojiBerry, remove merge remnants

selliott512:
GojiBerry should be credited for "sound" not "levels". Also, the "=======" at the end is almost certainly left over from a previous git merge conflict - it has exactly the seven characters expected. I kept it until now thinking it had some meaning to something, but that's unlikely.


Xindage (approved):
My fault, thanks for the corretion.

PR #988, opened 2023-06-24
genmidi: Import DMXOPL instruments.

fragglet:
This is ConSiGno's OPL instrument patch table, found here:
<https://github.com/sneakernets/DMXOPL>

This is the result of many years' development and probably the
highest quality Doom OPL instrument set that's ever been developed. I've
obtained permission from @sneakernets to submit it to Freedoom.


mc776:
This is an *enormous* improvement.

mc776 (approved):

PR #990, opened 2023-06-24
levels: various minor fixes.

mc776:
Thanks to Inuk for playtesting.

Mostly jump/squeezeglide-proofing and aesthetics.

Also a not quite as minor update to Map22, which was originally designed as a standalone challenge map rather than part of a longer moderate-skill campaign - the starting area is now much more generous to narrow the difficulty gap between pistol start and continuing from map21, and just be less frustratingly difficult generally compared to the other maps near it.


Xindage:
before i aprove that massive update it would be good to know what are the changes in the other maps, so i can modify the test roadmap if something needs testing.

mc776:
It's all in the commit messages themselves... i was originally thinking of compiling them in the PR message but it would be a bit unwieldy without any way to add a cut/"read more"

Xindage (approved):

PR #993, opened 2023-06-25
manual: update some sprites.

mc776:


manual/images/monster-technospider.png
fragglet, line 1: Side-on view of the technospider might be more distinctive but just a nitpick

fragglet (approved):

manual/images/monster-technospider.png
mc776: I like the dramatic "moments before disaster" vibe of the head-on initial attack frame, but here are the other two shots I took:

monster-technospider1
monster-technospider0

PR #994, opened 2023-06-25
levels: new e2m1.

mc776:
Starting area by Xindage, cave area and monsters/items by me.


nful2:
This looks EXCELLENT aesthetically.

But, I have no idea how to get these bonuses
Screenshot (32)
Screenshot (33)

mc776:
You have to time a change in direction while running along the top of those fans very carefully.

mc776:
To keep the original weapon progression there should be a secret containing the minigun.

I've now got two versions of the map, one where the incinerator secret also contains a minigun, and another where the minigun secret is much better hidden in the eastern tunnels (and can be made much less well hidden if needed).

Thoughts?

Talon1024:
Playing this level, and then playing the current E2M7 makes an episode 2 playthrough feel better, because it doesn't feel like you're coming back to E2M1 after some construction work any more.

On the other hand, why does the big black secret area behind the fake wall have nothing in it?

mc776:
It's an alternate route, and also a reminder if the player manages to miss that northern switch.

In DM it might be the only way out in the early game, but also a way to ambush people.

In co-op you can use it to telefrag the trilobite.

Talon1024:
Are there really supposed to be this many fans?

Screenshot_Doom_20230628_142250

mc776:
There's supposed to be a "four elements" theme going on - upper areas are air and water, lower earth and fire.

mc776:
And done.

nful2:
@Xindage can I submit screenshots?

nful2:
@Xindage freedoom-0.12.0-517-g0d16f842, gzdoom-4-10-0
Didn't face any issues

e new e2m1
m new e2m1
h new e2m1

Xindage (approved):

PR #1002, opened 2023-07-01
sprites: new life charge and life surge

Talon1024:
I did this mainly because other mods such as Eriguns2 use Amuscaria's Juggernaut sphere sprite for the berserk powerup replacement.


mc776:
Going from actual hand drawn sprites to vague geometric shapes is a massive downgrade.

mc776:
Sorry for being so terse in my initial reply, but my brain really violently NOPE'd when I saw the proposed sprites next to what they were replacing.

They'd be fine as pickups in another project, or a much earlier version of Freedoom, but at this point we'd be inserting into a project that has all these hand-drawn sprites with these sprites that look like low-poly 90s-style renders, which creates a significant art clash that I believe is a big step in the wrong direction.

Thinking about the symbolism of the shapes themselves, after some thought I can see how the dodecahedrons can be an allusion to the D&D barbarian class (the one that tends to get the most hitpoints) and the pyramid the death-and-rebirth imagery of ancient Egyptian funerary rites, but as actual things that the freedoomguy could be picking up to heal or regenerate they seem awfully abstract - which, again, I don't think is a step in the right direction.

This could be the basis for an eventual replacement, though, but it will need more work before it can get my support.

PR #1003, opened 2023-07-01
png: Map color 255 to color 133

selliott512:
Since color index 255 is a special transparent color in some game engines color 255 has been mapped to similar color 133 using make target 'fix-legacy-transparency-pngs'. The images have also been normalized with make target 'fix-deutex-pngs'.

---

This is my attempt at getting mapping fixing the D2TWID cyan pixel issue. For background see [PR #969](https://github.com/freedoom/freedoom/pull/969), where a different approach was taken.

In this PR I did the following:

1. I used make target `fix-deutex-pngs` to normalize all of the PNGs by copying them from the WADs built via deutex.
2. I mapped color 255 to 133 in all PNGs with make target `fix-legacy-transparency-pngs`.
3. I made sure that all 172 PNGs with the color change are in the at least one of the three IWADs (otherwise there is no point).
4. I ran make target `fix-deutex-pngs` again since that has better compression than the output `fix-legacy-transparency-pngs`.
5. I reset to `upstream/master` and then extracted only the 172 PNGs with the color change as they were after step 4.

I compared each of the 172 changed PNGs after `fix-deutex-pngs` and after `fix-legacy-transparency-pngs` to see just the effect of the color change. Here are four exceptionally changed images before and after the color change:

**patches/rw48_1.png** Tan added with an abrupt transition:
rw48_1 before 255 to 133  rw48_1 after 255 to 133

**patches/rw48_2.png** Tan added with an abrupt transition:
rw48_2 before 255 to 133  rw48_2 after 255 to 133

**graphics/endpic.png** Less red around the center star cluster:
endpic before 255 to 133  endpic after 255 to 133

**patches/dosnite.png** Less red around the center star cluster:
dosnite before 255 to 133  dosnite after 255 to 133

In some cases `fix-deutex-pngs` caused the PNGs to look significantly different since the PNGs were not in the desired form (not indexed with the Doom palette), but keep in mind it makes no difference to the WADs produced. The PNGs now more accurately reflect the WADs produced. For example, the Necromancer has acquired block dots around the perimeter:
 
vileh0 before deutex  vileh0 after deutex

My claim is that the above change for the Necromancer will not be visible in the game, but we should clean it up (unless we actually want to have black dots around the perimeter).

Notice that none of the problem PNGs are sprites. Should we limit this to just the 144 sprites that have a color change? Not only did I not notice any significant degradation with the sprites, but the Technospider seems improved - the 255 to 133 color change gets rid of a few seemingly extraneous pink spots.

And finally, I don't know if this is worth doing or not since I still don't know how often a game engine runs in this legacy mode where color 255 is transparent, but if we want to fix it this should do so.


selliott512:
@mc776 thanks for the feedback. I'm happy to change the replacement color to whatever is best. Do you think we need to have different substitute colors for different PNGs? If so I could add a feature to the script that only matches the appropriate PNGs for the the color change. So (for example) to use color 66 for just paths that start with `patches/rw48` and 133 for everything else the `fix-legacy-transparency-pngs` build target would end up running something like:
```
scripts/map-color-index -i patches/rw48 -p lumps/playpal/playpal-base.lmp . 255  66
scripts/map-color-index -e patches/rw48 -p lumps/playpal/playpal-base.lmp . 255 133
```
where `-e` is exclude and `-i` is include.

But if you think 66 is better as a whole then that's easier.

For the Necromancer the black dots should be exactly the same before and after my change, and that's the case based on my looking at it in Chocolate Doom, and my testing where I diffed the IWADs before and after my `fix-deutex-pngs` target (the other cleanup target that makes the dots more visible, not the color change one) - the IWADs are not changed, and least with regard to the Necromancer. You can read about the testing I did [in this merged MR for the scripts](https://github.com/freedoom/freedoom/pull/983).

For the Necromancer the attack frames are PNG color type 6, which means they are RGBA. So not only did deutex have to map each color to the closest Doom palette color with each build, it also had to figure the partial transparency at the border (instead of having a sharp border in some places (not all) it's a gradual aliased transition), which it did with black pixels, or ta least that's my theory. There are some black pixels as it is. The RGBA Necromancer PNGs are:
```
vileh0.png
vilei0.png
vilej0.png
vilek0.png
vilel0.png
vilem0.png
vilen0.png
vileo0.png
vilep0.png
```

As a matter of style the Necromancer always looked more sketch like, like series of colored pencil drawings, when compared to the other monsters. That black dots probably contribute to that.

I'll try to get to this this weekend.

selliott512:
Unfortunately the technospider looks noticably worse with color 66:
bspia3a7
whereas 133 matches almost perfectly.

At this point I'd like to push back suggest that we do 133 for now. Folks can the fix up select images based on PNGs before or after this change.

This [page](http://www.colortools.net/color_matcher.html) can be used to compare colors for similarity percentage. Comparing 66 and 255 is 94%. Comparing 133 and 255 is 96%.

There's a conflict now, so I'll need to go redo it before this can be merged regardless for the color chosen.

mc776:
Given that the biggest example in favour of 66 is from possibly the set of patches most clearly in need of replacement, and I can barely even see the difference in that star cluster, yes let's stick with 133.

selliott512:
Great. I redid it using the same procedure to resolve the conflict. There's 171 PNGs now instead of 172 since, apparently, `sprites/bexpe0.png` was changed to no longer have color 255.

mc776 (approved):

Issue #965, opened 2023-05-06
music: d_dm07 is too big


mc776:
Same problem that d_map30 had, it won't play in vanilla.

Xindage:
i'll tell korp about it.

PR #1006, opened 2023-07-07
New d_dm13.mid music

rostuhan:
Current d_dm13.mid is using the e1m1 music, so I've composed this to add a unique MIDI to the wad.


mc776:
Could we use this for DM07 instead? (see #965)

I think the upbeat tune actually fits that map better than DM13.

rostuhan:
Sure, replacing the glitched soundtrack is more important than replacing a reused audio. Do I have to edit anything or do you edit it?

rostuhan:
I'll close this to make a new pull request that will replace the DM07 music instead.

rostuhan:
(This pull request was moved to #1008)

PR #1009, opened 2023-07-08
New d_dm07 music

rostuhan:
I've originally made [this (#1006)](https://github.com/freedoom/freedoom/pull/1006) pull request, but I didn't know about [this (#965)](https://github.com/freedoom/freedoom/issues/965) glitch, so I've made this fork and the pull request.


mc776:
Just tried building it now and listening in chocolate/opl, sounds good to me.

How should you be credited? (See credits file for format https://github.com/freedoom/freedoom/blob/master/CREDITS)

rostuhan:
I'm not really sure what does each letter mean, does "S" stand for username?

rostuhan:
If I'm right, then it's like this:
```
S: rostuhan
E: rostuhan123@gmail.com
```

mc776:
I'm not sure what S stands for exactly but yeah that's where we're putting people's online handles.

Actually could you add that directly to the credits file as part of the PR? That would keep everything in more or less one place. (I think right now we're just adding people to the end of the file, no special order or anything)

rostuhan:
> Actually could you add that directly to the credits file as part of the PR?
Sure.

rostuhan:
Done, I've edited the branch.

mc776 (approved):

PR #1010, opened 2023-07-08
New d_dm13 music

rostuhan:
d_dm13 is currently using the E1M1 music, so I've decided to compose this.


mc776:
Since this isn't really a bugfix like DM07 I'll let this sit for a while so others can provide some feedback on the new tune.

mc776 (approved):

PR #1011, opened 2023-07-08
dehacked: give phases only their own par times.

mc776:


andOlga:
Just to provide context for why this is necessary:

GZDoom (and possibly other engines?) gets confused when the same dehacked file contains par times for both DOOM 1 and DOOM 2. The GZDoom behavior causes all par times from Phase 1 to be missing, while Phase 2 ones still display correctly.

This PR fixes the par times for Phase 1.

selliott512 (approved):
Looks good. have no idea what the par times should actually be, but that's a different topic.

PR #1018, opened 2023-07-12
music: restore korp's dm07; move rost's new tracks

mc776:
Addresses #965 with an edit of the original music rather than an entirely new track.

Moves rost's recent submissions to two other maps that need new music.


rostuhan:
I'm not sure if it fits that maps, but that's a good idea.

mc776:
I think your Map07 tune fits Map25 even better...

fragglet (approved):

Issue #1015, opened 2023-07-11
Freedoom wads cannot be built with python-pillow 10.x.


FredBezies:
I tried to build revision 15b03eb3, but it does not work. It stops with this error log:

```
make[1]: Leaving directory '/home/fred/freedoom-git/src/freedoom/lumps/dehacked'
make VERSION=v0.12.0-512-g15b03eb3 -C graphics/text
echo v0.12.0-512-g15b03eb3 > lumps/freedoom.lmp
make[1]: Entering directory '/home/fred/freedoom-git/src/freedoom/graphics/text'
./textgen
./create_caption ../titlepic/freedm_titlepic.png fdmtitle.png
echo v0.12.0-512-g15b03eb3 > lumps/freedm.lmp
Traceback (most recent call last):
  File "/home/fred/freedoom-git/src/freedoom/graphics/text/./create_caption", line 21, in <module>
    txt1_size = draw.textsize(txt1, font=font)
                ^^^^^^^^^^^^^
AttributeError: 'ImageDraw' object has no attribute 'textsize'
make[1]: *** [Makefile:157: fdmtitle.png] Error 1
make[1]: *** Waiting for unfinished jobs....
```

Any idea?

mc776:
Works fine for me and I have no idea how to unpack that.

What's the last commit that builds successfully?

FredBezies:
Last time I build it was on June 6, 2023. Commit: 4b002550

mc776:
The first commit after that one is just some updates to maps... I'm not sure the build system even looks into those files at all.

Did you try any of the commits in between then and now?

mc776:
Can you try this commit first? https://github.com/freedoom/freedoom/tree/e88e10208bfe290ee475d3a2c74dd3e17b7052bc

Immediately after this commit is a change to the build process.

FredBezies:
> Can you try this commit first? https://github.com/freedoom/freedoom/tree/e88e10208bfe290ee475d3a2c74dd3e17b7052bc
> 
> Immediately after this commit is a change to the build process.

Broken. Here is the output:

```

cp wivctms.png ../
./create_caption ../titlepic/freedm_titlepic.png fdmtitle.png
./rotate wikilrs_horiz.png 270 wikilrs.png
Traceback (most recent call last):
  File "/home/fred/freedoom-git/src/freedoom/graphics/text/./create_caption", line 21, in <module>
    txt1_size = draw.textsize(txt1, font=font)
[...]

PR #1027, opened 2023-07-16
create_caption: Python PIL 10.0.0 support

selliott512:
To support Python PIL 10.0.0 this change uses newer API textbbox() when available, and older API textsize() when not.

---

I've tested this using a static version in `Makefile` (the version otherwise causes the PNGs produces to be different):
```
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
-VERSION=$(shell git describe --abbrev=8 --dirty 2>/dev/null || echo v0.12.1)
+VERSION=static-version
```
and found that it produces exactly the same `graphics/fd*title.png` PNGs for PIL 7.2.0 and 10.0.0 with this change, which again match the PNGs produces for 7.2.0 without this change.

### DIFF
--- a/graphics/text/create_caption
+++ b/graphics/text/create_caption
@@ -18,8 +18,21 @@ background_image.load()
 background_image = background_image.convert("RGBA")
 image = Image.new("RGBA", background_image.size, (0, 0, 0, 0))
 draw = ImageDraw.Draw(image)
-txt1_size = draw.textsize(txt1, font=font)
-txt2_size = draw.textsize(txt2, font=font)
+
+# Getting the text size is tricky since for newer PIL, such as 10.0.0, only
+# textbbox() is supported, but for older PIL, such 7.2.0, only textsize()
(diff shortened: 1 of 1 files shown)


selliott512:
Thanks to FredBezies for finding this, doing the preliminary research, and opening an [issue](https://github.com/freedoom/freedoom/issues/1015).

selliott512:
Judging by the continuous integration failure I should catch all exceptions, and then fallback to the old API. I'll fix it later today.

It seems there are at least three different PIL behaviors as a function of its version, which is unfortunate, but I'll make it work.

selliott512:
This seems to be better.

mc776:
Is this ready to merge?

selliott512:
> Is this ready to merge?

Yes, ready as far as I know.

mc776 (approved):

PR #1032, opened 2023-07-18
patches: add two more screens to COMPUTE4.

mc776:
Some maps probably still have some COMPUTE4 textures that use the rightmost column of screens to the left of the leftmost column, which the recent expansion to 256 pixels would have broken. This adds two brand new screens to fill up the space, using colour schemes similar to the spectral graph and blue nebula.

I would like to add these to PLANET1 as well but there are probably maps that accidentally use the previously existing blank space to the right that would be broken by such an addition. If it is known for sure that I am wrong about this, let me know and I will add those as well.


Xindage:
Compute4 is a relative new texture, changing it now would not break things too hard, and  would recommend doing it sooner than late.

Xindage (approved):

Issue #1033, opened 2023-07-21
New LOGO textures


superdave938:
I redid the logo textures
![Screenshot_20230721_155012](https://github.com/freedoom/freedoom/assets/71569097/667fd95e-3fe8-4cc9-a442-a6e3ffbcd843)
![Screenshot_20230721_160936](https://github.com/freedoom/freedoom/assets/71569097/de827339-0f61-4203-9330-b335bc7a5501)
[fdmnewpatch.zip](https://github.com/freedoom/freedoom/files/12134721/fdmnewpatch.zip)

mc776:
I can turn this into a PR but how should you be credited in the [credits file](https://github.com/freedoom/freedoom/blob/master/CREDITS)?

superdave938:
S: SuperDave938
D: textures

PR #1034, opened 2023-07-22
patches: implement #1033. 

mc776:


Xindage (changes requested):
logo5 alt is not present in textures file and it'll be killed during build process, pleas fix it.

Xindage (approved):

PR #1037, opened 2023-07-23
levels: Small fixes, mostly for Eureka

selliott512:
```
freedm.wad
  DM22  Made invulnerability sphere spawnable
        Removed doom camera (thing #32000)
        Removed unused tags far five sectors
  DM32  Change map name from MAP22 to MAP32

freedoom1.wad
  E1M7  Removed shells (thing #704) in the void

freedoom2.wad
  MAP09 Fix missing texture behind teleport
  MAP10 Rebuild nodes
  MAP11 For texture BLOD3 replaced CYAN with PBLAK to avoid
        transparency on solid
```

---

The most controversial change is probably what I did for MAP11, which is not actually a level change. It fixes a Eureka error that complains about a transparent texture (BLOD3) being used on a solid surface by making the texture not transparent, and that's done by replacing CYAN (the apparent transparency color) with PBLAK (black).

It's a recessed skull painting in the room that's to the West of the exit area.

Do we need to have transparency at the end of BLOD3 even though Freedoom levels do not use it? Googling "BLOD3" this change may impact TNT switches, but I'm not sure.

I can fix this a different way - I can create a texture that consists of DOSW3C, and nothing else, and use that.

Or we can ignore this Eureka error, but that seems unfortunate since we don't have any Eureka errors of this type, so it would muddy the waters with regard to which Eureka errors are real.


mc776:
I was thinking of changing that texture back to the skull switch anyway - the only reason I used that texture in the recent Map11 update is because I thought I might be submitting a SW1SKULL replacement that made it not a skull anymore.

I've got a couple minor level fixes/tweaks I'm working on (no overlap with this PR) and once #1036 is merged I'll replace the texture on Map11 and submit it.

Best to leave the texture alone.

mc776 (changes requested):
don't change the textures lump, i'll fix Map11 separately

selliott512:
@mc776 sounds good. I've removed the texture change, and the references to MAP11 in the commit message. It's just other level changes now.

mc776 (approved):

PR #1038, opened 2023-07-23
levels: minor tweaks/fixes.

mc776:
E2M1
- The little pass-over route in the north with the medikit now blocks monsters so you don't get a surprise invisible obstruction.
- The armour bonuses on the steps in the main path between the outdoor areas are now closer to the ledge. You can still miss some if you charge down the steps at full speed but it only takes a slight slowdown to get them all in one go.

Map11
- The health bonus in the starting room pillar is replaced in hard mode by a shellbox.
- Untagged the thin panel armour bonus secret.
- Replaced the evil eyes in the SKAG room with skull rocks.
- Made all 3 doors in the starting room green.
- Fixed texture alignment on castle corpse hanger lift.

Map09
- Flagged some secret area entrance lines as secret.


mc776:
This can be merged anytime absent any changes anyone else wants - I'm just kinda throwing stuff in here as I notice it.

Xindage (approved):

PR #1040, opened 2023-07-27
levels: more minor fixes.

mc776:
E1M2
Fixed tutti frutti around exit door.

Also took the liberty of fixing a longtime pet peeve I had about the really bright grey-white of the exit room steps clashing with the rest of the room for no apparent reason. The tan softens the lighting contrast slightly without making it less visible, while the lights around the base of the higher platform help explain why the lower one is still so much brighter than its immediate surroundings.

Map22
The candle marking the secret would snap to the top of the lowering panel. Moved it very slightly away from the line.


Xindage (commented):
just confirm for me that we have no testing player spawn present.

mc776:
Confirmed.

Xindage (approved):

Issue #1031, opened 2023-07-18
New Punch sprites


superdave938:
![ezgif-5-f92a4d8a77](https://github.com/freedoom/freedoom/assets/71569097/0bdbebed-9bbf-4b48-a5de-5873c927a197)
[Fdnewpunch.zip](https://github.com/freedoom/freedoom/files/12074816/Fdnewpunch.zip)
I made new punch sprites

mc776:
I just saw this issue after I made this comment seeing the ingame video on Discord:

> The animation looks as good as the doom fist will get, though the sprite itself is going to need work to look more like other Freedoom assets and less clearly like it was cut out from photos... I've actually commissioned a friend to try to touch up another fist animation I tried to make and depending on how that turns out I might be able to get her to give this set a try too

But that said, we could commit this now and any additional improvements to the sprite can be done later...

mc776:
(adding the "bug" label since this would also fix the widescreen problem the current fist sprite has)

Xindage:
que quality seems low, maybe it still need some touch up.

mc776:
Just gave it an actual try ingame.

Here's a small edit that limits the left hand to be visible only on the A frame, and makes it less visible (and uses the hand position of the B frame). The reason for this edit is that when you hold down fire and punch continuously you never see the A frame, so with only one frame it just looks like the left hand is constantly blinking in and out of existence. The A frame fist is lower just so the transition isn't as abrupt.

I've also tweaked some of the offsets so there's a bit more of a contrast to imply some speed.
[fdpunch-edited.zip](https://github.com/freedoom/freedoom/files/12101536/fdpunch-edited.zip)

Thoughts?

mc776:
> The reason for this edit is that when you hold down fire and punch continuously you never see the A frame, so with only one frame it just looks like the left hand is constantly blinking in and out of existence with no apparent motion.

I just noticed that this happens with the current sprite, it's just only visible if you disable the status bar.

Anyway, here's an attempt to clean up some of the photo artifact edges, scale down the height of the images to compensate for pixel scaling and add some highlights to make it look a little more like skin again. I think it's as good as the current sprite at this point - if no one points out any major glaring errors I may post a PR soon.
[pung.zip](https://github.com/freedoom/freedoom/files/12208534/pung.zip)

mc776:
I'd have posted a PR by now, but with #1041 still outstanding I'm afraid of accidentally overwriting the offsets from that or vice versa...

Issue #760, opened 2021-09-05
Fist cutting off in ultra-wide ports


colinmahns:
Using the fists on my ultra-wide display (21:9) causes a cut-off on the fist's graphic assets when playing the WAD in gzdoom. 

Comparing to the source material, Doom handles this new aspect ratio change and stretch much better as the fist is much more centered, coming from around the weapons/health is on the HUD. Freedoom's on the other hand as it's coming from the far right of the screen, isn't able to handle this shift as cleanly. 

![Screenshot from 2021-09-04 21-26-15](https://user-images.githubusercontent.com/1781677/132114339-d8c813c0-50fb-439d-8552-ba0a2e81d554.png)

I opened a new issue as I didn't see any existing issues for this, but this is similar to #95.

mc776:
The only sure way to fix this is with sprites that only cut off at the bottom (similar to what we see in Doom).

There was some discussion about new fist sprites on Discord a few months ago. Did anything come of it?

colinmahns:
I just did a quick search in the Discord, is this what you're referencing @mc776 ? https://discord.com/channels/616148111076884520/692858923660607518/840758020933812264

mc776:
Yes, that was the discussion I had in mind... and it doesn't look like it did get anywhere beyond some bikeshedding about wrist alignment.

P41R47:
now that we are at it.
We can also take of those gloves from it.
The other hands sprites doesn't show any kind of gloves, so, to differentiate again more to the original assets, we should get rid of them here, too.

chungy:
Bottom of the screen is probably the best idea.

mc776:
Fist has brown gloves.
Pistol has no gloves.
Shotgun has no gloves.
SSG has brown gloves.

All other weapons do not show hands.

So right now it's an even split and we could go either way.

Personally I'm in favour of moving towards all gloves because that way people can have custom mugshots and characters with different skin tones and it wouldn't clash or require a recolour.

PR #1046, opened 2023-08-01
sprites: new fist.

mc776:
With modifications, see https://github.com/freedoom/freedoom/issues/1031.

Addresses #760.


Xindage:
Do these really looks better? for me it seems that the hand is made of plastic.

mc776:
It's better than the current sprite at least, which has all the same problems except the proportions are a bit off and feels like it's coming out of the player's right ear - it's only less noticeable because we're more used to seeing it.

selliott512 (commented):
I like it. It has a cleaner (Less hairy? Less gloved?) look.

However, the four modified PNGs are color type 6 (RGBA). Ideally they should be color type 3 (indexed, and ideally with the Doom palette).

selliott512 (approved):

fragglet:
Just a quick note - I missed the original PR but I wanted to say that this new fist is **awesome**. Probably the most satisfying punch animation I've seen in any Doom mod.

PR #1048, opened 2023-08-03
dehacked: remove armour blinking.

mc776:
The single most annoying creative constraint in creating an armour sprite is to explain why it's a blinking light. It never made sense in id and the explanation looks ridiculous in the new sprites in the recent devbuilds. (I say this as the person who added them - it's still in my view preferable to the alternative pure arbitrariness.)

If we get rid of the fullbright, we can design armour like any other mundane tech/ammo/weapon pickup.

This should be merged only after the current armour sprites have been replaced, as the current ones are clearly supposed to be fullbright in their B frames.

### DIFF
--- a/lumps/dehacked/dehacked.txt
+++ b/lumps/dehacked/dehacked.txt
@@ -27,6 +27,14 @@ Sprite subnumber = 32773
 Frame 689
 Sprite subnumber = 32773
 
+# The armour fullbright blinking never made sense.
+
+Frame 803
+Sprite subnumber = 1
+
+Frame 805
+Sprite subnumber = 1
+
 # The super shotgun's first frame is 7 tics long before it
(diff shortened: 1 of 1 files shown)


selliott512 (approved):
Looks good. I know you don't intend for this to be applied by itself, but it still seems like an improvement to me, at least in GZDoom.

mc776:
I navigated to this page just now for the express purpose of posting "actually I just tried Cascade's ARM*A0 on both A and B frames with no further dehacked and it looks fine" and closing this only to find it has been approved, lol.

Let's let this sit for a bit in case anyone else has any thoughts...

mc776:
Having tried the new sprites now, it seems the potential adverse gameplay effect overrides the aesthetic effect.

Also I don't feel like changing the radsuit as well.

PR #1049, opened 2023-08-03
patches: new RW33 series by Korp. [light version]

mc776:
I'm posting this one as a separate pull request from all the other patches @KorpKat has been working on.\* I've already posted about this on Discord but I definitely don't want feedback restricted to only people on there.

There's a dark version of these patches that was included as well, but it suffered from severe palette restrictions so where the current is a nice rough textured surface the new one would have these big splotches of pure solid colour.

This version solves that problem entirely, but is noticeably lighter than the current version (or its id equivalent).

I think it's a worthwhile tradeoff to get the details, but others may quite reasonably disagree.

Thoughts? Should I post a separate PR with the dark version so people can compare?

\*(For the others, see #948 for the gray/ick textures; all the other ones I think should be merged immediately (or at least before the next release) and will be put into a separate PR.)


mc776:
Some comparison shots (dark = new dark not current):
doom11
doom10

doom06
doom02

doom07
doom05

Xindage:
i dont think its a good idea right now, and changing the tone of this texture may cause some visual issues, i can point it out on some of freedm maps.

mc776:
People's thoughts on the dark version? I've been trying for a while now to see if I could noise up those big dark patches but it just isn't working, but looking at it again I'm not 100% convinced it's as bad as I initially thought, but I could be wrong now and right the first time...

KorpKat:
I would rather the darker variant gets added.

PR #1050, opened 2023-08-03
patches/flats: new textures by Korp.

mc776:
Replaces:
- all hex patterns
- all STARG\*
- the big tech doors
- some of the narrow doors


selliott512:
I don't have strong feelings about this, but I do like the doors, and I do tend to favor cleaner and lighter aesthetics in games. 

Also, the file format of the PNGs looks good. They seems to be the correct indexed mode. The palettes (PLTE chunks) seem to be the Doom palette, just truncated, so that's good.

mc776:
Can you approve it?

selliott512 (approved):

selliott512:
Approved. Should I just merge it, or is more discussion needed?

Not that it matters with regard to getting this merged, but a bit more about my PNG commit for PNG nerds - pngcrush tends to truncate PLTE at the last color used. It doesn't matter except that it can make it harder for people editing the image to add colors that are at the end of the Doom palette, but they can probably figure it out. FWIW I'm pretty happy with `make fix-deutex-pngs`. The result only has slightly worse compression than pngcrush, but things like the mode, palette and chunk types are reasonable.

PR #1051, opened 2023-08-04
manual: minor fixes.

mc776:
- Edited out the references to "Cannon".
- Clarified that F11 isn't an on/off toggle. (The Spanish seemed fine?)
- Reworded the missile description to do damage to imply it works better on *lower*-level monsters rather than higher, since the tripod and bigtech are both immune to the blast.
-  Created new illustration of the savegame menu to replace that "C" with an "E".


selliott512 (approved):
Looks good. I can't comment on the Spanish other than pasting a few of the changes into Google Translate produced reasonable results.

PR #1057, opened 2023-08-06
patches: new BIGBRIKs by Korp.

mc776:


Xindage:
no for now, i heave already applied changes on my maps that used that textures to fit with the current ones, i dont want do it again right now.

MerrittK:
Could it be added as a new set of textures? As a mapper resource and not a replacement.

It looks like a good texture ,but the metallic rivets change the context of the texture and make it inappropriate for it's current usage.

PR #1059, opened 2023-08-07
sprites: add Cascade's pickups (etc.).

mc776:
Incidental upgrades to a couple other sprites to better match this higher-contrast style.

Took the opportunity to fix the scaling on the missile projectile and give it longer, simpler-looking fins.

Note that the armour A and B sprites are identical and #1048 may be safely implemented if people want.


nful2:
Some suggestions:
1. Add the small helmet at the bottom right side in the new armor and megarmor
2. Add 'missiles' text on the new missile box pickup.
3. Add the top portion of missile poking through the missile box.
4. Large energy cell pack replacement looks worse than the original. Though I think the triangular caution symbol can be added somewhere in the original?

mc776:
No comment on 1, there actually is a "missiles" text on the lid, the additional redundant missiling makes the sprite unreadably busy, and disagree.

mc776:
Looking at the old sprite again and comparing it with the new I'm interpreting the right panel as readings for the status of the batteries being loaded into it.

How's this?
image

nful2:
Right side looks good! But the left side looks like 2 humans for me. Grey part Is the skin with a hand and the blue part looks like a cap and dress  XDD
Maybe place 2 small cell recharge stacked up, indicating it will give more plasma

mc776:
> Maybe place 2 small cell recharge stacked up, indicating it will give more plasma

That's actually what I interpreted that as, but now that I actually look at the two sprites together... well let's try making it that way:

image

nful2:
Perfect!

mc776:
Apparently Cascade hates it and is working on their own upgrade... personally I think this version communicates the idea of cell recharge perfectly, but I haven't seen theirs yet, so in the absence of more people giving their input you may end up with a casting vote...

nful2:
More suggestions - add red or green background on the left side. And like the old Sprite, add a rounded left side. Also, like the old sprite, make it a bit greyish seems too black to me

Xindage (approved):

mc776:
The "red background" never made any sense, it seemed to want to be lighting and suspension ooze at the same time while accomplishing neither - I honestly had no idea what that entire left side even was until the interpretation offered by these new sprites.

The "big car battery" symbolism of this new outline would be ruined by trying to make it more like the old sprite's shape, which feels more like a computer thing? (It might be an idea to refurbish that old sprite into a new automap sprite though...)

As for being too dark, I note you didn't have this issue with the first version of the dial, so I'm assuming you were fine with that... how is this in principle?
image

Issue #163, opened 2015-08-22
Generate translation file for Desktop and AppData files


GreenLunar:
and add a note, video game is available only in English.

jewellds:
That's probably because the English text isn't done yet!
In my opinion, it doesn't make sense to have the game in other languages while people are still working out what the level names, monsters and story are in English.
I reckon when the English text is in a more complete and stable, that would be the best time to translate the game into other languages.

GreenLunar:
I was referring to files in https://github.com/freedoom/freedoom/tree/master/dist

GreenLunar:
I am willing to do anything that needs to be done to complete the English text.

PR #1060, opened 2023-08-09
levels: Small missing texture fix for MAP26

selliott512:
freedoom2.wad
  MAP26 Fix missing texture on the back of linedef #163


mc776:
It's also incorrectly marked "not on map", could you fix that too?

selliott512:
> It's also incorrectly marked "not on map", could you fix that too? (I'd do it myself but I'm on the computer with the weird node builder)

I want to make sure I understand. Is "not on map" a way of referring to the visibility? Here's what I can set for the visibility for that linedef in the Eureka drop down, and the corresponding UDMF variables (which have the same names as the Doom flags):
```
Eureka     UDMF
------     ----
Normal
Invisible  dontdraw=true
Mapped     mapped=true
Secret     secret=true
```
Right now that linedef is "secret", but the other linedefs that make up the sectors above and below it are all "invisible". Do you want "invisible"?

selliott512:
Ok, digging deeper SLADE's "Not On Map" implies dontdraw=true. SLADE allows that boolean to be sent independently whereas Eureka thinks it's irrelevant if "Secret" is selected, or something like that. I'm still not quite sure what to do though. Maybe we can leave it be for now.

selliott512:
It would be helpful if these level editors just directly revealed the underlying Doom flags instead of trying to interpret them.

mc776 (approved):

PR #1062, opened 2023-08-10
patches: implement new AQCOMP01 properly.

mc776:
Fixes error introduced in #1039.


selliott512 (approved):
Looks good. However, the PNG is now color type 2 (RGB) instead of color type 3 (indexed).

PR #1063, opened 2023-08-11
levels: more minor fixes.

mc776:
Currently the map merges the METAL2 plates, like this:
doom01
which looks meh with these textures, but with something that's less uniformly noisy it's obviously bad:
doom00
so here's another light instead:
doom02


mc776:
Here's the Map03 fix:
DOOM00
DOOM01

mc776:
[map19.zip](https://github.com/freedoom/freedoom/files/12373821/map19.zip)

Here's one possible alternative. It's still a relatively easy secret to find, but getting there without taking any damage might be another issue.

This un-solves the problem of the pinkies blocking you from below though...

Xindage (approved):

PR #1064, opened 2023-08-11
sprites: new evil eye.

mc776:
Based on the current "skull rock".


selliott512 (approved):
Nice. This definitely seems closer to the idea concept of an "evil eye", which should be fair game.

A few observations that may or may not be an issue:

- The three previous PNGs were all 48x60, wheras the new ones have some variation (42 and 43 for the height):
identify `find . -iname ceye*.png`
sprites/ceyea0.png PNG 39x43 39x43+0+0 8-bit sRGB 1427B 0.000u 0:00.000
sprites/ceyeb0.png PNG 39x42 39x42+0+0 8-bit sRGB 1495B 0.000u 0:00.000
sprites/ceyec0.png PNG 39x43 39x43+0+0 8-bit sRGB 1505B 0.000u 0:00.000
Is that variation ok?
- Previously the offset was 24x58, which was inside the image whereas now it's 19x47, which is just below the center of the image. Is that ok?

mc776:
I kinda *wish* that the different image sizes would cause a bit of jitter but it doesn't seem to do so.

Right now the eye is basically at eye level in vanilla:
DOOM01

compared to id:
DOOM02

so absent some really specific use case I think it's good to go.

PR #1067, opened 2023-08-12
sprites: new skull rock.

mc776:
Also editing the BFG room in Map11 to use evil eyes again because neither this nor any id-like skull rock could have possibly matched anything in that room.


selliott512:
Here's what the BFG / SKAG room looks like with this PR, and the new evil eye from PR #1064 . I like it:
map11-new-evil-eye-small

selliott512 (approved):

selliott512:
The proposed flying skull is significantly wider. Are there comparability concerns with that?

mc776:
The Map11 SKAG room was the one place I could find where it was an issue.

It's 45 pixels wide, which is the same as the previous skull rock (the more id-like one before the current), but all 3 sprites are wider than the id which is only 32.

I might adjust it one last time before merge.

PR #1068, opened 2023-08-12
patches: new RW33 series by Korp. [dark version]

mc776:
See discussion in #1049.

If RW33_5 really is extraneous we can still get rid of it; otherwise it is now METAL9.

I've also taken the liberty of moving one of my recent textures.cfg additions to their proper place in the "Custom textures using resources of freedoom" section.


selliott512 (approved):
Looks good. I would describe the changes as brighter, cleaner, and with easier to distinguish features, which I tend to like in video games. As and added bonus all of the PNGs are indexed.

Issue #1070, opened 2023-08-13
100% secrets not possible on MAP19


Talon1024:
It is not possible to get 100% secrets on MAP19, unless you turn on noclip. One of the secrets is behind the newly-added guard rails.

selliott512:
Good catch. Are there other cases where barriers have recently been added to levels that could make either 100% kills or secrets impossible? I can try a quick run through of recently modified levels, but I thought I'd ask if anything comes to mind.

mc776:
This is just a case of editing an area where there was a secret and new sectors inherited the secret flag... we'd cleared a whole bunch of them before but this got re-introduced with the relatively recent edit.

mc776:
[map19.zip](https://github.com/freedoom/freedoom/files/12373821/map19.zip)

Here's one possible alternative. It's still a relatively easy secret to find, but getting there without taking any damage might be another issue.

This un-solves the problem of the pinkies blocking you from below though...

Issue #1053, opened 2023-08-05
MAP31 route to MAP32 is pure RNG


GrayFace:
MAP31 involves facing a Cyber Daemon and walking around it in a tiny room to reach a teleporter. Over 95% of the time it ends up blocking your path and then you're dead, at least with fast monsters. This is a level design bug.

mc776:
Which specific room is the issue? And what mods are you loading?

GrayFace:
The daemon in question is x550 y0 z0. No gameplay mods: `"./gzdoom.exe" -iwad "../freedoom2.wad" -file "../midi/menu" "../bonus/HDskiesD1.pk3" +sv_cheats 1 -fast`
(I'm normally playing with mods, but for this level I wanted to experience it the original way.

mc776:
Oh, the supersecret level entrance. Yeah, yikes.

I don't know what to do with it that doesn't make it less... "secrety"...

mc776:
Here's one attempt: create an optional shootable switch to let the tripod leave that area.
[map31.wad.zip](https://github.com/freedoom/freedoom/files/12267684/map31.wad.zip)

It's still a big test of one's patience to hit the switch instead of the tripod but at least you're not actively trying to move into its space at the same time.

mc776:
Moving that shootable switch a bit forward to make it easier to hit:
[map31.wad.zip](https://github.com/freedoom/freedoom/files/12319206/map31.wad.zip)

This also gets rid of orthogonal lines with tentacle textures so you don't see those stripes of fake contrast cutting through the organic shapes.

mc776:
This version moves some lines around to mitigate some texture alignment issues:
[map31.wad.zip](https://github.com/freedoom/freedoom/files/12341642/map31.wad.zip)

Absent any objections I'll add this to one of the outstanding minor level fix PRs.

mc776:
[map31.wad.zip](https://github.com/freedoom/freedoom/files/12346863/map31.wad.zip)
i hate secret sectors

GrayFace:
The fix I was imagining is just placing the teleporter on a side of Tripod's box instead of on the back, so that you still move to Tripod's space quickly, but then just strafe to the left or right.

GrayFace:
With this fix how do you kill the Tripod after releasing him? There's only enough ammo for 1 final Tripod.

mc776:
by not pistol starting i guess.

[...]

Issue #379, opened 2017-02-19
C2M1's difficulty seems wildly out of alignment


fluffymormegil:
The difficulty of C2M1's opening sections feels inappropriate for the first map of an episode, unless it is an episode directly aimed at people who think PLUTONIA.WAD MAP01 was too gentle.

Voros2:
Its nowhere near Plutonia's level of difficulty if you ask me.

chungy:
It is still fairly tough, giving the player spectres with only a pistol at the beginning. We ought to rearrange the episode.

Voros2:
Replacing the spectres with, say, imps or zombies should do it then, since he only says the opening section is difficult.

fluffymormegil:
It's not so much the spectres - I don't particularly care about getting 100% kills, and there's enough room to dance around them - as the blind drop into the field of fire of an army of hitscanners while armed with only the pistol.

Voros2:
I never found myself being confronted by an "army" of hitscanners. Upon jumping down, I kill the few hitscanners present. These hitscanners tend to be far away from the player, so not much damage taken. Plus, then you can pick up a shotgun from the dead shotgun zombie. Upon firing, other hitscanners become alerted and come in one-by-one, but never as a whole. So basically, I'm fighting waves of only a few zombies for a short amount of time. Using a shotgun.

Proluiz:
actualy this map is not that hard, for the skill called hard its very confortable lot of monsters and they come one by one so its easy to dispatch them, in medium the map is actaaly very easy to my flavors, just 20% less monsters more health and early acess to guns, Easy is a "piece of cake", an initial shotgun acess and chaingun just in left side of poll plus near -40% of enemys make it extreme fine fow who "dont know how to play" so i see no poblems at dificult balance.

Since from when i begin mapping i used a logic of someone tell me (i dont remenber who told anyway)

Easy, is for very begginers, player not familiar whit the game at most, extra health and ammo too and low tiers enemys.
Medium, usual players give them challenges whit proper equipment.
Hard, Expert gameplays, hard challenges+traps with bigger amounts of enemys, extra ammo for compensate the difficult and high tiers of enemys present.

Voros2:
I admit this map is bit difficult at times, and could use some nerfing, such as lowering the number of shotgun zombies.

ReakutheCrate:
[...]

PR #1071, opened 2023-08-13
levels: address #1070.

mc776:


mc776:
I somehow misremembered and thought #1063 had already been merged when I posted this.

For future reference, would it be preferable if I kept trying to put all these into larger "minor fixes" updates? (And if anyone feels strongly enough in the positive that they'd rather I closed this and added to #1063 instead, please let me know)

selliott512:
It may be obvious, but with this PR it's possible for the player to walk to secret sector #1, which does not seem hidden. Should it be trickier to get there, or not secret?

mc776:
Frankly I don't really get why the original was a secret - it was made at a time when pretty much everyone could be assumed to be able to SR50, so the only real difficulty is being distracted and approaching at the wrong angle, or being arbitrarily blocked by one of the pinkies below and even that latter one is usually irrelevant if we're not in vanilla.

(If the point of it being a secret/challenge was in fact the possibility of being blocked, we're left with the same problem as #1053 and the thing should be fixed - which is why I've got all these railings and the bridge. Arguably you can learn to kite all the pinkies away and clear a path, but the whole thing becomes an exercise in arbitrary engine limitation exploits that ultimately feel more buggy than any vanilla rendering trick any current map uses.)

It would not be incorrect in my view to just get rid of any notion of that area having a secret altogether, but I'd rather there be something there rather than nothing, hence the latest commit.

mc776:
So, having typed all that out, one other possibility to consider:

- Make that area on the north side much higher and the jump impossible.
- Add a secret route in the nukage somewhere that leads to a lift or teleporter.
- Remove the railings again? (Not sure about this one since it leads us back to the invisible jump block problem)

mc776:
Or possibly:
- *long* elevator from the nukage to the northern platform, which must be activated from below, guaranteeing you'll take significant damage if you just wait for it.
- secret radsuit somewhere.

mc776:
[map19.zip](https://github.com/freedoom/freedoom/files/12373821/map19.zip)

Here's one possible alternative. It's still a relatively easy secret to find, but getting there without taking any damage might be another issue.

This un-solves the problem of the pinkies blocking you from below though...

(sorry for the spam, I completely forgot where this discussion had even been taking place)

mc776:
[map19.zip](https://github.com/freedoom/freedoom/files/12374066/map19.zip)
One option is to just move all the worms out of the trench.

If anyone objects that it's too easy to avoid damage if you fall in, the teleporters can be replaced by an inconveniently placed lift...

selliott512:
Looks good. I like the new secret. I just blasted through it with IDFA and was able able to get 100% secrets and kills.

One small thing - move thing #379, the stealth worm, to to the left a bit. Eureka complains it's a "stuck actor", but I don't know that that is actually the case (at least not in GZDoom).

selliott512 (approved):

Talon1024:
[Here's my idea](https://github.com/freedoom/freedoom/files/12384685/map19_20230818.zip). I removed the bridge and guard rails on one side. Now, there is a thin walkway leading to a small opening on the east end of the other side. I also moved the teleport pad and a pillar to make it accessible.

mc776:
> [Here's my idea](https://github.com/freedoom/freedoom/files/12384685/map19_20230818.zip). I removed the bridge and guard rails on one side. Now, there is a thin walkway leading to a small opening on the east end of the other side. I also moved the teleport pad and a pillar to make it accessible.

I like this! The placement relative to the teleporter also minimizes the chance of a collision with a pinkie in the trench.

That said:
1. Aesthetically I regret putting any railing there at all. Then again, I can see some people rather having that than the shorter chimneys relative to the higher platform.
2. It turns out the initial addition of the railing eliminates a shortcut that lets you skip the entire western area. The high-platform fix restores this functionality.

Could it be preferable to have that thin catwalk lead to a lift instead?

mc776:
I've posted both maps on the Discord in hopes of feedback about which solution people like better.

mc776:
No one's stated any strong preference for railing+catwalk, Steven's given the go-ahead for raised-platform and I'm of the view that the railings were a mistake and a great disservice to the complexity of this map with that quasi-secret shortcut, so I'm merging this now.

PR #1075, opened 2023-08-23
music: reuse picklehammer's old d_map22 for E3M1.

mc776:
see https://github.com/freedoom/freedoom/blob/d74b3024d778109c20309c18963943dfd8290773/musics/d_map22.mid

Same dark gloomy-but-not-passively-gloomy vibe but fits the action a bit better and translates a bit better into OPL.


KorpKat:
I had discussed this on the discord server aswell, but this piece is just not fit for an opener. Its way more monotone and slow than the piece you're trying to replace.

mc776:
With a proper soundfont so you're blasted with the chorals and drums it's a perfect mood piece for suddenly finding yourself in a hostile dark world.

This is less so on OPL, but the way it fails to translate into OPL is *much* less obnoxious than the way the current track fails to translate into OPL.

Besides, you yourself were the one who objected when I suggested something like the current D_E3M5, on the grounds that not every opener piece should be so bright and aggressive.

KorpKat:
I would prefer if we dont prioritize OPL over wavetable. From what i know its not a vanilla limitation and it generally tends to sound bad regardless, not to mention not a lot of people use OPL so its quite unnecessary to fit music to sound good on it.

And the issue isnt that its slow, its that its more boring and obnoxious, it feels rather silly and quirky instead of the current track, which feels menacing and abrasive.

mc776:
AxelMoon rebutted everything you said just there sufficiently on Discord.

And we *are* aiming for vanilla compat, in case you missed the memo.

mc776:
I'll repost what I said on Discord in response to Melodica's valiant but doomed effort to fix the main problem I have with the current D_E3M1 as a whole:

> but listening to this again even best case scenario i still honestly think it's a worse fit for this map than the old map22 track, edited or not
> 
> multiple people have already said they were okay with the replacement in principle and a couple others have positively wanted to see this track returned

selliott512 (approved):
Sounds good to me.

PR #1080, opened 2023-08-25
music: use tremolo strings for d_map19 not piano.

mc776:
When I first started up Map19 with music for the first time in ages I thought the MIDI had glitched somehow and the instrumentation fell back on "Acoustic Grand" the way a lot of things did when I was younger. When I checked the MIDI it still sounded like it was supposed to be something else and the composer forgot to set the instrument, or it's a *really* experimental thing that frankly doesn't really fit a game - or at least not something that is not itself an extremely experimental abstract concept level which Map19 is not.

After some experimentation tremolo strings stood out by a fairly large margin as the best option.


KorpKat:
Ehh, i feel its more of a downgrade. The piano was more fit for this "tense" type of composition, the strings having no release make them sound very insistent, and the arrangement already has strings so the melody fades in the background, the piano has a stronger attack and more force, making it sound brighter and menacing in a sense. I'd prefer the current arrangement as is.

mc776:
It's disgusting and harsh! And it's a piano...

mc776:
I actually understand your point, but I think it's overridden by the fact that it *actually sounds like the track was bugged*.

Anyway, I'll copypaste the comment on Discord from the only other person to provide feedback:
> I really like this, it vastly improves this MIDI.

EnzoCarozza:
> I actually understand your point, but I think it's overridden by the fact that it _actually sounds like the track was bugged_. Someone playing a boomer shooter like Freedoom simply does not expect a piano here and the map itself doesn't give any particular reason to screw with that expectation like this.
> 
> Anyway, I'll copypaste the comment on Discord from the only other person to provide feedback:
> 
> > I really like this, it vastly improves this MIDI.

This was me.

mc776:
> Ehh, i feel its more of a downgrade. The piano was more fit for this "tense" type of composition, the strings having no release make them sound very insistent, and the arrangement already has strings so the melody fades in the background, the piano has a stronger attack and more force, making it sound brighter and menacing in a sense. I'd prefer the current arrangement as is.

I just realized I misread a key part of this and thought you meant you'd intended the previously existing strings should sound very insistent, etc., when in fact that's your complaint about the proposed change to the tremolo strings.

The dissonance and darkness and insistence actually fits this map very well. It's a relentless and often very poorly lit slog through some highly unforgiving terrain and monsters that keep showing up when the player would want to rest and get their bearings.

I'm not too worried about the melody being lost because, frankly, it's a single constantly repeating dissonant chord motif that doesn't ultimately go anywhere. If anything, relegating it to a supporting quasi-bassline actually really helps prevent it from becoming this annoying insistent voice not-quite-finishing the same message over and over again.

selliott512 (approved):
I don't have strong feelings about which one is better, but I do agree that the piano seems out of place given the music in the surrounding FD2 levels.

PR #1081, opened 2023-08-26
music: tweak instruments, volume on d_dm05.

mc776:
Take it down from "actually physically painful" to "blaring" - the song is more singing to you (if still aggressively) than shouting at you.

Help bring out some of the harmonics and the bass riffs.


KorpKat:
I think it would be better of with a new piece, im not sure how good this track fits for DM altogether. But as it stands it definitely sounds less obnoxious (as good as it can with the instrumentation i reckon).

KorpKat (approved):

mc776:
Agreed!

Xindage (approved):

PR #1082, opened 2023-08-26
graphics: new Phase 1 ending sequences.

mc776:
> The massive brutes collapse onto the ground, falling apart bit by bit. The starport is yours.

> You kick the corpse in anger, and it feels good. You kick it again, and a remote falls out. You press the button, and the ground gives way to reveal a teleporter.
DOOM00

> The ship rumbles as she wakes up; you think of Earth as she lifts off.
DOOM00


mc776:
> click

The alternative to keep it readable is to make exaggerate the technospider's corpse by setting it on fire:


Thoughts?

mc776:
In case anyone's concerned about style clash with the E2 end pic, I've also prepared this:
victory2

Xindage:
seems good for me

mc776:
I don't think there's anything for me to do further (E1 end is just a text), can you approve?

mc776:
The more I look at this the less happy I am about that VICTORY2 - it's more dramatic and aggressive, but doesn't actually tell more of a story than the current, so it just looks Poochie-ishly annoying.

I'll try to come up with something better...

mc776:
> The massive brutes collapse onto the ground, falling apart bit by bit. The starport is yours.
DOOM00

selliott512 (approved):
I like the apparent skyward curled first from the serpentipede.

PR #1083, opened 2023-08-27
sprites: make SKAG blasts symmetrical.

mc776:
This is one of those tiny little things that's bothered me for years but I always assumed someone out there knew better.

If that someone objects to this I'll close the PR (but I won't promise not to stick this in my autoloads), but this definitely felt better to shoot than the current SKAG.

(And if that someone does make themselves available: why *does* the bright part switch sides between frames?)


mc776:
(And if someone agrees in principle about the symmetry but wishes to differ as to which side to use (or which part of which side), I'm open to suggestions.

Xindage:
while i see some improvements, like in the first frame i see some regressions for example the third frame because the symmetry there looks "forced" while the asymmetric one looks natural, still i'm neutral towards it because its a thing that for me dont even really matters, i doubt that we'll even pay attention to these little details in the frenzy of a deathmatch round with 4 or more players at once.

mc776:
So right now the feedback is:

1 for the change
1 abstain

Can you approve it?

axredneck:
I like old (asymmetric) one more.

mc776:
tbh i only really care at this point about the opinion of the original artist, if they had some kind of worldbuilding reason (something comparable to chirality in electromagnetism) why it looks like that... aesthetic improvements can be made later but the asymmetry suggests gameplay information that isn't actually in the game mechanics at all.

mc776:
I'll give this some more thought - it's not worth an intermediate solution.

PR #1085, opened 2023-08-28
sprites: PA/DB shotgun redraws by Korp.

mc776:
Incidental update to the pickup sprite.


mc776:
oh wow the perspective is *really* different now that I see the diff side by side like this.

the new design is also *much* simpler, for better or for worse...

nful2:
Looks good! But should it be gloved? New first sprite is not gloved.

KorpKat:
> Looks good! But should it be gloved? New first sprite is not gloved. Also, I think pure red for shooting impact seems too bright/overwhelming in a way? Maybe mix red with orange? Do the same with minigun

Im planning on giving gloves to all weapons, i organized a small poll over on the discord server to fully decide, rest asure sprite inconsistency wont be an issue.

mc776:
Affected weapons: shotgun, SSG (already gloved), fist (i've got a WIP but gloves can be added in the final), pistol (people have been clamouring for a replacement for ages).

None of the other weapons show hands.

At this point I'd rather that both shotgun flashes were red and both bullet flashes were orange rather than all red. (I'd *really* rather they were all orange and looked more like conventional firearms, but I seem to be outvoted on this one for now.)

selliott512 (approved):
I like it. They seem higher tech, and they take a bit less screen real estate. I do see that four of the modified PNGs are color type 6 (RGBA):
```
sprites/sgn2a0.png
sprites/shota0.png
sprites/shtfa0.png
sprites/shtfb0.png
```
It should be color type 3 (indexed), but that's ok. Maybe we'll have a PNG cleanup someday - they certainly aren't the only ones.

PR #1086, opened 2023-08-29
sprites: new backpack by Cascade.

mc776:


nful2:
Those straps seem like bamboo sticks for me. Also, the upper part I think should be elongated to be in same line with the pockets in bottom.

mc776:
......are we even looking at the same sprite? o_O

Xindage (approved):

PR #1088, opened 2023-08-30
patches: new midbars by Korp.

mc776:
No visual conflicts with anything, can be merged immediately.


selliott512 (approved):
Looks pretty similar to me, but if you guys want it it, that's good enough for me.

PR #1089, opened 2023-08-30
graphics: Improve the CD-ROM graphic

fragglet:
The original graphic was derived from a scan of a CD-ROM that I scaled down, aeons ago at the start of the project, and it isn't very good. I decided to try my hand at some pixel art and make a new one. I think the end result isn't too bad.


Calinou (approved):
Looks great!

mc776 (approved):

Issue #992, opened 2023-06-25
manual: common terms for things should be translated?


mc776:
The descriptions of health, armour, ammo and hazards already are translated into Spanish in the Spanish manual, which just makes the weapons and monsters and powerups stick out even more.

PR #1091, opened 2023-08-31
manual/dehacked: clarify and update.

mc776:
Some probably overambitious attempts to address #992.

[Note: I do not speak Spanish. Some longer translations are copied almost verbatim from DeepL output. Efforts have been made to parse the basic grammar of the output before committing anything but idiom is by no means guaranteed. If you find *anything* odd (that isn't obviously motivated by alliteration for the IDBEHOLD cheat), or any item/monster names already established that I'm not using here, please go ahead and fix it or tell someone who will! (Or tell me in the replies to this PR if you manage to catch it that soon)]

**General**

Rephrased the "Installing Freedoom" to make it a little easier on the kind of person who would actually need it: vanilla compatibility is also implied. FreeDM is also mentioned.

The armour example has been rewritten so we're only dealing with amounts the player will lose, instead of making the reader do internal arithmetic to track what went where.

The colorblind accessibility section for the keys is now just part of the description of the key pickups, and is also present in the Spanish manual. The keys are presented in the same order as they appear on the status bar and I've added a mnemonic that I found helpful.

Loading quicksaves is clarified: no need to die and menu is still available.

Miscellaneous rewordings to make things more to the point and avoid overexplaining.

Removed some references to "higher-level" and "lower-level" monsters. Someone who actually needs to be told this stuff might think this was an RPG thing.

Converted some CxMy references in the cheat code section.

**Ammo**

It's needlessly confusing to have "ammo" be used for both the pistol/minigun ammo and ammo generally. The gun pedant in me died a little inside, but the pickup messages now say "bullets" in line with both "bullets" and "balas" in the manuals.

(I've floated the idea of replacing the ammo box sprite with something without a label, but that's fairly low priority since everyone can see themselves touching the green square thing and only seeing one number go up.)

Clarified that backpacks don't stack their carrying capacity increase.

**Weapons**

Spanish weapon descriptions now bold the name as they are bolded in English.

Handgun description now uses the pickup sprite.

Pump shotgun's spread is clearly neither tight nor long for anyone who is playing at higher resolutions than vanilla 320x200. And it's not the main workhorse weapon in Phase 2.

Double-barreled shotgun isn't "twice" as powerful as the pump by any measurement. (Even its rate of shell consumption over time is more than double.) The much dimmer and unorthodoxically distributed red flash in #1085 can provide an excuse for the damage disparity for now.

SKAG no longer implies the ball does radius damage.

**Monsters**

The animal names are an evocative metaphor but the poetry is lost in doubling down on them in the descriptions when every single one is clearly taxonomically not that thing. ("Worm" was left untouched however since really that could mean any long wiggly thing.)

Matribite and hatchling image alt text should be capitalized for consistency; "hatchling" de-captalized in matribite description.

Painlord Spanish description seemed to imply it could only survive 5 missile *explosions* not direct hits, whereas the English is a bit clearer by using the same word for both painbros.

Large technospider and tripod explicitly said to be immune to explosions (but not SKAG).


manual/manual.adoc
```diff
@@ -239,23 +242,26 @@ monsters with your bare hands. Ammo: None
...
-3 | The main “workhorse” weapon; fires seven pellets
-in a tight cluster and is effective over both short and long ranges against
-low- and medium-level monsters. Ammo: Shells
+3 | Shoots seven pellets in a fan pattern, letting you hit multiple
+targets or one big one. Ammo: Shells
 | **Double-barrelled Shotgun** +
 image:../sprites/sgn2a0.png[Double-barrelled Shotgun] |
-3 | Twice as powerful as the pump-action shotgun
-but less effective at long ranges; good at short range against crowds of
+3 | The simpler flash hiding system guarantees better
```
fragglet, line 256: I'm not sure what this means. Flash hiding system?

manual/manual.adoc
```diff
@@ -239,23 +242,26 @@ monsters with your bare hands. Ammo: None
...
+projectile fragmentation for almost 50% more hits per shell
+across a wider spread. Good for short range against groups of
 enemies. Ammo: Shells
 | **Minigun** +
 image:../sprites/mguna0.png[Minigun] |
-4 | Conceptually like a faster version of the basic handgun, but
-eats ammo much more quickly. Ammo: Bullets
+4 | A much better use for your clips than the handgun.
```
fragglet, line 262: A bit confusing since clips aren't the only way to pick up bullets?

manual/manual.adoc
```diff
@@ -349,21 +350,25 @@ You may also encounter any one of these special items while exploring:
...
+image:../sprites/bskua0.png[Blue Skeleton Key]
+image:../sprites/yskua0.png[Yellow Skeleton Key]
+image:../sprites/rskua0.png[Red Skeleton Key] |
 Allow you to open certain locked doors and activate locked switches.
 Usually essential to be able to progress, although they sometimes allow
 access to secret areas.
+
+If you can't see the colors, check the center of the card or the horns
+on the skull: +
+• the first two letters of "Azure" have lots of diagonal lines. +
```
fragglet, line 365: I'm so confused. Azure? The planet horizon?

manual/manual.adoc
```diff
@@ -380,33 +385,15 @@ normal damage until the end of level.
...
+| **Negentropic Surge** +
+image:../sprites/megaa0.png[Negentropic Surge] |
+Maxes you out to 200% health and armor.
 | **Vanguard Device** +
 image:../sprites/pinva0.png[Vanguard Device] |
 Makes you immune to all damage for a limited time, allowing you to get
 past overwhelming defences and unavoidable traps.
 |==========================
 
-=== Color blind accessibility
```
fragglet, line 389: Can we keep this as a separate section please?

fragglet (changes requested):
In general these are all really good changes. I have a few comments though.

manual/manual.adoc
mc776: TIL in the absolut enshittification of everything: if you look up "azure" in DuckDuckGo, you get absolutel no results whatsoever for the colour, it's *all* goddamn Microsoft.

(This mnemonic actually started out in the Spanish version where there would be no expectation that someone would have to look up the word.)

And I just remembered that the entire line "Horizon is a whole new world" isn't even in the E3 endtext anymore.

Fine I'll get rid of it.

manual/manual.adoc
mc776: This assumes #1085 will be merged. This combined with the current DBSG creates a situation where the DBSG has a bright orange flash entirely out of the muzzle, while the PASG has a much darker red flash that's also coming out of the barrel shroud holes for some reason. By adding this passing bit of rationalization here we can at least kinda harmonize both this visual difference and the fact that the PASG is so much weaker.

It's not a good explanation at all, but my hope is that the new-old DBSG gets in and the entire lore about the shotguns gets the rewrite I keep mentioning in the Discord.

manual/manual.adoc
mc776: ..........do we really need to spell that out? I don't think someone who honestly can't infer that connection is going to get far enough through E1 to *see* a bullet box.

manual/manual.adoc
```diff
@@ -13,18 +13,22 @@ image::../graphics/titlepic/titlepic.png[Freedoom Title Image,align="center",wid
...
-Freedoom is distributed as two files named `freedoom1.wad` and
-`freedoom2.wad` which contain the artwork, levels, music and sound
-effects that make up the game.  To be played it needs to be paired
-with a source port. Fans of the original Doom games have created a
-number of these; the Doom Wiki’s
+Freedoom is distributed as three files named `freedoom1.wad`, `freedoom2.wad`
+and `freedm.wad` which contain the artwork, levels, music and sound
+effects that make up the game.
+
+To actually play the game, you can use DOOM.EXE, but probably it'll be
```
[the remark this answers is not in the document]
mc776: Turns out the language flows way better without it anyway.

manual/manual.adoc
mc776: That said, I will change it since this is actually a holdover from a lore thing I was trying to do to avoid calling full cartridges "bullets", which was never in the Spanish version of this sentence anyway.

manual/manual.adoc
mc776: Still thinking about this.

On the one hand, I stand by my view that splitting the key information is not good.

On the other, I *do* think that "virtue signalling" is actually helpful and works well with the new "free as in freedom" section - especially to clarify that we're not *that* kind of fossbro outfit.

Maybe we can just add a blurb about accessibility features and considerations generally...

mc776: something like this, just before the "freedoom is about freedom" section:
> === Accessibility
> 
> Freedoom is a volunteer project that receives contributions from a wide variety of people with differring abilities and needs. The community tries to address what accessibility issues there may be with Freedoom but there is always room for improvement.
> 
> Some areas that have received attention as of fall 2023 are described below. This remains a work in progress and the list is incomplete, and contributions and feedback are welcome:
> 
> * **Color blind accessibility:** Graphics are selected so that nothing is distinguished solely by hue. Each key sprite has different markings, armour designs have different silhouettes, and switches have up/down positions; where color is the main difference, it always accompanies a change in brightness as well. (The pain brothers are confusingly similar even for those who are not color blind. :))
> 
> * **Controls:** Areas that can only be accessed by straferunning are discouraged, as it may not be available on certain controller configurations.
> 
> * **Cognitive:** *<okay i know for a fact that i've had a bunch of shit merged
> that was primarily motivated by accommodating my ADHD but i can't remember any particulars at the moment>* ...

manual/manual.adoc
mc776: The new explanation is even flimsier but at least I'm not trying to pack multiple firearms concepts into 3 words I can't even begin to figure out how to translate without DeepL or Google. The sort of person who'd notice this new explanation is bullshit will probably know the origins of the bizarre game mechanics the writer was stuck with trying to explain; the sort of person who doesn't notice or care should be fine with the repeated "strength" words giving the proper vibe.

manual/manual.adoc
mc776: I went ahead and added it.

mc776:
One thing I like about the pickup messages now is that all the consumables and supplies ("a"/"an") are latinate (or at least French) iambs, while the game-changing singular items ("the") are germanic monosyllables.

manual/manual.adoc
mc776: #1085 now includes both shotguns so I *could* introduce the warp gun theory into here, but tbh the wording that's in the PR now is vague enough to include even that, so I'm leaving it alone.

mc776:
So I think we've officially exceeded the number of commits on this that I had on my "commit every little change on the manual because I can't preview" working draft.

If this gets approved, my intention is to squash with a much simpler commit message.

mc776:
Added a much-needed warning about savegame overflow buffers in the saving game section.

manual/manual.adoc
fragglet: I'm sorry but I really dislike this change.

I think your take on this being about "virtue signalling" is completely off base. The text here serves a specific practical purpose - there is a visual language in play that is non obvious (eg. the way the horns point matching the visual shapes), and the purpose off this section is to explain it. It's nothing to do with being performative or boastful about how accessible we are. This is exactly the kind of thing that *needs* to be documented in the manual. As it currently stands, you've almost deleted that entirely.

I'd be curious to see if anyone else got the same impression about this section because I'm not understanding this take at all. I read through the text again several times and I don't get it. Maybe I'm blind to it because I wrote it. It certainly was never my intention.

I don't like the current version because the new accessibility section really is just performative and I don't think adds anything substantial that needs to be documented.

Anyway. The core of the change as it stands is that it merges the part about color blindness into the section with the keys. There's certainly a reasonable argument for doing this - the two are logically connected. The reasons I think this should be its own section are:

1. I don't like bloating the keys section by adding too much text. Putting too much text inside tables undermines the visual benefit of placing information into tables, plus I think it's just bad style in general.
2. Having a dedicated section makes the information more discoverable because the section has a heading that catches the eye. If you're color blind and you scan through the text, you can notice the heading "color blind accessibility", and think "yes, I'm color blind, I should probably read this". For the same reason, I think it's important that we explicitly use the term "color blind", which your change removes.
3. Having a dedicated section allows us to use illustrations to demonstrate the visual language that's used. There is an illustration in the section as it stands; it could probably be improved.

manual/manual.adoc
```diff
@@ -41,13 +44,16 @@ episodes of eight levels each, while _Phase 2_ is a single, 30 level campaign.
 This gives 62 levels to play through, and there are also secret levels -- if
 you can discover how to reach them.
 
+FreeDM (`freedm.wad`) is a monster-free set of levels, made specifically
+for player-versus player. For how to host a multiplayer game,
```
fragglet, line 48: Missing hyphen, should be player-versus-player.

manual/manual.adoc
```diff
@@ -167,19 +175,19 @@ messages shown when you collect an item.
...
 to defend yourself.
 Portions of the levels may be inaccessible until you find a particular key, or
-find a switch to open a locked door. This gives a puzzle element to the game in
+find a switch to open a passage. This gives a puzzle element to the game in
 addition to the action.
 
 By default, the keyboard cursor keys will move forward and backward, and turn
 left and right. The _Control_ key fires the current weapon, and the spacebar
-will open doors and activate switches. All source ports allow you to rebind
+will open doors and activate switches. Consult your source port on how to rebind
```
fragglet, line 190: Consult the documentation for your source port? Or should we instead recommend checking the settings for your source port?

manual/manual.adoc
```diff
@@ -226,9 +234,12 @@ ammunition>>, along with the maximum of each you can carry.
 [[weapons]]
 === Freedoom’s Weapons
 
-You start the game with only a handgun, 50 bullets and your fists to fall back
-on once they run out. You’ll want to find yourself some better weaponry as
-soon as possible.
+You start the game with only a handgun, 50 bullets and your fists.
+Exploring the level will reveal more weapons and ammo that you can pick up and use.
+
+Pressing the numbered key on the keyboard switches to the given weapon
```
fragglet, line 240: Number keys, not numbered.

manual/manual.adoc
```diff
@@ -107,6 +112,12 @@ and killed monsters come back to life after around 40 seconds.
 [[savegame]]
 === Loading and saving the game
 
+[**Warning:** the original Doom program has a bug that makes it crash
```
fragglet, line 115: I think this would fit better at the end of the section, since it's only applicable to a single source port.

manual/manual.adoc
```diff
@@ -239,23 +250,26 @@ monsters with your bare hands. Ammo: None
...
-monsters but it’s best to upgrade before taking on any tougher opponents.
+| **Handgun** +
+image:../sprites/pista0.png[Handgun] |
+2 | Your starter weapon. Its main purpose is to let you fight your way
+to a better weapon, and to hit shootable switches without wasting a second bullet.
 Ammo: Bullets
 | **Pump-action Shotgun** +
 image:../sprites/shota0.png[Pump-action Shotgun] |
-3 | The main “workhorse” weapon; fires seven pellets
-in a tight cluster and is effective over both short and long ranges against
```
fragglet, line 248: I'm okay with this change, but to give some context for the "effective over long range" thing: one of John Romero's design principles when designing Doom's weapons was that no weapon should become entirely useless when picking up a more powerful weapon. The shotgun fires less pellets than the SSG, but has a tighter spread that remains more effective over a long distance.

fragglet (changes requested):

manual/manual.adoc
mc776: I always assumed there were other ports out there that did this as well. If not then this is a good deal easier to write.

manual/manual.adoc
mc776: Oops, didn't realize I'd completely removed the words "color blind".

I'll add the original section back but will still keep a mention in the key section itself since it is helpful information even for people who aren't colour blind, or whether or not they're colour blind might just happen not to be cognitively able to go all the way to the end of the section.

> I don't like the current version because the new accessibility section really is just performative and I don't think adds anything substantial that needs to be documented.

That new section was actually because I thought about that "virtue signalling" comment again, decided against it and ended up swinging in the opposite direction. For what it's worth, writing that new section was actually what got me started on #1104.

selliott512 (approved):
This looks great. I like the tutorial. I didn't see anything that needs fixing other than possibly the "on on".

I believe fragglet's suggestion that "Color blind accessibility" continue to be its own section has been addressed. Actually, everything has been addressed that needs to be, as far as I know.

I've approved, but approval is also needed from fragglet.

Squashing the commits may be a bit tricky since the later commits depend on changes that were made in master, and merged into the branch. I can help unravel it, if you want.

selliott512:
Actually, I should mention that I did not review the Spanish manual.

manual/manual.adoc
```diff
@@ -115,23 +120,25 @@ if you die.
...
-Pressing _F6_ again in the future will overwrite your quicksave slot with
-just a couple of keystrokes. If you die you can also restore your
-quicksave slot by pressing _F9_.
+Pressing _F6_ again in the future will overwrite your quicksave slot
+without navigating the menu.
+
+You can restore your quicksave slot with the menu or by pressing _F9_.
+
+[**Warning:** the original Doom program has a bug that makes it crash
+when you save a game while too much is going on on the level. Chocolate
```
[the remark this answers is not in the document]
mc776: I did!

mc776:
If it's a lot of work to untangle it now I can close this and re-post the final.

mc776:
Yeah I think I'll try to reup... even if the commit history could be made decent the commentary on the PR itself is all over the place.

PR #1094, opened 2023-08-31
manual: Add a section about Freedoom's true goal.

fragglet:
I've seen multiple Youtube videos recently that talk about Freedoom as though the "free" part in the name refers to price. It's an understandable misconception but it deserves to be corrected. Let's add a short section to the manual to discuss this.


mc776 (approved):
hell yeah

PR #1097, opened 2023-09-01
patches: new bronze set by Korp

Talon1024:
I also re-did the rw25 series based on Korp's work


mc776 (changes requested):
Looks great but I'm just a tiny bit concerned about the change to the cracked grey rock patch rw18s' tiling pattern - the seams are very, very visible:
seams
It tiles correctly but there's this clearly visible jump from one shade to another.

Could the shading be smoothed out a bit more so there isn't this jump in shade?

Talon1024:
I'll remove the gray rocks for now.

mc776 (approved):

fragglet:
The height of the top bar on the WALL02_* patches has shrunk a couple of pixels. Sometimes WADs make use of that alignment. It's not a huge deal but it might be preferable to keep the original height

mc776:
For what it's worth, I'm running around E1M9 which uses this WALL02_* stuff a lot and the new update hasn't really affected anything noticeable:

DOOM05
DOOM04
DOOM03
DOOM02
DOOM01
DOOM00

PR #1099, opened 2023-09-02
sounds: fix dsbspsit.

mc776:
I don't know what happened but apparently something got corrupted in the original post that gets fixed in SLADE import but not DEUTEX build.

I've actually tested this one using FD's build system. None of the other technospider sounds were affected.


selliott512 (approved):
Regardless of what happened I've confirmed that it sounds better now. Neat sound.

PR #1102, opened 2023-09-04
sounds: new DSPEDTH.

mc776:
[Undead003's old sound from way back in 2015](https://github.com/freedoom/freedoom/commit/c48644582cc46ec67678aa3fcf75f0e6c6539100) never got heard ingame because of a typo, and only got rediscovered while reviewing #661.

This is that sound, cut off and faded to match the pain elemental's death animation duration and mixed with a slowed-down DSBAREXP and the hatchling sound.

I confirm I've actually tried building this one.


selliott512 (approved):
I like it. Played outside of the game old sound sounded like it had clipping, or some sort of distortion. The new one has a hint of dinosaur, or some sort of scary monster.

That being said, I find this to be surprisingly hard to hear in game. I tried MAP32 where there are lots of matribites behind the door marked with the yellow skull, and I found it hard to hear with all the hatchlings being launched, and other things.

selliott512:
Also, interesting story behind the misnamed file. I probably should've looked into what happened more than I did, but at least I didn't delete the good version.

mc776:
Unfortunately it is noticeably quieter than the old one, for the same reason why there's so much awful clipping in the old one, but it's also missing that distinctive distortion that makes it easier to pick out in the midst of the numerous other fireballs, explosions, gunshots, other monsters and actual attacking hatchlings.

At least it's enough that it's not a weird silent death explosion animation.

PR #1104, opened 2023-09-05
levels: various QOL fixes.

### LINKED ISSUE #996: levels: outdoor linedefs need to face the right way.
https://www.youtube.com/watch?v=9GoSRw4TDw0

To summarize the vanilla bug behaviour:
1. Have two sectors, both with sky ceilings.
2. Make one of them have a floor high enough to shoot when you're standing on the other one.
3. Set the linedef between them so that its front side is towards the "inside" of the wall and the back facing whoever's standing on the lower floor.
4. Stand on the lower floor
5. Shoot a rocket at this 2-sided linedef.
6. It disappears, having registered as having been shot into the sky.

A lot of Freedoom's maps have no accommodation for this. These lines are everywhere.

Inuk's got a script for this for Ultimate Doom Builder but I can't use UDB, hence posting this issue [EDIT: updated after some discussion on Discord]:
```
`#version 4`;
`#name Select raised walls facing inwards`;
`#description Selects any non-action linedef that faces inside a raised sector. This includes selecting walls that can cause the disappearing rockets bug.`;
UDB.Map.getLinedefs().filter(ld =>
    ld.back !== null &&
    ld.action == 0 &&
    ld.front.sector.floorHeight > ld.back.sector.floorHeight
).forEach(ld => ld.selected = true);
```

Basically we need to do this (also credit to Inuk for the diagram):
noingrownhairs

### LINKED ISSUE #1109: E4M2 - The safe path back from blue key area is inaccessible.
Here is the place on the map: 
screenshot 386

When you get the blue key and go out through toxic waste, you can go back (through a teleporter or do it the normal way). When you do so, you can open the blue door, but it's blocked by a platform:
screenshot 385

The only way out is to through the toxic waste again, this time without a hazard suit.

### LINKED ISSUE #1110: E4M9 - You can accidentally lock yourself out of an area
In an area at the end of the map you get ambushed and have cell door closed behind you. Here is the place:
screenshot 392

You can escape the place without opening the cell door by dropping into toxic waste and riding the lift up. If you do so, the cell door stays closed and you can't go back into the area:
screenshot 391

mc776:
Generally there should be nothing *necessary to finish a level* (i.e., not a secret) that requires any of:
- straferunning;
- extremely sensitive timing that could softlock you if you're on keyboard, lagging in multiplayer or have motor issues;
- checking only for a sound cue that something has happened;
- remembering how to distinguish two visually nearly identical areas; or
- backtracking to a previous area on the map that you had previously been given no reason to revisit.

I haven't caught all of them by any stretch of the imagination but it's a start.

Also some regular fixes of varying degrees of severity, as well as more "floating item" glitches found thanks to Inuk's new UDB script.

E1M1
The main goals here are to make things actually easy on easy skill, and to make the lower route actually a viable starting choice instead of a shitty gotcha (or, more charitably, an extra challenge). Tested by using only keyboard in Chocolate while trying to avoid moving and turning at the same time.
- In the courtyard room, there is now in easy a dead shotgunguy and a shotgun, right by the door.
- One of the starting medikits has been moved to the door after the lowering pillars, swapping places with the stimpack.
- Removed the soulsphere in the starting corridor. Sorry, it's just too much especially given the *second* one outside, and it really doesn't help mitigate...
- In accordance with the above, removed pinky no. 244 and serpentipede 241 from easy skill. An ambush being totally surrounded by 3 serps at once is not exactly an E1M1 easy mode sort of encounter!
- The easy-only shotgun in the southern route is now moved to the climbable crate pile, rather than requiring the trick jump.
- Pinky 214 has been moved somewhere where it's less likely to be hit by a stray bullet from below and alerted prematurely.
- Removed secret flag on that crate you can jump onto, since the only reward is maybe a dropped shotgun in a level filled with ammo and there's nothing else indicating you should be trying to get up there, and the soulsphere pillar is a much better tutorial for jumping anyway.
- The column-lowering switch now starts in the more neutral-looking "1" position.
- Replaced the health boosts in the exit to the upper cave with a trail leading into the cave, to let the player know right away that this is not a decorative area. (It's not even counted as a secret!)
- Southwestern blue key corridor had a lone pinkie on easy. Added two zombiemen instead.
- Removed easy-only shotgun from said corridor. Zombieman 88 is now on easy as well.
- Removed shotgunguy 96 from easy skill.
- Removed zombieman 176 from easy skill and added an easy-only zombieman behind the door.
- First door switch sector needed to be 72 units high to avoid tutti-frutti effects.
- Removed fake contrasts on corner "AGM" pillar.
- More attempts to fix #996.

E1M3
- Floaty health boosts near sector 74.

E1M7
- Shells no. 174 was floating on that bit of wire connecting the two computer banks around it.

E1M9
- Fixed some textures around the big blue-trimmed lift and removed an extraneous use line that triggered a faraway lift for no reason.
- The red key bridge lowerable section is now textured differently from the rest of the bridge.

E3M2
- The lava *is* the light source down there. It should be *brighter* in the confined tunnels not darker.
- Added a little catwalk to reward players who can make the (non-straferun-required) trick jump across the bridge.

E2M7
- Floaty armour boosts 57 and 62.

E2M9
- Floaty anti-nukage supplies 102 and 132. Moved 133 as well just in case.
- Floaty gibs 139 by teleporter pad.

- Lots of tutti-frutti around the teleporter pad to the east of the floating gibs. ~~Never point a COMPUTE2 at a map you are not willing to destroy.~~

E3M3
- Floaty candle in sector 232. All moved for symmetry.

E3M5
- Teleporter platform to get back up to the catwalk from the northeastern blood maze is now clearly marked as having a switch, as it is a mandatory progression rather than a secret.

E3M8
- Replaced one hanging gore actor. See #941.
- Made the panel behind that replaced gore actor a marble face.

E4M2
- Closes #1109 by making the bridge to the one-way blue key route raisable repeatedly (but collapsible only once). While it is raised, the bridge is also usable as a safety lift, as the normal safety lift is inaccessible if you fall in on the north side. There is now only one teleporter onto that patio/boardwalk thingy, down where the radsuit was; the only way to get back to the blue key after a successful run is to raise the bridge again, crossing which a second time will (a) undo the raising floor in the next room and (b) not cause the bridge to collapse again, thus allowing free movement back and forth without touching any hurtfloor.
- Adds some stairs out of the slime pit in between that switch and that bridge.
- That raising floor is no longer crushing and if you're caught in it you can slip out back into the main slime area. This avoids the (extremely improbable but theoretically possible) situation where you can actually get softlocked by getting caught in there but not dying from the resulting damage.
- Made the silver trim on the switch in line 2287 almost fully visible as it was already awkwardly peeking out slightly and all the obfuscation did at this point was risk it being missed.
- The slow up elevator in the northwest is now activated by a switch *inside* the car, and made both switches repeatable. The old regular-lift use line is now a second move-down line, just in case you want to get back up there after someone's gone up in multiplayer and then rocketed themselves.
- The slime pit with the bridge leading to those elevators now also has safety stairs and the bridge itself is a usable lift.
- That first chunk of pre-raised bridge leading to the exit is now a usable safety lift.
- "Not on map"'d a bunch of control sectors.
- The lines on sector 592 were a flat-bleed trick and should not have had textures.
- Line 1728 now starts in the 1 position rather than 2, because I ran three laps around this area thinking I had already hit this switch.
- The big super-bright lift en route to the yellow key is now called down by using it directly. The remote switch simply raises it to the correct position so it can be called down.
- Armour boosts on east side of bridge were close enough to be subject to the floating bug, stopped only by the impassable line.
- Fixed some textures in the cross room and the secret exit room.

E4M8
- Got rid of some fake contrasts on the noodles at the start.
- Added a radsuit for the northwest switch. While it is possible to avoid damage even without straferun, unless you've got a tic counter display and can time it to the damage interval this is basically RNG.
- The water flat on top of the lowering wall in the east was very, very noticeable. The switch is now stepped on instead of hit; the secret nearby had to be made a little more obscure as a result but not impossible to see.
- Removed asymmetrical doortrak on the slime bridge on the southeastern piston switch.
- The linedefs of said slime bridge pit are flipped so a deathmatch opponent trying to grab the berserk in there is not magically immune to rocket blasts. (see #996)
- Realigned the four pistons by the gate to the starship. They also reveal moving parts when activated - not nearly as good as the crushers on the original DI, but better than nothing.
- Made the southern walls use PLAT1 to make it more obvious that those walls will lower later (with the added bonus that they match the four pistons).
- The southern light bronze area now has a strip to guide the player towards the switch in case they lose track of their direction while fighting monsters and forget to explore inside that area, as well as to better distinguish it from the southeast.
- The gate threshold to the southern light bronze area now matches that of the pre-opened southeast.
- There is now actually a threshold where you can tell where the starport ends and the ship begins.
- The two light bronze areas are a bit too similar-looking. Added a few health boosts so the player can spot them and know this is an unexplored area.

E4M9
- Closes #1110 by removing the forcefield as soon as the player jumps off the switch ledge into the slime pit.
- Added safety lift to southeast nukage pit.
- Reduced all green nukage damage to 10%.
- Light no. 117 submerged illusion pit was smaller than the light hitbox and the wrong shape.
- Floaty serpentipede no. 236.

Map05
- Numerous attempts to fix #996.
- The switch letting down the red key now also releases the trilobites, removing the need for those additional walk lines which were preventing the player from activating the switch at the usually permitted distance.
- Massive makeover to the first courtyard. The outside walls are now orange only on the west and the insides are tan.
- Darkened all outside areas so the distance fade does something, and also to better fit the sunset look.
- Made all TANROCK5s STONE6 instead where it should be lit up by the "sunset" implied by the new STARG walls.
- Got rid of a lot of the weird unnatural zig-zags in the river.
- Lengthened the red armour secret sector so it's impossible to grab the armour without triggering the secret.

Map01
- Moved minigun 70 deeper into the alcove. It wasn't actually floaty but it looked bad there anyway.

Map09
- Added a couple health boosts on top of that crate that needs to be moved away, to remind the player that they might need to revisit that spot later.
- Changed the texture on the door leading into the lava pit since the new one doesn't really work with it.
- Moved the southeast missile launcher secret so the entrance is directly overlapping the fake crate edge you need to walk on. While the *degree* of difficulty was acceptable for a secret, the *cause* of it - the constant snagging of the invisible square corners of the player hitbox - was not.
- Floaty serpentipede corpse in red cracked area.

Map11
- The lift going down into the yellow key room requires a switch that is out of sight from the lift itself, which is not clearly marked as a lift to begin with. The only real way to realize what's going on if you don't already know about the lift is to locate the sound and immediately turn to investigate before the lift comes back up. I thought this was annoying when I first did my big overhaul of this map, but ultimately left the basic mechanism alone out of an abundance of caution; however, with the recent discussion of accessibility in the proposed changes to the documentation I'm revisiting this. That upper switch now lowers a wall to reveal the lift which is triggered by a walk line. (A second trigger to lower the wall is added in the room below in case you come in from that way.)
- The lower far switch on that same lift was actually literally *impossible* to make on keyboard and no straferun (and no vanilla wallbounce exploit), even if I change it to a regular lift instead of fast. This is completely unacceptable for something necessary to progression (rather than an obscure secret). The lower switches are now permanent repeatable floor-lowerers, while the line crossing from the lift into the lower chamber is a permanent repeatable floor-raiser, with the line crossing into the lift from above being a simple lift line.
- The lift with the bronze texture on the front is also no longer time sensitive - it's a floor-lower switch and a floor-raise walk line now.
- Retextured the stairs out of the water in the eastern branch so it's not an unreadable mess of criss-crossing grey lines.
- Realigned the new skull switch texture in the skull room.
- Floaty health boosts 715 and 719.

Map12
- Extended the corridor in front of the main yellow door and made it open instead of a dead end, then extended the windowed building by the tracks to the southwest. This cuts off the view from the corner to the northeast where all that HOM is. Since there's nothing back there for the player and most of the activity is going to be eastwards, the technospiders that spawn in hard generally will not get stuck, though the player does have quite a bit more cover. (Meanwhile, opening up that corridor makes this fight a lot less claustrophobic.)
- Moved corpse 653 out of being exactly on the lift boundary.
- Fixed the misaligned textures on the big train gates.
- Added yellow key indicator on both sides of the non-see-through yellow doors.
- Simplified the geometry of the northern corridor leading out into the southwest crate area.
- The green panels used to have a big SUPPORT3 section where they touched the lifts. Only some of these remain and they look like accidents. Replaced all of them with warning stripes.
- Made the crates on the southwest near the lift full height again. There does not appear to be any visplane issue resulting from this.
- Numerous miscellaneous texture alignment fixes.
- Many, many attempts to address #996.
- Added a secret behind the train. Just a little guy hangin' out.

Map14
- Moved armour boosts inside minigun secret door into the chamber itself.

Map15
- Thickened the lowering panel with the pump shotgun to prevent possible grabbing through the wall.

Map17
- Floaty shotgun on pillow.

Map18
- Floaty missiles around launcher.
- Floaty clip 100.
- Addressed #996 in the courtyards.

Map19
- One of the stealth worms was stuck in a burning barrel.
- Removed monster block flag on line 2083.
- Unmerged and remerged a few identical sectors to better match the intended sound travel.
- Moved the lowering wall switch to the left so you're not stuck trying to hit it from a step above.
- Flagged line 281 as a monster blocker. This allows the player to always be able to make the jump onto that bottom stair without being blocked by the octaminator.
- That octaminator is now a pain bringer in easy mode. The far end of that platform path is well outside the maximum vertical autoaim range in vanilla, which means that to actually hit the octaminator without up/down aiming you'd have to be on one of the later platforms - i.e., confined to a relatively narrow area with no cover *against an opponent that has seeker missiles*. The best way to solve this is to charge towards the octaminator as fast and as soon as possible with the SSG, minigun, or ripsaw+prayer to RNGsus that you'll get a good painlock. This is not the kind of tactic the sort of person who *needs* to play on easy will think to do, or could do while also being ready to sidestep if the octaminator fires at the wrong time.
- 347 and 249 are now also monster blockers, and the worms in that slime pit have been moved to the platform just behind the combat slug since they're awakened early on and that's where you'll first encounter them anyway.
- That slime pit platform gap is not a jumping puzzle, but a game of luck whether you take damage from the slime on your way to the northwest platforms. The entire route is now connected by a thin catwalk (wide enough to progress, not wide enough to dodge while on it) and all of that slime does 10% if you somehow fall into it anyway.
- Replaced the teleport pad in the vertical platforming sequence with a lift, to minimize disorientation and going the wrong way. (In retrospect I probably could have just made the teleport destination face the pit you came in from, but the lift worked aesthetically better anyway.)
- A good chunk of that entire platforming area has been moved 8 units to the west so that things would align with the flat grid.
- Lines 307 and 309 are now also monster blockers. The worm that would be trapped between them is now moved further down the route and marked ambush.

Map20
- Removed the useless, misleading skull switch texture on the bars at the start.
- Down switch added to the upper side of the big pipe lift, in case the stealth worms are in the way.
- The door leading into the blue key arena needs no blue key; the door leading out needs a blue key. Both are marked with blue-light trim. Removed the blue lights on the first one.
- The lowering wall leading to the teleporter now uses a pipe texture.
- All 3 rockets in the blue key arena were floaty.
- The door leading out of the giant quadruped arena now has a bright flickering light.
- At least 1 floaty armour bonus in the red button brown room.
- The lift in the small brown cave area is lengthened (and the upper catwalk widened) to lessen the chances of the worms getting stuck.
- Yellow key is another case of effectively-randomly-mandatory damage. Added a path.
- Same with the lava tunnel on the red key route.
- Expanded the red key column to better ensure the player will cross that line when picking it up.

Map21
- Disambiguated "sinky" ripsaw position.
- Deleted sectors representing the pillars around the soul sphere platform as there was no reason not to use 1-sided lines here.

Map23
This is a mandatory damage exit that could kill someone even if they do everything else (more or less) right in beating this map. Worst case scenario, they "learn" from their "mistake", "realize" this is a "fake exit", and spend forever trying to find the "real" exit before giving up and learning much later that they'd actually found it before.
- You can pick up at least 20+ health right next to the exit teleporter pad.
- The actual teleport landing itself is not a hurtfloor. It is, however, extremely narrow and right in front of an ambushing trilobite. The only way to dodge its fireball is down.
- Made the exit portal huge and fireblu and pulsating and not a St. Peter's cross (which is often used elsewhere as a purely decorative thing).
- The "EXIT" map lines are now actually faintly glowing on the floor. (We'll say it's in English for the same diegetic reason demonic stuff in pop culture today is in Latin. Or, come to think of it, maybe it *is* Latin...)
- As an additional bonus edit: if you exit without finding that creepy area in the facility and have all those guys warp in, those un-warped guys are now all free to roam the entire area as soon as you fall off that high ledge. This means the necromancers can go into that crowd of zombies and raise them, as well as close range to attack you. Enjoy this epic battle - or panicked rush for the exit.
- Moved the helicopter pad lights to a more conventional arrangement, in part to give more room to a player who just wants to rush straight down to the bottom without invisi-colliding with the torches.
- Gave the start of the map a makeover for a bit of environmental storytelling.
- Floaty health boost 265.

Map24
- Ripsaw is now a minigun outside of hard. I believe this ripsaw had been added back when FD did not have to consider that the ripsaw is much more unreliable in vanilla against something as big as a trilobite. Given the near-complete arsenal from Map23 making this encounter extremely lopsided versus pistol start, this is probably an unduly conservative change.
- Adjusted the contents of the two hub side rooms. The armour is now visible from the starting view angle.
- Adjusted some of the tekgreen textures so you didn't see that blue window half buried in the floor.
- Extended the ledges to let you walk all the way around the red key slime pit.
- Adjusted the lighting and markings on the door in the long corridor leading back into the main hub.
- Moved some vertices in the green rock hall to avoid unsightly unintentional fake contrasts.
- Made a bunch of switches eye height for the player.
- Realigned the big bricks in the southeast reuins.
- The room to the second switch in the southeast ruins is now marked by an analogous appearance to the first switch, missing only the gargoyle itself. Trails of armour bonuses are also added to get the player to check that area, because the pain bringer does not ever acquire line of sight without the player actually committing to going that way.
- The pads that actually let you advance have different glyphs.
- In hard, you will still get a minigun, but in the form of a lone random minigun zombie who wanders in through the entrance teleport late into the level.
- Got rid of the fake contrasts in the exit room.
- Exit room necromancers are reduced to 1 on medium and 0 on easy.

Map25
One general comment: this map could use a more consistent and clearly visible way to distinguish switch gargoyles from decorative gargoyles. I've taken care of the two most annoying and frequently encountered cases, but a much more thorough review could be done.
- Redid the central pillar in the first room. There is no way to avoid making those lion faces look like they're supposed to be switches.
- The eastern door of the first hub has a gap large enough that going there will always force you to touch the hurtfloor. I can see the rationale in theory for making this visually distinct, but frankly I never even noticed the difference until getting hurt just now while testing it. That door now has the same platform in front of it as its western counterpart and is marked with skull poles instead.
- On the topic of marking the doors, both shells are now by the door leading to the SSG and both health refills are by the door leading to the surge.
- The blood stairs were also (RNG-dependent) mandatory damage. The streams now only merge at the bottom.
- So was the yellow key fountain. The edge is now climbable.
- So was the rightmost big cross room switch. Switch platform embiggened.
- So was the SSG. There's now a big bottom rim on the "O".
- The stairs leading down to the painlord ambush no longer have anything to the right that can snag you as you try to sweep up all the bonuses (which are now placed slightly closer to the edge of each step so you can't accidentally jump over them).
- The switch for the door leading out of the water T is now stay-open.
- There was no apparent light source in the big hall to the west. Added skylights.
- Moved the lion heads on the yellow key pillar to the top to make it more clear they are not switches.
- The first door into the sigil sequence is opened by a switch that is next to a vista overlooking a clearly unexplored area. A right-handed player's first instinct is to look that way, not towards the door; given an identical switch immediately before this was clearly operating something out of sight, and the precedent with the red key having that remote switch+window combo, the first instinct after seeing the outside unaffected is to leave the room to find out what the switch did. Redirecting the player's attention to that previously totally blank wall requires either a sound or being hurt by a monster or both. The door is now clearly marked as an advance warning.
- The silver lift near the river and shack is now activated by SW1GSTON switch right on the wall at eye level, rather than counterintuitively and invisibly recessed into additional sectors.
- The painlord ambush lift is now accessible after the encounter. A small health refill has been added there for easy and medium.
- The switch lowering the red key now has the player facing the red key. It's nowhere near as visible as I'd like but it's better than before. The same switch also opens the rescue suit closets. And if that's not enough, hopefully the nearby red skull pillars that lower until you can pick up the items on them should be.
- The water in the marble T room turns into blood again.
- The misplaced marble face texture under the marble face door in the T room is now regular marble.
- The second rescue suit appears on easy as well as in multiplayer. This technically results in three suits if you play easy multiplayer but practically you'll always grab both at once so there's no actual difference.
- "Not on map"'d a bunch of control sectors.
- Sector 59 now completely unambiguously houses two hatchlings.
- Floaty armour and medikit in the square floor halls.
- Floaty medikits by the yellow key.
- "Sinky" health refill on the newly accessible painlord lift.

Map27
- Floaty diagonal armour boosts around red key pillar.

Map28
- The starting switch takes up and takes centre place in the entire pit wall.
- The skull key signs now raise to reveal the entire skull.
- The skin panels have finally received the care they desperately needed.
- The candles on the western skull piles are all moved closer to their respective ledges to avoid floating to the ledge above them. These sectors have also been merged and the bloody part now uses the new flat.
- Floaty health refill 313. It is now in exactly the same spot as the sphere that appears instead on easy.

Map29
- Blue key pillar was doing damage. It should not.
- Moved the *potentially* floaty stealth worms in the dark pit building away from the moving platform.


mc776:
> Added a few **health** boosts so the player can spot/**be attacked by them**

they were originally zombiemen marked as ambush, but I was concerned they'd just get mixed up in the rest of the fighting anyway. A harmless, immobile, always-picked-up item right in the middle of the lane is a better marker for an unvisited area.

fragglet (approved):

PR #1105, opened 2023-09-05
dehacked: fix obituary string assignment syntax

fabiangreffrath:
Boom-based DEHACKED parsers, as used in e.g. PrBoom+ [1] and Woof! [2], expect string assignments to follow a strict `a = b` syntax, whereas the DEHACKED lump embedded into Freedoom uses `a=b` syntax for the obituary string assignment. This is an issue, because these parsers terminate the "key" part at the last char *before* the '=' char, expecting it to be a space char. I don't say that obituary strings are targeted at these ports, but things may change in the future and it will be better to get the syntax right in the first place. Thanks!

[1] https://github.com/coelckers/prboom-plus/blob/969515162c5aebea4ad7a125ee178dcad3d576ad/prboom2/src/d_deh.c#L3332
[2] https://github.com/fabiangreffrath/woof/blob/652827f4cb87718bd2abb300934917c092fb832d/src/d_deh.c#L3228


mc776:
TIL, oof and yikes... (not your PR but the fact that it would be needed)

lumps/dehacked/dehacked.txt
```diff
@@ -610,54 +610,54 @@ BGFLAT30 = SLIME13
 BGFLAT15 = AQF004
 BGFLAT31 = AQF021
 # Obituary texts, for ZDoom-based ports.
```
mc776, line 612: Can we replace this line with a comment so someone doesn't undo this fix later?
```
# Obituary texts for sourceports.
# NOTE: Boom-based DEHACKED parsers like those in PrBoom+ and Woof! require a strict
# "a = b" pattern *including the spaces*. Do not remove them!
```

mc776 (changes requested):

lumps/dehacked/dehacked.txt
mc776: Actually, forget it, everything else already follows this format and I'm not even sure why the obituaries don't. It's fine without the comment.

mc776 (approved):

mc776:
> , for **ZDoom-based** ports.

...i guess this can be fixed later

fabiangreffrath:
> ...i guess this can be fixed later

Well, obituaries strictly *are* a ZDoom-specific feature. This change just makes it easier for other code bases to adapt it. So, there's nothing wrong here.

PR #1107, opened 2023-09-07
sprites: new puffs and bloods.

mc776:
Puffs originally by Korp, edited by me for size and balance.

After seeing it in action I knew the blood had to be redone as well. Now it's a bit more "full-bodied" and viscous-looking and actually lies on the floor.

DOOM02
DOOM01


mc776:
Comparison of old, new and id blood, in GZDoom with a mod that makes the minigun fire faster.

fragglet (approved):

PR #1118, opened 2023-09-12
sprites: add handgun sights and correct flash perspective; center minigun.

mc776:
Added the rear sight as mentioned on Discord.

Made a wider, less upward-pointing flash based on the minigun - it is, after all, firing the same round with the same force. It's still smaller, though, because of the different perspective.

(It was also discovered that the minigun was markedly off center. Offsets for the entire weapon have been adjusted.)

Adjusted the offsets so there isn't quite as much movement with the trigger pull - that weird whole-arm jerk was a mistake from id that we need not emulate, and reducing the movement helps bring out the way the slide actually moves back in the C frame.

Adjusted the shape of the hand so there wasn't that awkward sudden perfectly straight line down the thumb side.

Made the light fade out after each shot instead of just blinking.


fragglet:
The new flash looks a bit big to me, I find it distracting. I know there's a real-world rationale with it being the same round etc. but not everything needs to be perfectly realistic.

mc776:
It's also an issue of perceived versus actual weapon power for me - the minigunner using the shotgun sound (which was correct for the spider mastermind since it actually called A_SPosAttack) was bad enough, without having a player think the minigun is always going to be more powerful than the pistol. (See also that note earlier about the shotgun having better range than the SSG)

How's this look? If it's good I can add a corresponding adjustment to the CHGFs.

alone
ingame

mc776:
One possible alternative (and just to be clear I have **not** committed this yet): have the less-bright frame be firing from the bottom barrel.

mc776:
So after alternatingly staring at each version for too long I was starting to hear *"plap plap plap"* in my head with the unaltered-B-frame shape-changing one so I went ahead with committing this version.

MerrittK:
I second fraggle's point on the flash, the new pistol flash looks a bit too big.

mc776:
Here's a *really* crude mockup of what the minigun could look like if it were made consistent with the current pistol flash:
DOOM00

mc776:
Looking at that I suspect my issue is less the size and more the shape - the current flash is pointing almost straight up.

Here's a crude attempt to foreshorten it without making anything bigger:
DOOM00

mc776:
I've just been reminded that the minigun's (and id chaingun's as well as the rocket/missile launchers') perspective is not where you'd expect in a normal view but a closeup of the muzzle.

This means, in FD's case at least, that the minigun barrel takes up almost twice as much width as the pistol:
image
the ratio is about 4:7.

So to scale them properly:
Current handgun is 25 wide
Current minigun is 56 wide
First-revision handgun/minigun in this PR is 40 wide

A pistol flash 32 wide would be perfectly scaled with no changes needed to the minigun.

MerrittK:
Looks better, but the flash is a bit pointy at the end.

mc776:
Honestly..... yeah. My primary motive for not having addressed it already was laziness (but also in case some folks wanted it pointy because they're used to flashes being pointy)

MerrittK:
Looks beautiful.

mc776:
I haven't changed the pistol sprites since the first commit on this PR, but the new flash is a bit brighter than the one it's intended to replace.

selliott512 (approved):
Looks good. I like how the how lights on the back of the gun give it a more futuristic look.

mc776:
The lights at the back were actually part of the original design, the glow just lasts a little longer now.

PR #1120, opened 2023-09-17
sprites: blend missile launcher muzzle flash.

mc776:
TIL while doing this that that weird circle on the C frame actually is visible sometimes.


selliott512 (approved):
Looks good. It's as subtle difference based on my quick before and after test. It looks a bit brighter, and to have a smoother boundary, which is what you probably mean by "blend".

This PR changes four PNGs that used to be index to not indexed, but we can worry about that alter.

fragglet:
Can we see a screenshot of what it looks like in-game?

mc776:
DOOM07
DOOM08
DOOM09
DOOM10

mc776:
While doing this it really hit home how weird this flash really is - the way the cloud of smoke fades it gives the impression of opening up, then retracting downwards and backwards towards the shooter's feet or knees, instead of floating up and away.

But the thought of having to open up Krita and mess around with the smudge tool for an hour or two makes my hand cramp right now...

PR #1121, opened 2023-09-18
levels: minor fixes and tweaks.

mc776:
E1M1
- All the monsters in the trench are now +ambush, so they don't move until you're down there. On easy the zombieman even faces away from the start.
- More monster blockers added to prevent invisi-blocks jumping into the starting room trench or down from the cliffs in the outdoor area.

E2M7
- Armour bonuses still floating on sector 599. They now adorn the side pillars instead.
- The minigun is comically easy to miss - pretty much every passage into that area is designed for you to look the other way or overshoot it jumping in there, and you might go through the next 80% of the map before having any reason to go back into that part of the room. Its platform has been moved to the lower area, with the upper platform being ostensibly a shotgun but the shotgun guy now standing there took it.

E3M4
- Stealth worm 16 stuck on overlapping ledge.
- Changed the St. Pete's cross into a sealed-off fireblu portal.
- Extended the stairs up from that long bridge so that it's more obvious from a distance at low resolution that those are legit stairs up and you're actually expected to backtrack that way.
- Redecorated the exit room.

E3M5
- Jump down into painlord courtyard in the west would be invisi-blocked by said painlords. Monster blocking line added to keep the way down clear. Moved the tree out of the way too, for good measure.
- Addressed #996.

E3M9
- Fixed numerous tutti-frutti textures and unstraightened some unintentional fake-contrasts in some organic areas.

E4M2
- Got rid of the flat bleed trick in that slime tunnel. It's just a big steady spotlight now - and it doesn't mitigate any nukage damage.
- The secret exit room multiplayer contingency entrance is now opened immediately after the trap-bars are triggered. The trap-bars themselves are no longer crushing, preventing a possible immobilization softlock.
- Said multi-contingency entrance area is remodeled to make it clear it has been jump-proofed.
- Aligned hex textures in secret exit room.

E4M9
- Flagged line 1765 as monster blocking. It's possible for a monster, while that platform is descending, to step right onto that window ledge and get stuck there as the platform continues its movement.
- Added glass to the sides of sector 230 to prevent the player from falling in under certain circumstances.

Map02
- Changed the multi-wallpart elevator so that the middle is the "rungs" that the side "pulleys" are lifting up the platform against.

Map04
- Numerous cosmetic fixes to the starting area.

Map05
- A couple lines on the edge of the stream used the surrounding brown rock instead of the zimmer brown.
- Made the helipad symmetrical and made it look more weathered, including adding a small tree where there would once have been a light.

Map11
- The walk line to open the yellow door switch is now south of the minigun, so those who "know" how it works (without knowing the exact location of the line) don't have to go all the way up to the door to intentionally trigger it.
- Removed some hard-only monsters from the eastern branch, removing some matribites and octaminators and replacing one of the former with a formerly easy-only trilobite and one of the latter with a stealth worm.
- The dock area is now opened up by the same switch that lowers the two platforms, which are now lowered permanently.
- The red skull shootable switches are now located so high up that it's impossible to gamble on a stray SSG shot hitting them from below.
- Addressed #996 in some small gaps in the final arena.

Map23
- Moved all health bonuses around the broken teleporter to look less like mushrooms.

Map26
- Blue skull door trim was misaligned, obscuring the shape of the horns.
- Zombieman 130 stuck inside a monster blocking line.
- Moved all health bonuses around the starting area corpse to look less like mushrooms.

Map29
- Addressed #996.

Map32
- Added a way to climb out of the lava pit.


mc776:
Looking at that I suspect the issue is less the size and more the shape - the current flash is pointing almost straight up.

Here's a crude attempt to foreshorten it without making anything bigger:
image

selliott512 (approved):
This has been around a while, so I'm going to assume it's not too controversial. I'll approve it.

I didn't verify each fix. I just verified that the levels listed in the description matches what was actually changed. I also loaded each level for a few seconds in GZDoom.

PR #1123, opened 2023-09-21
graphics: new END*.

mc776:
The transparency's broken on the current one and it doesn't look good anyway.


selliott512 (approved):
Looks good. I don't have strong feelings about the color or size, but the working transparency is nice, and the text feels more consistent with other text in the game.

I like it in the lower right hand corner, and the "...".

For those not familiar this is unique to Freedoom 1. It's displayed after defeating the large technospider at the end of E3M8. Here are some before and after screenshots:

Before this change:
to-be-continued-old

After this change:
to-be-continued-new

PR #1124, opened 2023-09-22
music: new tracks; shuffling.

mc776:
New track "Oxidine" by Korp (based loosely on Jute's previous E1M6 track) slotted for E1M6 replacing "Panic" by Ajanddino.

New track "Hidden Between Spades" by Korp slotted for E1M9.

"Derelict" by continuum.mid moved from E1M9 to E1M7.

"Sewers" by Discoholic and Blastfrog moved from E1M7 to Map12, replacing BobFromReboot's untitled track.

Rationale for posterity:
I really appreciate how Korp managed to take the old E1M6 track and rework it into something that slaps. The current track, while functional and not derivative in a way that would require its replacement, is still very reminiscent of Doom's rock/metal tracks in a way the rest of FD E1's music is not, which just makes the resemblances stick out even more.

"Derelict" has always only marginally worked for E1M9 (or rather it doesn't work, but still works better for it than the current E1M5). The dark, brooding ominous track works *much* better for the penultimate level, while the low-key funky riff of "Sewers" works much better for the cluster starter warming up after the intermission text than either "Derelict" or the current Map12 track.


mc776:
\> antepenultimate
\> can't count up to 2 backwards

Commenting as a reminder to fix that in the squash commit message.

selliott512 (approved):
Sounds good. I don't have strong feelings about music, so others should feel free to comment, but I'll approve since I think we want to get this in, and because I don't know of any controversy.

I did listen to each impacted map briefly. My only comment is that the new E1M6 "Oxidine" is loud compared to some of the other levels, but that isn't necessarily a problem.

PR #1127, opened 2023-09-25
manual/dehacked: overhaul.

### LINKED ISSUE #992: manual: common terms for things should be translated?

### LINKED ISSUE #1096: "Lore" item names make items less understandable
#1021 a few weeks back changed some of the item names to imply some lore. This I'm fine with in principle, but I'm concerned that the lore "gets in the way" in some cases and potentially makes it confusing to new players. This is especially the case for some of the most basic items which are the "bread and butter" of the game and the things that are first encountered by the player.

These are the ones I think need revising:

> GOTHTHBONUS = Harvested an ectoplasmic polyp.

This is for the health bonuses. It's not obvious what an ectoplasmic polyp is or does, without studying the status bar to try to figure out picking one up does. I think there should be an explicit reference to health in some way, for example calling it a "health polyp" or "healing polyp".

> GOTARMBONUS = Connected a polaric field emitter.

Same deal here. What is a polaric field and what does emitting one do? There ought to be some explicit reference to armor, for example we could call it "polaric armor" instead of "polaric field". Actually we 

> GOTSTIM = Applied a regenerative nanobot canister.

> GOTMEDINEED = Nano-surgeried some really sensitive parts!

> GOTMEDIKIT = Applied a regenerative nano-surgery kit.

I appreciate that these are trying to add some lore (that the health packs contain nanobots doing some kind of nano-surgery that restore your health) but again, I think it gets in the way. The current sprites look like medikits and not anything technological, I think we should dial back the lore.

mc776:
Continuation of #1091. Closes #1096 and #992, subject to better terms from an actual Spanish speaker.

Corresponding changes have been added to Spanish translation using machine translation. All reasonable efforts have been used to try to vet the output so it actually makes sense but it will need an actual Spanish-speaking reader; I've got an alternate version that *only* does the obvious fixes (but will not close #992) if this turns out to be irredeemably bad.

**Obvious Fixes**
- Converts a few stray CxMy references.
- Updates some screenshots that still showed old assets.
- Fixes a lot of miscopypasted alt text.
- Does not claim DBSG to be "twice" as powerful.
- Adds Handgun pickup sprite.
- Adds section for what happens when the player dies.

**Terminology**
- Addresses #1096 by restoring abstract references to health and armour.
- Adds distinction between "refill" and "boost" for health.
- Gives pickup messages a distinct verb for each class of item being picked up.

**Clarifications**
- Much more explicitly describes certain monster and weapon behaviour, prioritizing the exact game mechanic over what one might think it's trying to depict.
- Makes the deceased name appear as the first element in all obituaries, and uses the same term for the "same" weapon where a monster is using an equivalent to a player weapon.
- Emphasizes that the misconception is that Freedoom is *only* about providing a gratis IWAD, to better avoid the unintentional reading that it is a misconception that Freedoom is being provided gratis.

**Organization**
- Makes new sections for keys and the general introduction to how items work.
- Moves the colorblind section to right after the new key section.

**Aesthetics**
- Rewords some obituaries to better match the Freedoom monsters rather than the id ones.

**Putative Target Audience**
- Adds an explanation of what a source port is.
- Adds the "press Enter a bunch of times to start the game without reading anything" trick.
- Adds a warning about Chocolate Doom's savegame buffer overflow emulation.
- Adds a detailed tutorial walking the player through a live example of every type of ingame interaction.
- Rewrites all the controller help to reflect modern, post-WASD boomer shooter paradigm, doing away with the previous-century assumptions that you only had turn and forward/back and lateral movement was an unusual extra.


lumps/dehacked/dehacked.txt
```diff
@@ -156,43 +156,43 @@ STARTUP5========================================================================
...
-GOTSHOTGUN2 = You got the double-barrelled shotgun!
+GOTCLIP = Acquired a bullet clip.
+GOTCLIPBOX = Acquired a case of bullets.
+GOTROCKET = Acquired a missile.
+GOTROCKBOX = Acquired a crate of missiles.
+GOTCELL = Acquired a small energy recharge.
+GOTCELLBOX = Acquired a large energy recharge.
+GOTSHELLS = Acquired some shotgun shells.
+GOTSHELLBOX = Acquired a box of shotgun shells.
+GOTBACKPACK = A backpack for all your ammo storage needs!
```
Talon1024, line 188: How about "combat pack" instead of backpack?

mc776: Trying to look up that term on DuckDuckGo brings up
1. a lot of video game stuff that has nothing to do with carrying devices at all
2. some actual military gear that appears to be more not-backpacks than backpacks

I think it's best to leave it as it is for now (and we've still got "shotgun" anyway).

mc776: On further thought it might be a good idea to take a bigger look at whatever remaining stuff (shotgun, backpack, etc.) that uses the same terminology as id and see if they can or ought to be rewritten.

But as a setting/lore thing it does feel outside the scope of this already-sprawling PR though.

mc776: I can see how "combat pack" could be intended to make it clearer that this thing's only supposed to store ammo and not health or something.

But then that just makes imagine a "combat trauma pack" that lets you carry 4 medikits for rapid deployment at any time... and is the soulsphere replacement...

selliott512 (approved):
Looks great. I checked the following things:

1. There are only minor changes since the last PR having to do with "pistol" vs. "handgun" and some other very minor things.
2. The Spanish manual has the same structure and images (but I did not check the content) as the English manual *except* the Spanish manual is missing a section for the ammo at the bottom of page 20. As to translation GPT 4 says:

> In the context of a video game, the word "ammo" (referring to ammunition) can be translated to Spanish as "munición" or "municiones". If you're referring specifically to an ammunition pickup in a game, you might see it labeled as "munición" or "recoger municiones".

Certainly not a critical issue though.

selliott512:
I also gave the added lines a quick spell check via `spell` and didn't find anything. I like the dehacked changes.

mc776:
> The Spanish manual has the same structure and images (but I did not check the content) as the English manual except the Spanish manual is missing a section for the ammo at the bottom of page 20.

I have no idea where page 20 is. Can you copypaste a recognizable section of text from that area?

PR #1130, opened 2023-09-27
flats: new FLAT4 by @mc776

Talon1024:
This one is based on the Aquatex flats


mc776:
It was actually Korp who posted it, I just lightened the background a bit so the overall shade would be a little closer to the existing/id FLAT4.

mc776 (approved):

mc776:
I've approved it, but before it's merged can you see if you can run it through pngcrush or something? I just noticed the file is 420% the size of the original.

PR #1131, opened 2023-09-29
sprites: new health bonus.

mc776:
My edit of Magnolia's edit of Korp's original bottle sprite. I think this strikes the correct balance of sharing the same symbolism as the other non-Special health pickups while diegetically looking like something fundamentally different enough to justify the different behaviour.

couch


nful2:
Looks good!

CREDITS
```diff
@@ -991,3 +991,8 @@ N: Enzo Carozza
 S: uni
 E: dudetank4@gmail.com
 D: CREDITS-MUSIC
+
+S: Magnolia
+D: sprites
+
+
```
selliott512, line 998: There's some extra whitespace at the end of this file now. No big deal, and we'll probably have opportunities to address it before the release.

buildcfg.txt
```diff
@@ -2173,10 +2173,10 @@ BKEYB0	7	19
...
 BLUDB0	6	7
 BLUDC0	3	5
-BON1A0	5	21
-BON1B0	5	21
-BON1C0	6	21
-BON1D0	6	21
+BON1A0	5	15
+BON1B0	5	15
+BON1C0	5	15
+BON1D0	5	15
```
selliott512, line 2179: The image sizes have changed from 11x22 to 11x15. Since the image coordinates are zero based previously the offset, or foot of the image, was a pixel on the bottom row inside the image. Now it's just below the bottom of the image. Inside the image feels more correct to me, but I don't know that it matters since it seems to look fine the way it is with this change.

selliott512 (approved):
Looks good. I liked the dancing (mushroom?) pickups, but this feels more consistent with the game since there aren't other places that plants play a significant role that I can think of.

Incidentally, it seems odd to me that the least valuable health pickup has animation, but not the more valuable ones. But that's inherent to Doom, so it makes sense to continue that behavior.

buildcfg.txt
mc776: The idea of having the offset be a little bit up from the bottom of the sprite is to give the illusion of depth, like the foot of the thing is extending forwards a bit.

It starts looking a lot less consistent in sourceports that let you look up and down; in some hardware accelerated implementations (e.g., GZDoom) the bottom part just clips into the floor like it's sunken in.

This particular sprite isn't drawn with that perspective trick in the first place, though, so the offset being at the very bottom is correct.

mc776:
Notwithstanding/further to everything I've said about the previous sprite, I just realized the way the new ones look like little green dots in vanilla reminds me of something:
1110417-simant-amiga-beware-of-the-spider

PR #1133, opened 2023-09-29
manual: use "vanguard" for both languages.

mc776:
Turns out it's a perfectly functional word for this purpose in Spanish.

The English "device" double entendre, however, is apparently untranslatable. (Please correct me if I am wrong on this.)

### DIFF
--- a/manual/manual-es.adoc
+++ b/manual/manual-es.adoc
@@ -840,7 +840,7 @@ los enemigos y objetos.
 | **IDCLEVxy** | Empieza un nuevo juego (que reinicia todo) en ExMy (Phase 1) o MAPxy (Phase 2).
 | **IDMUSxy** | Cambia la música por la de ExMy (Phase 1) o MAPxy (Phase 2).
 | **IDCHOPPERS** | Te da una sierra de hender.
-| **IDBEHOLDV** | Te da la **V**ia Gladii.
+| **IDBEHOLDV** | Te da el artefacto de **v**anguardia.
 | **IDBEHOLDS** | Te da una **s**imbionte de fuerza.
 | **IDBEHOLDI** | Te da el **i**nvisibilizador.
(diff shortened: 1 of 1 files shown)


selliott512 (approved):
I'm not a Spanish speaker, but GPT 4 mostly liked the change, and actually had a lot to say about it, which I'll leave here for those interested.

GPT 4's opinion on this change:
----
The proposed changes in the Spanish version aim to translate the terminology from English to Spanish while retaining the meaning and context from the English version.

Translation of "Vanguard Device":

The initial translation as "Via Gladii" doesn't seem to correspond with the term "Vanguard Device". It seems like an entirely different phrase.
The proposed change to "Artefacto de vanguardia" is a more literal translation of "Vanguard Device" and helps retain the original meaning.
Translation of the phrase "Gives the vanguard device powerup":

The initial translation "Te da la Via Gladii" again doesn't correspond with the English phrase, and it references the incorrect initial translation.
The proposed change to "Te da el artefacto de vanguardia" is a more accurate translation of the original phrase and corresponds with the updated translation of "Vanguard Device".
However, there are a couple of nuances that could be considered:

The phrase "Artefacto de vanguardia" is a literal translation and does seem to convey the meaning of a device that is at the forefront or leading edge. However, without knowing the exact context or function of the Vanguard Device in the game, it's hard to ascertain if this is the most fitting translation. Sometimes, game-specific terms have unique translations that are not literal but are accepted and understood by the gaming community.

In the second proposed change, the word "vanguardia" is not capitalized in the second instance whereas "V" was capitalized in "Via Gladii". If the capitalization is significant (perhaps due to a formatting or coding requirement within the game), this might need to be adjusted to "Te da el artefacto de Vanguardia".

In summary, the proposed Spanish changes are reasonable as they attempt to provide a more accurate translation from the English version, but the exact appropriateness might depend on the specific context within the game and the gaming community's accepted terminology.
----

PR #1136, opened 2023-09-30
dehacked: improve messages.

mc776:
Just a little QOL thing after playing with the new dehacked update for a while and realizing some of the changes were actually kinda annoying.

The long latinate verbs in particular were getting to me. They are now "put on" for armour, "took" for health and "picked up" for ammo.

I think Map11 is the only map in both IWADs that distinguishes skeleton keys from passcards, so it's always the colour that's the most important distinction. There's no way in English to sound natural to put the adjective last, so it has to be in the beginning.

Similarly, "bullet clip" is now "some bullets" so the ammo type name is at the very end like all the other ammo pickup messages.

The 1% pickups are similarly rephrased so the words "health" and "armor" come first rather than second-last. If people find the exclamation points annoying for such a frequent message I can delete them but they feel right to me ingame, at least in chocolate where I can only see one line at a time.

The current Freedoom key messages are already a big improvement over id but they would still benefit from having the required colour stated at the forefront. The "object" language was always offputting and distracting (in game terms it usually makes me think of actors/things not map geometry) so I've omitted it; "for this door" can also be omitted in my view but there's probably some distinction that some map out there actually needs made.

The earnest "Never give up!" seems to be more in line of Freedoom's overall vibe than the more laconic, Doom-like "Not dead yet!".


mc776:
What if we deleted the words "small" from the health and the energy? That way the line of text would itself be smaller.

selliott512 (approved):
I like it. I also think that non-English speakers will appreciate the simplified language.

I verified that only the values were changed. There are still the same keys (to the left of `=`) in the same order. I ran it though spell check - no problems. I also verified that each key and value seems like they could go to together (there's not an obvious discrepancy where the key refers to "RED" and the value "blue", for example).

PR #1137, opened 2023-10-01
flats: replace SWATER with fireblu.

mc776:
This was briefly discussed on Discord a while ago and I'll throw it out here and see if it sticks.

Basically nobody has any use case for that choppy bright yellow and blue animation. It doesn't really look like anything, so it's either:

1. some hazardous undefined pollutant for a hurtfloor
2. some high-contrast dimensional-rift warp sort of deal

Between nukage, blood, brown sludge and lava - heck, even blue water is good for radioactive stuff - we've pretty much got every use case covered. I've actually never seen the current SWATER used this way.

A lot of messed-up dimensional-rift stuff is represented by a combination of hot rock flats and fireblu - in a lot of places where the hot rock flat is pretty obviously motivated by the lack of a fireblu flat.

This proposed change will allow that fireblu to be used on flat surfaces while breaking no actual use case for what it replaces.

If I am wrong about this please let me know which specific wad and level. (But please don't use Map26 as an example - this PR would actually fix that visual anomaly)


KorpKat (approved):

MerrittK (approved):
Great Idea. If people want the older version as a PWAD mapping resource, they can borrow from Aquatex

Xindage (approved):
While i approve that change as well, just keep this one to be applied after the next stable release.

mc776:
Can I ask why the wait though? I'm only aware of a single existing use of the current SWATER flat and the person responsible for adding it (me) would actually rather have this flat in that place.

Xindage:
well if you're sure of it and wanna proceed fell free to so.

PR #1138, opened 2023-10-02
sprites: make worm and tripod flippable.

mc776:
There is no reason whatsoever for the worm to have 8 rotations separately defined on any of its animations.

The tripod unlike the cyberdemon (or large technospider) is fully symmetrical as well.

I'd suggest doing this for the zombieman since no one's going to notice anyway, but I'm not the one who had put the work into those additional rotations.


selliott512 (approved):
Looks good. This should make the sprites easier to maintain. Here's what I did to check this change after reviewing [sprite naming conventions](https://zdoom.org/wiki/Sprite):

1. I looked at it in game, and the difference was not obvious to me.
2. I verified that the file renames did not introduce any changes to those files as well (sometimes it can be hard to tell with git).
3. I verified that redundant angles 6, 7 and 8 were replaced with flopped versions of angles 4, 3 and 2, and then deleted.
4. The difference between a flopped version of the deleted image, and the mirrored image is minimal.

Here's an example for step 4 for `CYBRF6`:

The deleted image flopped (about center Y-axis):
cybrf6-flop

The new mirrored image:
cybrf4f6

The difference of the above two images. The original image is faint / light with dark red pixels where differences are:
cybrf6-diff

As you can see the difference mostly has to do with his back, but it's not much, and probably not noticeable. This was the biggest difference I saw.

Here's how the above images were produced:

```
convert sprites/cybrf6.png -flop sprites/cybrf6-flop.png
compare sprites/cybrf6-flop.png sprites/cybrf4f6.png sprites/cybrf6-diff.png
```

I'm curious how this could be expanded. You mentioned zombieman, but maybe there could be an automated script that finds sprites that are close enough to mirror. Do you think such a script is worth writing?

mc776:
I don't think such a script is worth writing. Off the top of my head, of the current designs these are the only ones that unequivocally need this treatment - the zombie fires right-handed from its right hand and someone spent a lot of time and work making that consistent across the board for some reason, and the large technospider has that little flag and antenna that are each on only one side, so we are losing some existing detail if we flip those.

PR #1139, opened 2023-10-03
music: correct drum instrument on d_e3m1.

mc776:


selliott512 (approved):
It's a subtle difference to me, but good.

PR #1143, opened 2023-10-07
patches: remove bright spots on edge of gargoyle.

mc776:


Big-the-Fat (approved):
Yes.  A good tweak.  Small things make a big difference.

selliott512 (approved):
Looks good.

If you feel like making more changes IMHO the grey pixels around the perimeter are suboptimal on a non-grey surface.

mc776:
> If you feel like making more changes IMHO the grey pixels around the perimeter are suboptimal on a non-grey surface.

Do any textures actually do that right now? For me the difference very easily gets lost in the wood grain but then again so did these green dots.

Xindage:
> > If you feel like making more changes IMHO the grey pixels around the perimeter are suboptimal on a non-grey surface.
> 
> Do any textures actually do that right now? For me the difference very easily gets lost in the wood grain but then again so did these green dots.

maybe this one? still this texture is yet not used in any freedoom map 
image

mc776:
tbh i have no idea why that texture exists. Anyway there's no way to get that to blend right without making the existing "dark mode" textures look at least as bad if not worse.

If it looks okay on the wood I think it's ready to merge.

PR #1145, opened 2023-10-08
sprites: reshade candelabra.

mc776:
The way the part of the sprite right under the flames was evenly lit made it feel like something other than a light source.

I tried making the top of the central column lit on either side but it didn't look right at all so it's a third flame now.


selliott512 (approved):
I like it. It's more clear to me what the center flame is than than what it replaced.

PR #1147, opened 2023-10-10
sounds: new DSOOF

Talon1024:
The sound was originally recorded by me, and edited by @mc776 and TrashMarine on Discord.


mc776:
I'm messing around with this and I'm starting to think it would work better if it were around a third, possibly even a quarter as loud - a lot of ground surfaces are supposed to be softer or less sound-conducting.

Talon1024:
I feel like reverting the commit I just made, because I tried the quieter version of this sound ingame, and most of the time, I can barely hear it. But the equivalent sounds in other boomer-shooters are clearly audible.

mc776 (approved):

PR #1148, opened 2023-10-15
graphics: tidy up help screen.

mc776:
before:
help_old

after (old):
help-new2

after (new):

### DIFF
--- a/graphics/text/Makefile
+++ b/graphics/text/Makefile
@@ -99,7 +99,7 @@ help.png: text_strings helpbg.png
 	   172,115 "Armor"                           \
 	   280,110 "file:../../sprites/arm2b0.png"   \
 	   242,107 "file:../../sprites/arm1b0.png"   \
-	   226,114 "file:../../sprites/bon2a0.png"   \
+	   226,114 "file:../../sprites/bon2b0.png"   \
 	   145,139 "Map"                             \
 	   175,127 "file:../../sprites/pmapa0.png"   \
 	   10,139  "Big Boost"                       \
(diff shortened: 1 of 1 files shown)


graphics/text/Makefile
```diff
@@ -70,55 +70,55 @@ help.png: text_strings helpbg.png
...
-	   68,62   "file:../../sprites/ammoa0.png"   \
+	   20,38   "file:../../sprites/csawa0.png"   \
+	   90,23   "file:../../sprites/shota0.png"   \
+	   90,38   "file:../../sprites/sgn2a0.png"   \
+	   155,18  "file:../../sprites/mguna0.png"   \
+	   155,38  "file:../../sprites/launa0.png"   \
+	   235,23  "file:../../sprites/plasa0.png"   \
+	   235,38  "file:../../sprites/bfuga0.png"   \
+	   10,67   "Bullets"                         \
+	   77,62   "file:../../sprites/ammoa0.png"   \
```
selliott512, line 81: It seems sub-optimal having a this in a makefile with all the `\`s and whatnot. Like there could be a separate tab separated file with the coordinates followed by the text or file.

I know you didn't do it, so no need to change anything. Just my two cents on the way it currently works.

selliott512 (approved):
Looks good. However, IMHO, a tiny bit of separate between the small ammo pickup (the shells, for example) and the large ammo pickup (box of shells) might be nice just to make it clear that it is two separate pickups. People will probably figure it out though.

graphics/text/Makefile
mc776: Visually on my text editor it's fine, but if there's any way to do it that, say, lets you comment out one arbitrary line in the block that would be good...

mc776:
> However, IMHO, a tiny bit of separate between the small ammo pickup (the shells, for example) and the large ammo pickup (box of shells) might be nice just to make it clear that it is two separate pickups. People will probably figure it out though.

I like conceptual groupings myself but I know a lot of people don't, anyway it turns out there was enough whitespace to go without and I just needed to shift a few other elements.

I'll squash and merge later today if nobody else brings up anything in the meantime.

PR #1149, opened 2023-10-15
patches: new brick patches by @KorpKat

Talon1024:


mc776:
I do not see the purpose of this. This isn't an unquestionable improvement like the greys.

Many of them substantially change the apparent brightness or the aesthetic of the texture, drastically altering the look of maps in ways the mappers never anticipated.

Numerous elements do not tile the same way or are even in the same place as the current.

The change to the cracks in RW5_1 is a flat-out downgrade, trading some cool decayed-looking damage for what look like outlines of cartoon sound effect bubbles.

Every reason Xindage had for not approving #1057 seems to go double for this PR.

The new window looks nice but even there that new white edge (again all these right angles and clean lines *are not antialiased in vanilla*) leaves me concerned about how that would look on the existing maps.

I won't close this now in case another maintainer really wants to add these for other reasons but this may need some reconsideration. At minimum this should not be merged before the next release.

KorpKat:
Most of the current brick set looks weird on PWADS, i'm not worried about how itll look in our maps, but on other ones using the original Doom tileset. Textures BRICK10 through 12 are edits of BRICK1, which the originals clearly arent. 
This set is meant to be more faithful to the original doom alignments, because as i previously mentioned: We can fix our own maps but not community made ones.

MerrittK:
Just to re-iterate, a bunch of the textures that have been summited recently have horrible aliasing artifacts. They look fine as PNGs when seen in the discord, but moving around in a software render port breaks it.

We "could" replace our own texture usage to match, but I don't endorse replacing the current graphics right now, especially when there are textures that need replacement more urgently.

mc776:
Seconding what Merritt said.

As for being closer to Doom, I understand the concern (and now understand why rw5_1 looks like that, and as I've said about other things before *we are not obligated to replicate id's mistakes*) but this has us going dangerously back into "derivative work" territory the more "faithful" we try to be. The change on the lamp in rw12_3 is especially blatant in this respect.

For reasons that are rather fundamental to the project - and which go to the BIGBRIK textures that #1057 was supposed to replace which were my very first contribution to Freedoom - I'd be more comfortable with alignments being slightly off in PWADs not made for Freedoom than perfect interchangeability.

All that aside, the green and red bricks are a significant departure from the overall brightness of the originals, and I suspect (given previous proposed and committed replacements in this style) a lot of the other bricks will end up revealing the flatness and sharp edges of Doom's limited map geometry in ways that the current textures had been carefully shaded to avoid.

fragglet:
Closing for now; my take on this is not a hard "no":, but "not now", based on @Xindage's previous comments on #1057 and the fact we have been talking recently about putting out a new release. But I agree with others that these textures seem like an arbitrary change and there isn't any obvious improvement over the old ones. If it's going to disrupt existing levels then it doesn't seem to be worth the change.

PR #1155, opened 2023-10-21
sprites: touch up shotguns.

mc776:
PASG
- Palettization left some weirdly brightly saturated green pixels on highlights. Changed them to olive manually.
- Embiggened the arm slightly to better match the fist.
- Darkened the arm slightly on the A frame to better match the mugshot, etc..
- Made the flash a bit brighter in the inside and less spiky at the top, to minimize the impression of a single blotch of red.

DBSG
- Added more contrast to the hand and shells.
- Cleaned up some stray pixels on the right edge.
- Mad the flash shorter vertically and made the peaks a little inward following the path of the guns instead of shooting straight up from just in front of the muzzle.

J1
J2


selliott512 (approved):
Looks good. Most of the changes are subtle to me, but I like the less spiky muzzle flash - it makes more sense to me.

PR #1156, opened 2023-10-21
graphics: offset M_PAUSE X by -8.

### LINKED ISSUE #1153: M_PAUSE off center
Doom's source code has a hard coded width value for the pause graphic of 68 pixels. Freedoom's is only 52 pixels wide.

mc776:
Closes #1153.


selliott512:
Would you mind clarifying where this is? In PrBoom+ the keybinding setup menu looks like this before and after this change for me:

pause-menu

unithecutie:
image
Top of the screen when you use said function.

selliott512:
Thanks for the explanation. I don't normally use that feature, but still seems like I should've tried it in hindsight.

That being said it seems like this change makes the "PAUSE" less centered for PrBoom+. Here are before and after screenshots imported into The Gimp with vertical guidelines added, demonstrating that it was centered around the pistol before, but not after:

before:
pause-before-guidelines

after:
pause-after-guidelines

Averaging the guideline coordinates gives the center of the word in both cases:
```
  before: (844 + 1076) / 2 = 960
  after : (880 + 1114) / 2 = 997
```
The before value of 960 is exactly the center of my 1920 width monitor. Am I missing something? Do different engines behave differently?

mc776:
So PRBoom+ dynamically centers the pause graphic unlike vanilla?

mc776:
I just gave it a try on Crispy. It behaves the same as vanilla.

If we have to fix this with a larger graphic I'd like to see something clearly visually distinct from the regular fonts, just in case that stuff gets updated...

mc776:
I'll post a new PR with fawtytoo's spaced lettering.

selliott512:
I don't know if it is helpful, but the "PAUSE" is centered before and after this now closed PR in GZDoom.

Issue #1158, opened 2023-10-24
Jaws of Defeat textures contain copyrighted material


fragglet:
They contain parts of a Hexen metal texture. Thanks to Talon1024 on Discord for noticing this and PGE for the following illustrative GIFs.

![JoD_Hx1](https://github.com/freedoom/freedoom/assets/96821/680f72e2-3b76-47aa-9b40-184565e8f767)
![JoD_Hx2](https://github.com/freedoom/freedoom/assets/96821/be41b33d-d0b8-4161-b4f9-803232b8f2b6)

Texture patch filenames are:

* rwdmon1.png (small parts)
* rwdmon4.png
* rwdmon5.png
* rwdmon7.png
* rwdmon8.png
* rwdmon9.png
* rwdmon10.png

fragglet:
Obviously we need to replace these textures but I'm not sure how much work is going to be involved. If someone doesn't step up to provide immediate replacements then I recommend we either replace them with blank textures or copy a different similar texture.

mc776:
Has anyone volunteered to do new ones yet?

fragglet:
Putting together a PR now.

mc776:
Should BOSSBACK be replaced? I can't quite see it but it does seem to have a bit of that original panelling: https://github.com/freedoom/freedoom/blob/master/graphics/bossback.png

PR #1164, opened 2023-10-28
textures: New boss wall texture set

beau-g:
With the situation of #https://github.com/freedoom/freedoom/issues/1158, I've created a set of replacement textures here for the boss wall inspired by the old texture (using Fragglet's replacement as a base). 
The metal paneling here is recreated without any of the paneling from the original graphic, with some other parts of the textures using elements of the patches "GRATE", "RW33_2", "W109_1", "W109_2", and "W110_1", alongside "w17_1" from the attic.
In addition it provides some extra PWAD compatibility over the original texture, with "monitor" graphics provided in certain locations for PWADs expecting their placement.


mc776:
Personally I actually honestly liked fraggle's rush-job no-panel fix over having any panels at all, but if people do prefer this than I'm OK with it too...

mc776:
DOOM00
Scratch what I said before, this looks way more fitting than either the infringing or the quickfix.

mc776 (approved):

PR #1165, opened 2023-10-28
patches/flats: revised brick textures by @KorpKat

Talon1024:
This is a newer version of #1149. It also contains a new FLAT8, which fixes a copyright issue noticed by @KorpKat on the Discord server.


mc776:
I'm just going to close this given #1149 is also closed.

mc776:
What was the issue with FLAT8? If that can be PR'd on its own that could be merged right away.

Cescade:
Very identical texture edited minorly, also fixes current alignment.

mc776:
flat8-comparison
I'm not quite convinced that the current FLAT8 is a copy looking at the texture of the bricks, and "fixing" the alignment verges on undoing the fixing of the bigger problem, but it does seem like an overall improvement anyway... which maps even use this flat anyway?

Also fwiw here's a similar comparison for FLOOR5_4 which does not appear to have this problem and I think should be left as it currently is:
floor5_4-comparison

mc776:
A side note while looking at some of these flats: A lot of the proposed flats in this PR have lighting highlights on the top left of the image, which doesn't really work for "lighting from above" when they're flats. The current FLOOR5_4 introduces this problem where the id doesn't, but the proposed totally different replacement makes the problem much worse.

If we're trying to line things up with fake contrast https://doomwiki.org/wiki/Fake_contrast though then the highlights should be on the sides or something, but even then I'm not sure that would be a good idea for something that tiles.

Cescade:
> I'm not quite convinced that the current FLAT8 is a copy looking at the texture of the bricks,
It's mostly similar texture with a brighter color and aligned downward.

> "fixing" the alignment verges on undoing the fixing of the bigger problem,
There can be more bigger problems loading on different non-freedoom maps with just improper alignment (that uses doom texture alignment). Map c1m6 uses this flat (outside pole area) and uses the current allignment, could be simply solved enough with minor edit.

SarahMagnolia:
845p5s
After aligning the brick patterns so they match up I made an animated gif of the two FLAT8s.

As you can see it's a lot more blatant.

PR #1166, opened 2023-10-28
patches: new TEKGREN textures by @KorpKat

Talon1024:


mc776:
My comments on #1149 actually had these in mind as well.

lumps/textures/textures.cfg
```diff
@@ -1406,6 +1406,16 @@ TEKGREN4        64      128
...
 *   RW39_1      0       0
+TEKGREN6        64      128
+*   RW37_5      0       0
+TEKGRENB        64      128 ; Big
+*   RW37_6      0       0
+TEKGRENT        64      128 ; Tall
+*   RW37_7      0       0
+TEKGRENW        64      128 ; Wide
+*   RW37_8      0       0
+TEKGRENS        64      128 ; Square
```
selliott512, line 1417: I don't understand these comments since they are all 64 x 128, but that's ok.

selliott512 (dismissed):
Looks good.

I have a soft spot for bright clean textures in video games, which this seems to be. I like the little light, for example. But I know that may be a topic of some controversy. Some may prefer dark and gloomy textures.

lumps/textures/textures.cfg
Talon1024: Those comments describe the shape of the panels on those patches. Also, I had to use something other than a number for the suffix because `TEKGREN10` doesn't fit into 8 characters.

Talon1024:
Before this gets merged, I recently tried this out ingame, and it seems to be... Blander than the previous TEKGREN set.
Screenshot_Doom_20231029_020836
Screenshot_Doom_20231029_021259

mc776:
> Before this gets merged, I recently tried this out ingame, and it seems to be... Blander than the previous TEKGREN set. Screenshot_Doom_20231029_020836 Screenshot_Doom_20231029_021259

I agree, and despite Steven's approval I do not intend to be the person who pushes the green button on this.

mc776:
I'm also against the idea of turning this set of textures into a mere "METAL2 light mode" rather than leaving it to be its own thing - it just makes the texture set as a whole less varied, to no apparent advantage.

mc776:
And on a purely aesthetic level the change to RW37_4 is a pure downgrade - poorer balance in shapes and positive/negative space, increased contrast clashes and makes the pipes feel more pasted-on than the old ones.

mc776:
pipe1
pipe2
blue1
blue2
4a1
4a2
4b1
4b2

It just feels like going from Quake to Chex Quest for me...

mc776:
As with the METAL2 replacements, the resulting textures are *much* less visually forgiving of any deviation from the standard texturer-intended alignment, making anything look worse that might have played with any of the looser tolerances of the original, like with that curve in Map04.

This implicitly pushes people towards sticking with square angles and wall lengths of multiples of 32 or 64, removing much of the discretion for unusual shapes that was one of the really big things that Doom improved over its predecessors.

**This change is detrimental to mappers.**

As I've already got my own fork running I can of course just defer to majority opinion if everyone else thinks this change is awesome and great, but again I think this makes Freedoom worse.

Big-the-Fat:
Agreed.  In isolation the newer textures look fine but in-game they are stark and as shown above less flexible.

PR #1172, opened 2023-11-03
New pistol offsets by @KorpKat

Talon1024:


mc776:
No.

mc776:
This was discussed at length on Discord.

No one who has any idea what they're doing holds a pistol like that.

No one with a functioning hand who is trying to aim at all jerks their **entire arm** in anticipation of pulling the trigger like that.

Doom's implementation was idiotic even for its time and we have no obligation to slavishly follow it.

PR #1174, opened 2023-11-05
levels: Small fixes, mostly for Eureka

selliott512:
```
freedoom1.wad
  E2M1   Fix non-visible missing texture after exit with GRAY1
  E2M4   Fix missing texture before start with GRAY1
  E2M7   Fix missing upper texture for linedef #4431

freedoom2.wad
  MAP05  Removed unused tag #16 in monster closet on sector #137
         Made upper case incorrectly lower case texture "fwater1"
  MAP12  Removed non-visible Medusa texture on linedef #7219
  MAP24  Fix missing texture on linedef #1636 in monster closet
  MAP25  removed unused tag #31 and type on linedef #2741
  MAP31  removed unused tag #9 from sectors #17 and #71
```

My usual cleanup. There are a few areas of concern having to do with `tag` in the above. I removed unused tags, which shouldn't change the behavior since they were unused, but I'm concerned something may be wrong, particularly with MAP31.


Xindage:
"Fix missing texture before start with GRAY1" this is going to be a default error when i do mapping, i dont apply textures on 0 height sectors, that's why it report errors (nobody can see a line who dont have any height.

Xindage (approved):
it does not conflict with the newly incoming demos so i'm ok with this being applyed.

selliott512:
Thanks Xindage. `GRAY1` is indeed the default texture for Eureka. I figure it's a good default for things the player will never see. It signals that it is something like an error fix that can be revisited.

PR #1176, opened 2023-11-05
sprites: new tech lamps by Cascade

Talon1024:


mc776:
Whatever my thoughts about CANDA0, these ones are pretty clearly an improvement over the current. (Someone will need to PR an update to GZDoom's light definition files after this though)

That said, it would be great if the light glow animation could affect the shading on the body of the lamps themselves - the current short lamp has that though the current tall lamp does not.

mc776:
DOOM02
DOOM01
DOOM00

Since the entire light appears fullbright and the brightness of the actual light source portion visibly changes I think it would be nice to have the bottom of each column get a bit darker (and that brown reflection bit on the right could be darker in any event), but if another maintainer thinks it's enough of an improvement over the existing sprites to go as-is I have no objections to merging it now.

mc776:
One possible variant on the tall lamp. I've gotten rid of the side shaft panels because they tended to detract from the "lamp" aspect of it, and it now has four prongs below the light source so it's radially rather than bilaterally symmetrical.
DOOM00
tlmp-spiral

mc776:
And the other one.
DOOM00
tlmp2

Cescade:
Not too fond of what you did to the tall lamp but could you smoothen the transition of the fade of the light on the bronze "screw" pipe on the bottom?

mc776:
The animation or along the length?

mc776:
How's this?
tlmp
DOOM00

Cescade:
Yeah

mc776:
Sweet.

@Talon1024 if you can update this and #1177 I think both will be ready to merge.
tlmpa0
tlmpb0
tlmpc0
tlmpd0
offset should be (9,79)

mc776 (approved):

PR #1178, opened 2023-11-05
sprites: new gore poles and alien eggs by Cascade

Talon1024:


mc776 (approved):

buildcfg.txt
```diff
@@ -2456,10 +2456,10 @@ PLSGB0	-50	-100
 PLSSA0	9	12
 PLSSB0	9	12
 POL1A0	22	62
-POL2A0	19	62
```
mc776, line 2459: The Y offsets are all at the very bottom of the image even though the sprite is drawn with a vertically extending pool of gore. To look best in vanilla the offset should be moved up by 5:

DOOM03
DOOM02

Talon1024: So the Y offset should be 67 or 57?

mc776: The "raised offset" image has 43, 53 would have the sprite unequivocally hovering in the air.

Just one warning, though, these do get clipped into the ground in GZDoom so the bottom part is invisible. It's a tradeoff but personally I think it looks less worse clipped in GZDoom than hovering in vanilla or any sourceport:

Talon1024: I just found out that GZDoom has a new "Smarter" option for "Adjust sprite clipping" while I was testing that out myself.

PR #1179, opened 2023-11-05
sprites: new keen/vat baby by Cascade

Talon1024:


mc776:
I really like the way the new tube animation looks but I'm against the redesign - the little fat lizard baby works for what it is.

Alternatively, if everyone wants to replace it with a dead fetus, it should go more "splat" than it currently does.

PR #1180, opened 2023-11-05
sprites: new tech pillar by Cascade

Talon1024:


mc776:
Oof.

This reminds me of an earlier version of the current version of the techpillar where I'd added a lot of blue trim and lights - it really clashed with existing stuff.

The huge blue tube is not going to play well with other stuff.

I'm tempted to close this right now but let's see what other people think.

mc776:
That long thin highlight causes some awful flickering in and out as you move in vanilla, something I spent quite some time trying to get rid of when doing the current techpillar:
DOOM00
DOOM01

mc776:
Aesthetically it's just really too much, too thick, too monster-view-blocking for where we often see this decoration in FD. (The absolute image width is closer to id, but the main body is noticeably thicker - 32 this one, 20 current, 28 id.)
DOOM04
DOOM03
DOOM02

Even if the other issue were dealt with, this may require adjustments to existing maps to look right, and if we're at the "final bugfixes and recording demos" stage we should at least hold off on this until after release.

mc776:
TIL what Wolf's/Soda's techpillar's predecessor looked like: https://github.com/freedoom/freedoom/commit/610b5e2bafc7c06e44586bc4a8be83a325dff606

Was this new one based on that?

mc776:
Just throwing this brainstorm out there...
eleca0-concepts

mc776:
A demonstration of how much of your view gets blocked by a full-size column versus only using the current column for the middle:
DOOM00
DOOM01

And here's how colourful it looks using the old blue greeble and the current monitor, but not the current antenna/radiator bits:
DOOM02

Cescade:
I'll try some of those brainstorm ideas later.

Big-the-Fat:
I feel the new pillar's components are just a bit too 'in your face'.  They take up a little too much space are a bit too colourful and they draw too much attention.  I feel like the current one works better as a decoration - it adds interest to the map without being such a strong statement.  

I feel that the current one could perhaps stand another pass or two with some shading but all in all it's a good prop.

mc776:
I think my vote is for the third one, dark grey with blue trim going up.

Here's a comparison between that one and the pure dark grey:
DOOM01
DOOM00

Although I think the sprite is a lot stronger without that glare on the upper left of the screen:
DOOM03
eleca0

mc776:
Looking at this again I'm starting to like the pure dark grey variant better. I can go either way with this so if anyone has any strong opinions one way or the other please speak up.

Cescade:
I only did the blue with black segments based off those ideas you posted. The dark grey goes pretty well with the tech textures its mostly associated with.

Cescade:
Updated.
SET

mc776:
I definitely like the previous version better - the design was already very top-heavy and this just feels like it's a giant mace instead of a pillar.

The less solid feel of the base being split up like this (which also makes the top brighter in comparison in addition to the top actually being brighter now) adds to that unsteady feeling.

It should be noted that the player is typically seeing this from (well) below the sprite's center line, so the top will seem a bit bigger even if it were drawn exactly the same size as the bottom.

boogiebogus:
Seconded. I think that proportionally, the last version was better--my vote is for the pure grey variant.

Cescade:
Simpler top part design.
SET3

Cescade:
> It should be noted that the player is typically seeing this from (well) below the sprite's center line, so the top will seem a bit bigger even if it were drawn exactly the same size as the bottom.

I don't understand
tech

mc776:
Disregard that previous comment, I just gave this a try in GZDoom's X/Y billboarding mode and the difference isn't even enough to be measurable on my resolution.

The top is definitely better now, though I'm still partial to the previous base with its fewer flats, more "going up"-ness and being more evocative of a couple elements from the old techpillar.

Cescade:
I've no idea which "flats" you're talking about, "going up-ness" aswell, the top part is supposed to connect at whatever ceiling it is connected to (as intented from previous iterations)

mc776:
going up: sense of moving up the column. lines

flats: big areas of single colour

There are a lot of areas where the column is used where it's clearly not connected to the ceiling which is either much higher or is the sky; the id techpillar doesn't have anything on it necessarily implying it's connected to teh ceiling, it's just a flat disc.

Big-the-Fat:
I think the 'flats' mc776 is referring too are the panels on the bottom which make the pillar feel broader than it is and takes away from the verticality of the sprite which in turn make the top feel even bigger.  It's like the opposite effect of a weapon sprite that doesn't have a good perspective and fore-shortening.  Both effects have the same result:  They don't 'fool' the eye in the way they're supposed to.  Thus the weapon looks like it's pointing straight up the screen and your sprite here looks more truncated than it should.  

I hope that makes sense.  It's really difficult descirbing these things without it all becoming word salad.

Here's another idea following on from what I posted above.  Yes, it's narrower than your sprite but that's not the only thing that makes it feel tall.

ELECA0

Cescade:
Last post before I actually stop doing more.
SET4

mc776:
I think I like the second from the right best...

PR #1186, opened 2023-11-07
patches: new CEMENT set by @KorpKat

Talon1024:
These look so much better than both id Software's CEMENT set and Freedoom's previous CEMENT set

Screenshot_Doom_20231107_014611


Talon1024:
In case you were wondering why the warning stripes are so much wider on the new version of RW11_3 than the previous version, they were like that on the id version of that texture as well.

mc776:
The only thing I said in the other closed patch PRs that *might* not apply here as well is the whole "looking flat" thing, which I can't really discern just looking at that screenshot.

Everything feels worse proportioned than before - big empty spaces with little bits of stuff plopped on top of them. The dirt on the original textures was on there to help mitigate that and create more visually balanced images. There does not seem to be an appreciation for that in these proposed patches.

Why the constant insistence on reusing that one bit of techwall? It actually hurts suspension of disbelief even more than having everything look slightly different.

All the warning stripe sections are shaded so they look recessed into the wall (where previously they appear painted on or part of a panel that's added) and there's no apparent reason why.

Some parts of the walls look more worn and some less but there's no apparent reason for the change.

I'm seriously concerned with all these seemingly totally needless submissions that this is turning into a mass erasure of previous contributions by the existing FD community - I mean, if we're really all aiming to be a constantly moving Ship of Theseus and everything needs to be replaced every few years I guess that's up to whoever's in charge, but that constant change is not doing any mappers any favours.

PerdGateEnjoyer:
Disagreed. These jive better with Freedoom's existing stuff and don't look like they're stained with decades of piss.

What do you mean "Not visually balanced"? If anything, the originals weren't balanced (incredibly noisy and jagged).

"That one bit of techwall" makes it sound like it's in the exact same position every time (it's not, took me a good 3 tries to find the similarities)

The hazard striped do look a little funky, but I wouldn't say "recessed".

SarahMagnolia:
The intent with these textures is to move FreeDoom closer to having a consistent art style, which is something it's never had. As the project moves further and further towards a brighter, cell shaded aesthetic, these bitcrushed, grainy, sharp imitations of ID textures are going to stick out like a sore thumb. They look like they came from modern Fallout while the monsters and props look like a cartoon. 

The majority of the community is in agreeance that Korps Kat's textures are good. The only detractors from what I can see have given good faith criticism that the textures need better contrast and depth so they don't distort with distance too much, but that's about it. Nobody is as viscerally opposed to his textures as you. 

Things get replaced in FreeDoom all the time. Props, sounds, music, sprites, entire maps. Textures should be fair game. I don't frankly buy the concern for how this supposedly effects mappers because you just approved the lion's share of Cescade's new decorates. Decorations play just as vital a role in establishing a map's theme and setting as textures do, and most maps are pushing 20 anyway. More polished retakes of old textures don't ruin maps. It would be one thing if they were radically different textures altogether but they're not. 

It's time to set aside your own personal feelings on this and actually allow a discussion instead of rejecting it within hours before anyone else has a chance to weigh in just because it breaks your personal "suspension of disbelief".

mc776:
Then someone else can reopen and merge this.

Talon1024:
The thing is, the only people who are allowed to merge stuff are @mc776 and @fragglet. Anyone else with maintainer privileges is just not active enough to review or make any meaningful changes to Freedoom.

That said, this new CEMENT set is, IMHO, a definite improvement over the current CEMENT set, and maybe even Id's CEMENT set. The current CEMENT set looks blurry, unappealing, bland, and ugly in general, and this new CEMENT set is an improvement in that regard; it's more crisp and detailed.

Xindage:
> The thing is, the only people who are allowed to merge stuff are @mc776 and @fragglet. Anyone else with maintainer privileges is just not active enough to review or make any meaningful changes to Freedoom.
Like i said in discord i'll tell here, i dont like to work with its not my area, i dont know how to judge textures, musics or art in general so i prefere to stay away from that.

fragglet:
Things have obviously come to a head here and I need to step in. I'll start by addressing one particular comment:

> I think the fact that this was closed is proof that it isn't so much a democracy as it is a dictatorship

Let me just say outright, in case anyone is unclear about this:

👉 **Freedoom is not a democracy**

The project is run by consensus decision making, which is the method used in most open source projects. For anyone who doesn't know what this means, I recommend looking at the guidelines for Wikipedia, a different project that isn't so different to our own:

* [Not a democracy](https://en.wikipedia.org/wiki/Wikipedia:What_Wikipedia_is_not#Wikipedia_is_not_a_democracy)
* [Wikipedia:Consensus](https://en.wikipedia.org/wiki/Wikipedia:Consensus)

With that out of the way, I'll address some of the points here:

* It is inappropriate to immediately and abruptly close the PR before others have had the chance to comment. In the future, can we please give others the chance to comment? We can't form consensus if people aren't given a chance to comment.
* It is out of line to accuse another contributor of being a "dictator". Please assume good faith. I've seen more examples of this on the Discord and it needs to stop.
* It's also inappropriate to describe other contributors' work using phrases like "stained with decades of piss". 
* Freedoom has always revised its textures and other resources over the years and there is no reason why we have to stick to existing versions of textures. Multiple people here have expressed that they believe these textures are improvements over the existing versions.

Going forward I am going to insist on the following:
1. Insults and personal attacks are unacceptable and will cease on both sides. People who cannot follow this direction will be banned from the project.
2. Insulting language will not be used to describe people's work. This includes existing material in the project, even if it has been used for a long time, and even if you do not know who the author is.
3. Contributors will assume good faith of others. If you think someone is acting in bad faith, escalate privately to me, do not accuse people publicly.
4. Reviews will be unbiased and new work will be judged on its merits. If you have strong opinions on the author that you cannot put aside, recuse yourself and leave the review to others.
5. No more proxy submissions. Korp will create an account and submit their work themselves. @Talon1024 please assist Korp to do this if they are unable to figure it out for themselves.
6. Changes to visual resources like textures will include multiple screenshots to illustrate the new work. Ideally these should be "before and after" shots to illustrate the improvement. It is not possible to review new work just by looking at the images in the Github diff. WADs are built automatically for every PR. [See here](https://github.com/freedoom/freedoom/assets/96821/6ce68e88-cbad-4603-9b67-8b62956d04df) for how you can find them for download.
8. Reviews should be as objective as possible, although I appreciate that with artwork this can be difficult. If possible, try to use screenshots to illustrate your point.

PerdGateEnjoyer:
*note: I wasn't describing the current textures when I said "...decades of piss", I was describing the original iD textures.

fragglet:
Finally - to address the PR itself: @Xindage made a good argument in #1057 for not changing those textures on the grounds that he had put in considerable recent work to tweaking maps to fit the current textures. Since we are currently working on putting a release out, it is reasonable to defer such PRs until after the release. To this end, I have created a new tag for PRs named "deferred", that we can use in this situation.

This does not mean that such changes will never be accepted, it just means "not now". Again - we have always changed and revised textures as part of this project. But there is no reason that all PRs have to be immediately merged. It's also a reasonable criticism at any time if a texture changes "too much" from the previous version, if it's going to produce a lot of unnecessary work.

@Xindage although you do not consider textures "your area", texture changes do affect maps, so your opinion about whether a change is likely to be disruptive is greatly valued.

> *note: I wasn't describing the current textures when I said "...decades of piss", I was describing the original iD textures.

Thank you for clarifying.

Donowa93:
now that the new release has been made, could it be possible to review these again? a lot of people agree that these are improvements over the current ones, they don't have the same contrast issues as korp's other textures, they are more compatible with pwad maps (since they have the same layout as the id textures), and the alignment problems in current FD maps could be sidestepped by adding B textures (like how the wolf textures have alternate versions for FD's base maps)

aside from the re-alignments in FD's maps for 0.13, it doesn't seem like there's any objective reason as to why these were rejected, so a second opinion after the fact would be nice

KukoBerry:
I will say, while the textures look really good, they DO look a lil' flat in the screenshot - maybe the shadows for the tube-y parts of the cement and the duvets between each "slice" could be a smidge more pronounced...

PR #1188, opened 2023-11-08
sprites: adjust TLP2 lighting.

mc776:
The original attempt didn't make it into #1176 but it had issues anyway.
DOOM00


Talon1024:
> The original attempt didn't make it into #1176 but it had issues anyway.

You didn't post the D frame, so I didn't add the TLP2 sprites.

selliott512 (approved):
Looks good. I see the Illumination of the shaft varies with the light now instead of being static. Seems more realistic. It's a neat effect in a dark room.

PR #1193, opened 2023-11-11
Fist animations tweak and two handed pistol

Anonysussy:
Tweaking the fist animations and gave the pistol sprite an extra hand


mc776 (approved):

mc776:
Just gave it a try and I was *not* expecting this to be as much of an improvement as it was. Stuff just feels so much more snappy and solid with this.

I thought the left fist in the B frame would have that awful "blinking in and out of existence" effect but it's low enough not to be a problem at all.

mc776:
Btw feel free to add yourself to the CREDITS file for spriting work... I think the format should be pretty self-explanatory at this point

buildcfg.txt
```diff
@@ -2186,12 +2186,12 @@ PINSA0	15	39
...
 PINSD0	15	39
-PISFA0  -144	-88
-PISGA0  -136	-105
-PISGB0  -136	-101
-PISGC0  -136	-99
-PISGD0  -135	-93		; Unused
-PISGE0  -132	-89		; Unused
+PISFA0  -144	-69
+PISGA0  -125	-97
+PISGB0  -125	-82
```
mc776, line 2191: Consider increasing B y offset to -87 and C to -92, and the flash to -74. The animation looks a lot smoother.

mc776 (changes requested):

mc776:
Can I ask what the rationale for getting rid of the back lights in the flash frame is? Personally I don't particularly care either way (tbh I'd get rid of those lights altogether if it were up to me), just wondering.

Anonysussy:
It's mostly due to an alignment issue I found when I testing out the pistol for myself where the flash lines up but the lights didn't

mc776:
DOOM01

Here's the current version... I really can't see how this got messed up just by adding the second hand.

mc776:
Looking at the pictures again and I think the real problem with the flash is that I overdid the bright part over the front sight post, so the entire thing looks lower than it actually is.

Can you replace the PISFA0 with this one?
pisfa0

Here's a preview of how it looks, with both the old and new B frame:

mc776:
Actually, no, use this one instead, the flash really is overlapping the gun too much:
pisfa0
prev


(Note that this requires moving the PISFA0 offset up by 2 - -67 for the current, -72 for my suggested offsets.)

selliott512:
Looks good. I was initially concerned that this would take up too much screen real estate, but that seems to not be an issue in game given the way it is cropped.

mc776:
One more little enhancement I'd like to suggest: visible trigger pull on B frame.
pisgb0

mc776:
B frame with trigger pull:
pisgb0

C frame with the corners near the fingers smoothed out:
pisgc0

mc776:
C frame with slightly smoother corners on the slide and a bit more shadow around the index finger:
pisgc0

mc776:
Any comment about the requested changes?

Anonysussy:
Can I position the pistol a bit higher? It would probably help highlight the 2nd hand a bit more

mc776:
I think these offsets work pretty well:
```
PISFA0  -144	-70
PISGA0  -125	-82
PISGB0  -125	-85
PISGC0  -124	-84
```

mc776:
Did you see the updated flash I posted earlier? It actually fixes some proportional problems in the current one.
pisfa0

With this and the recent gun frame updates my suggestion goes to
```
PISFA0  -144	-71
PISGA0  -125	-82
PISGB0  -125	-85
PISGC0  -124	-84
```

Anonysussy:
I've been considering going with
PISFA0 -144 -66
PISGA0 -125 -84
PISGB0 -125 -80
PISGC0 -124 -88
The reload animations of yours is smoother I must say but the firing animations is kinda weak, I also doubt many people would go out of their way to fire the pistol once at a time to see the reload animations, so I don't really bother with it that much, either way, should I go with your suggestion or going with my own offset?

mc776:
Let's go with mine.

mc776 (approved):

PR #1195, opened 2023-11-11
build: Delete unused files

selliott512:
Delete files that have no impact on the three IWADs built by searching for references to them. Deleting these files has no impact on the three IWADs built so long as "VERSION" in "Makefile" is changed to a static string.

----

This is the same thing as the now closed [PR #661](https://github.com/freedoom/freedoom/pull/661) . I have again confirmed that these deletes have no impact on the IWADs generated after updating to latest master. They are also not referenced elsewhere in the build.


mc776 (approved):

mc776:
I'm merging this since every outstanding issue appears to have been addressed.

selliott512:
@mc776 wow, thanks. The old one was open since 2019. Phew. :-)

PR #1196, opened 2023-11-13
graphics: add website to help screen.

mc776:
Also added a reminder to hit Escape to get out of the help screen when I kept pressing F10 while testing.
help

### DIFF
--- a/graphics/text/Makefile
+++ b/graphics/text/Makefile
@@ -69,58 +69,60 @@ helpbg.png: ../interpic.png
 help.png: text_strings helpbg.png
 	./smtextgen help.png 320x200                 \
 	   -background "helpbg.png"                  \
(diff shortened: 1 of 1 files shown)


mc776:
Question: Do we really need that "Help" there? It seems pretty self-evident what that screen is, and without it we'd have so much more vertical room to put the actual help stuff.

selliott512:
I don't mind the "help". I could see a newbie user trying buttons, or accidentally pressing F1, and then thinking they are looking at their inventory.

selliott512 (approved):
Nice.

I compared the images before and after. I see that you moved things up a bit by adjusting the Y-coordinates to make room for the final line with the two new text strings. Seems like a safe and reasonable change.

PR #1197, opened 2023-11-14
sounds: louder DSOOF

Talon1024:
This is the original, unsoftened edit of DSOOF. Other old-school FPS games like this have loud DSOOF (or equivalent) sounds, so why does Freedoom's have to be quieter than the other sounds?


mc776:
The loud oof sounds were one of the worst things about that era.

Talon1024:
I have to make a conscious effort to hear the current DSOOF over the music, and even then, it's still very hard to hear, even if I turn off the music.

mc776:
I remember what happened in #1147 now.

The louder DSOOF was *obnoxious* and made very little sense to be that loud on basically any surface that wasn't a metal panel.

The current one is plainly audible for me nearly every time. Maybe it's an issue with your sound settings or bass?

Talon1024:
I use Linux (Kubuntu 22.04 LTS), which doesn't have bass settings in the volume controls.

The unfortunate truth is that I suffer from [misophonia](https://www.health.harvard.edu/blog/misophonia-sounds-really-make-crazy-2017042111534), so whenever I play video games, I turn the volume down low so that it's comfortable for me. I know it's ironic that I would make a sound louder, but there's a middle ground between "not loud enough" and "too loud".

mc776:
I don't want to do it because I don't know how to change the volume on a wav file losslessly on Saucedacity, but can you try/upload a version that's halfway beween these?

Talon1024:
To reduce the volume of a sound, set the gain on its track, and then export your project.

mc776 (approved):

PR #1199, opened 2023-11-17
sprites: adjust PISGA0 perspective.

mc776:
The recent two-handed sprite update raised the sprite further up the screen - which looks good, except the gun now looks like it's pointing somewhere above the center line of the screen.

This flattens the A frame out a bit. B is left as-is as it just ends up looking like recoil when you shoot.

Before:
DOOM00

After:
DOOM01


Anonysussy:
I would like to see both the 2nd hand and the finger grip get a bit of rework though, they just looks off to me, maybe some rework with the shading of the 2nd hand and the finger grip so it hold the other hand more naturally.

mc776:
It looks fine to me tbh, but go ahead and post some revisions if you have them.

Anonysussy:
Nah, that would took a lot of effort, I mostly did sprite edits rather than spriting from scratch, though I made this unfinished concept of the pistol few days back and it should probably gave a better idea of how the finger would look like
pisga0concept

selliott512 (approved):
Looks good. I'm not sure where the gun is pointing exactly, but it gets it closer to proper sight use (lining the front and back sights up). Here's an after picture with a small green cross hair in the center of the screen (a GZDoom feature). It seems plausible that it is pointing to the cross hair (if infinite distance is assumed), but others are welcome to draw the perspective lines and figure it out.

new-gun-sight

Issue #588, opened 2019-08-23
Textures: Implement FD2 stock textures into FD1


Proluiz:
Why we havent have all textures from freedoom phase 2 inside fredoom phase 1?

For example we can see small chuncks of phase 2 texture data inside phase 1, like some metal textures, aquatex and plutonia bricks.

Based on what i said here: https://www.doomworld.com/forum/post/2020360
I really want to start using these textures to make episode 2 very unique, if you guys let me have these textures in freedoom phase 1 i'll be extreme pleased.

And yes i'm already aware that doom1 and doom2 dont share the same switchs sets and must be kept untouched to keep the compatibility with mods.

fragglet:
Sounds reasonable to be honest.

MatthewTheGlutton:
What about other assets like P2 monster sprites and sounds and SSG?

Proluiz:
not necessary, doom mods that uses doom2 exclusive assests like monsters do it by itself, an example:

Complex doom and Knee deep in zdoom

Ferk:
Honestly it would be quite convenient if there was a version of the wad that used some MAPINFO magic to add phase 2 as if it was an episode 5. Then everything would be contained in a single wad (would wads from both Doom1 and Doom2 work with it?).

I know this would not work in anything close to vanilla, or even prboom, though.

chungy:
There could possibly be a (wadsmoosh)[https://bitbucket.org/JPLeBreton/wadsmoosh/] fork to combine all of FreeDM, Phase 1, and Phase 2.

Other than that, Phase 2-specific sprites like monsters and SSG should be excluded from Phase 1. The rest of the textures should be fine.

PR #1200, opened 2023-11-19
levels: fix map15 southeast safety lifts.

mc776:
They still had the 20% damage inherited from the old hurtfloor value.


selliott512 (approved):
Looks good. 

For those curious this changes sector #588, which is in two parts. It's the lift that lifts you out of the outdoor slime pit on the right side of the level:

map-15-sector-588

Only that sector was changed and nothing else:
```
diff --git a/levels/map15.wad b/levels/map15.wad
index 20b1c761..151942af 100644
--- a/levels/map15.wad
+++ b/levels/map15.wad
@@ -134882,7 +134882,6 @@ sector
 heightceiling = 184;
 heightfloor = -160;
 id = 69;
-special = 16;
 textureceiling = "F_SKY1";
 texturefloor = "AQF024";
 }
```
Which makes this a very safe change. 

I can elaborate how I setup `git diff` to produce output like the above if it's helpful.

PR #1201, opened 2023-11-21
levels: minor tweak for E1M1

Anonysussy:
Slightly tweaked E1M1 to be more beginner friendly, (the blue keycard are now optional for unlocking a secret)


mc776:
Can you re-do this PR by making a new branch? See here for how to do it: https://github.com/freedoom/freedoom/pull/1198#issuecomment-1815175395

(Might want to hit that "Sync Fork" button on the right beforehand though)

mc776 (changes requested):
This has all the commits from the previous pistol update. This can cause confusion and accidental reversions of unrelated things. Please do a new branch based on an updated fork.

mc776:
...and only now do I read the actual description of what's changed.

I'm going to disagree that this is helpful at all - it'll just create more confusion than anything else. (do you need the blue key or not?)

I would like to see the route *to* the blue key be simplified, though - that ambush and barrel hazard is what you'd expect from a much later map. (Maybe just go back to more or less what it was before Shino added that ambush last year...)

mc776 (dismissed):

mc776:
Rethinking this concept again and... yeah this actually does make sense.

That entire upper route works much better as-is as an optional challenge rather than trying to dumb it down so it work as an easy starting map while still somehow not losing interesting content for more advanced players.

But that said... should this call for getting rid of that trail of health bonuses that was designed to get them to go right?

mc776:
I just took a look at the old (latest release) C1M1.

It's... actually no easier to see that door now than before *in the slightest*, the pickups are doing literally all the signposting work there!

With that in mind here's an alternate placement, feel free to suggest any others that don't point towards the upper door: 4 HBs on the sides of the gate, meds on the left in a corner, the rest of the HBs in a trail at the bottom of the lift as a useless-for-signposting pleasant surprise:
[e1m1.zip](https://github.com/freedoom/freedoom/files/13433136/e1m1.zip)

DOOM00

mc776:
On further reflection I'm going to retract my approval for now - I'm not insisting on the specific solution I posted, but we really should get rid of that health trail and keep the signposting in that direction to a minimum.

mc776:
Need to add back the monster blocking lines that got deleted when dumbing down the armour pickup.

mc776:
I also do not approve of dumbing down the armour pickup - it's just 5%.

(Also kinda looks like a dick...)

mc776 (changes requested):
outdoor area needs work, i'll upload the fix

Anonysussy:
> I just took a look at the old (latest release) C1M1.
> 
> It's... actually no easier to see that door now than before _in the slightest_, the pickups are doing literally all the signposting work there!
> 
> With that in mind here's an alternate placement, feel free to suggest any others that don't point towards the upper door: 4 HBs on the sides of the gate, meds on the left in a corner, the rest of the HBs in a trail at the bottom of the lift as a useless-for-signposting pleasant surprise: [e1m1.wad.zip](https://github.com/freedoom/freedoom/files/13433504/e1m1.wad.zip) (EDIT: rebuilt nodes)
> 
> DOOM00

Just played around the map for a bit, so far no problem in sight here.

mc776:
Here' an attempt to re-implement the proposed changes, along with an alternate solution to the armour.
[e1m1.wad.zip](https://github.com/freedoom/freedoom/files/13434196/e1m1.wad.zip)

Similar placement of the starting area health, except all the lower potions are far enough that you don't overshoot them if running off the platform:
DOOM01

The southern area is moved back up north a bit, making it a little closer to the pre-winter2022 E1M1, since there really was no point in having it down that far if it doesn't do anything to aid *or* conceal the southern route anyway:
DOOM03

And finally, the armour. Learning how to look for small gaps and jumps to avoid hurtfloors is an important skill that this little platform is perfect for helping the player discover - and if they don't, at least they get to learn to do a little cost/benefit when considering whether to take 5-10hp for that armour when there's health lying around elsewhere.

But that all said, yes maybe a new player should get a little handholding with finding the small gaps. So I propose a trail of health bonuses:
DOOM02

Anonysussy:
Looks good enough for me, btw why's the wall offset reverted back to the previous one? I think I already fixed it previously
Screenshot_20231122_133757

mc776:
Oops, didn't even realize that was a thing or that you'd addressed it in here.
[e1m1.wad.zip](https://github.com/freedoom/freedoom/files/13434912/e1m1.wad.zip)
Slightly different fix: bump everything up a bit since we no longer need that extra space for the blue indicator.

Anonysussy:
[e1m1.zip](https://github.com/freedoom/freedoom/files/13435500/e1m1.zip)
did some minor edits on slade, you could test it out to see how it goes

mc776:
Oops, just noticed I'd misaligned a flat when simplifying the exit door:
DOOM00

This fixes that by shrinking the exit room just a tiny unnoticeable bit. I haven't even moved any of the actors inside it so there shouldn't even be any LOS consequences:
[e1m1.wad.zip](https://github.com/freedoom/freedoom/files/13441912/e1m1.wad.zip)

Anonysussy:
Looking pretty good so far

mc776:
From the manual:
> Explore the rest of the area. You will find two doors, one of each type: the bottom one can be used directly, like the lift, **while the top one is opened by using a nearby switch**.

If we go forward with the current state of things this part of the tutorial in the manual will need to be updated.

Anonysussy:
Updated the English manual, can't really do much of the Espanol one though, I can't speak Spanish

manual/manual.adoc
```diff
@@ -281,9 +281,8 @@ redoing a previous thing.
   then hit Use to call it down. Get on it and it will take you back up.
   Pick up the items in the upper area to restore or boost your health.
 
-* Explore the rest of the area. You will find two doors, one of each type:
-  the bottom one can be used directly, like the lift, while the top one
-  is opened by using a nearby switch. Once you've decided which way to go,
+* Explore the rest of the area. You will find two pathways, one of each type:
```
mc776, line 284: I actually already rewrote this:
```
Explore the rest of the area. You will find two doors, which can be opened with the Use key just like the lift. The lower one will take you closer to the exit, while the higher one will open up an optional more difficult but more rewarding route. Once you've decided...
```

mc776:
Here's what I got from DeepL:

```
Explora el resto de la zona. Encontrarás dos puertas, una de cada tipo:
  que se pueden abrir directamente, como el ascensor. El inferior te
  acercará a la salida, mientras que el superior te abrirá una ruta
  opcional, más difícil pero más gratificante. Una vez que
  hayas decidido qué camino tomar, abre la puerta y prepárate para
  empezar a jugar Cucú de las Pistolas nuevamente.
```

Anonysussy:
@mc776 about the requested review, i just misclicked, sorry about that.

mc776 (approved):

PR #1205, opened 2023-11-26
scripts: Add endoom-tool

selliott512:
Add script endoom-tool which can display, split, cleanup and join ENDOOM lumps.

----

Our recent work with the ENDOOM lumps inspired me to create this tool in part because I didn't want to install Playscii, and part because I thought it might be nice to have a simple tool for the task without any new dependencies.

With this installed if you type `make display-endooms` you should see the two current ENDOOM lumps (one for FD1 and FD2 as well as one for FDM). Here's what it looks like on my Fedora (GNOME) system:

fedora-display-endooms

Here's what it looks like an a MacBook Pro:

mbp-display-endooms

The difference has to do with the way those systems each handle ANSI color codes, but it may be reflective of variations in game engines as well. I don't have Windows to try, but from Googling it should work with an updated Windows 10 or 11 system.

The tool has the ability to split and ENDOOM lumps into three ordinary text files:
 * foreground
 * background
 * text

Where `foreground` and `background` consist of letters representing the color, and `text` is a UNICODE (UTF-8) text file with just the text. It is possible to edit these files with an ordinary text editor, and them join them to form an updated ENDOOM lump.

It should be possible to use this tool to make things like the URLs that appear in the ENDOOM lumps variables in the `Makefile` since the `text` file can then be edited in some automated way (`sed`, etc.) but I haven't tried that.

Our main ENDOOM contains extraneous text that is not visible due to the foreground and background both being black, which can be seen with the `-p, --plain` option (plain text, no colors):

fedora-plain

It probably doesn't matter much, but it can be fixed with the `-c, --clean` option as well as the warning about nulls that appears at the bottom of the screenshots (which also probably doesn't matter much).

### DIFF
--- a/scripts/endoom-tool
+++ b/scripts/endoom-tool
@@ -1,0 +1,565 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# endoom-tool - Display, split, cleanup and join ENDOOM lumps
+#
+# This script can be used to work with ENDOOM lumps. To see the ENDOOM lumps
+# used by Freedoom the following make target can be used:
+#   make display-endooms
+# There are other tools to work with ENDOOM lumps, such as Playscii, but this
+# tool can be helpful in some cases.
+#
+# If the display feature is used then it is important that the terminal support
+# both Unicode and ANSI color codes. The former is almost certain, and ANSI
+# color code support is likely, but try something like xterm if not.
+#
+# For information about the ENDOOM format:
+#   https://doomwiki.org/wiki/ENDOOM
+#
+# Color letters, see in the foreground and background files after a split,
+# relate to the Doom color numbers in the following way:
+#
+# Doom Color Number  Color Letter     Doom Color Number  Color Letter
+# -----------------  ------------     -----------------  ------------
+#                 0             .                     8             *
+#                 1             b                     9             B
+#                 2             g                    10             G
+#                 3             c                    11             C
+#                 4             r                    12             R
+#                 5             m                    13             M
+#                 6             y                    14             Y
+#                 7             w                    15             W
+#
+# The upper case letters in the above correspond to bit 0x08 being set, which
+# is extra bright when applied to the foreground, and blinking when applied to
+# the background.
+#
+# Examples
+#
+#   The following examples assume that the script is run from the top level
+#   build directory, but it should be possible to run it from anywhere.
+#
+#   To display (-d) the main ENDOOOM lump used by Freedoom:
+#     scripts/endoom-tool -d lumps/endoom.lmp
+#
+#   To display (-d) a plain (-p) (no colors) version of the same ENDOOM lump:
+#     scripts/endoom-tool -dp lumps/endoom.lmp
+#
+#   To display (-d) a randomly (-r) colored version of the same ENDOOM lump:
+#     scripts/endoom-tool -dr lumps/endoom.lmp
+#
+#   To split (-s) the ENDOOM lump into three files (foreground, background
+#   and text) into directory /tmp/endoom
+#     scripts/endoom-tool -s /tmp/endoom lumps/endoom.lmp
+#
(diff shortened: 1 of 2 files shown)


mc776:
I am in favour of this in principle. (but I do not trust my ability to review code and thus why this is not merged already)

selliott512:
> I am in favour of this in principle. (but I do not trust my ability to review code and thus why this is not merged already)

I understand. @fragglet , would you be willing to review/approve it? It should be pretty self contained and safe.

scripts/endoom-tool
```diff
@@ -0,0 +1,565 @@
...
+        read_write_binary = "wb" if args.join else "rb"
+        endoom_hand = open_file(endoom_path, read_write_binary)
+    except OSError:
+        if args.tolerant:
+            endoom_hand = None
+        else:
+            print("ENDOOM", endoom_path, "could not be opened for",
+                ("write" if args.join else "read", ), ".", file=sys.stderr)
+            sys.exit(1)
+    if args.join or args.split:
```
fragglet, line 336: You have a lot of these `if args.foo` lines throughout the file (24, if my search box is to be believed). It's a good indication that the code is factored wrong.

This function (`process_endoom`) doesn't have a clear purpose as it currently stands. The name is a good indicator of this - it's basically "do what this script does". As a suggestion, try making four copies of this function named `join()`, `split()`, `clean()` and `display()`, and then go through each and remove all the code that doesn't apply to that particular one. For `clean()` and `display()` you'd be able to delete this block, for example, and then for `join()` and `split()` you don't need the `if`.

You might find that the result is much clearer. If there's duplicated code you can then factor that out into common functions that they use.

fragglet (changes requested):
As it stands it needs more work before I'm able to properly review it. However, I also have questions about the purpose of the script and how it relates to Freedoom.

The program has subcommands to split and join ENDOOMs into three separate plaintext files. It's an interesting idea in that it fits with the [philosophy of making things easily editable as plain text](https://www.oreilly.com/library/view/the-pragmatic-programmer/9780135956977/f_0035.xhtml) - I can see the appeal. However, there are a couple of drawbacks. One is that it seems like it would be challenging to have to edit three separate files simultaneously to keep them in sync. Secondly, it seems like something that a lot of contributors to the project would find complicated and really struggle to use. 

So if your intention here that we eventually have three separate text files that get combined into ENDOOM as part of the build process, I have some reservations. If that's not your goal, then it's effectively just a new ENDOOM editing tool - and that's fine, but does it belong in Freedoom? There are plenty of other tools out there that people use for editing Freedoom and those tools live in their own repositories. Maybe this would be better off as its own separate project, and it will also be clearer that it's a tool anyone can use for their WADs, not just people contributing to Freedoom.

selliott512:
> However, I also have questions about the purpose of the script and how it relates to Freedoom.

Just to clarify, I in no way want to require people to use this tool to edit endooms. The definitive data will continue to be `lumps/endoom*.lmp`. Most people will only use this tool to see the current endooms with:
```
make display-endooms
```
If they choose too. Or they can just run an engine that supports it, or some other tool.

It's possible that only I will ever use the split and join features. It's also possible that it may be useful to use it to update the endooms as part of the build process so that they could contain the date, version, or whatever we want to keep in sync with some global variable, but that's just an idea. We don't have to do that. Even if we were to do that the definitive copy would still be `lumps/endoom*.lmp`. The build process would then split, update the text, and join.

scripts/endoom-tool
selliott512: One possibility is that we commit this as is now and you can make those changes as you see fit. Or we can close this PR and I can just add `endoom-tool` to one of my personal repos.

selliott512:
I've moved `endoom-tool` to my [wad-tweak repo](https://github.com/selliott512/wad-tweak).

selliott512:
I've moved `endoom-tool` to my [wad-tweak repo](https://github.com/selliott512/wad-tweak).

Issue #1210, opened 2023-11-30
Cyberdemon Sprite Error


uhbooh:
It seems like CYBRA and CYBRE are swapped in the Assault Tripod sprites. Looks like its feet are twitching around when it fires rockets. When you swap them around the animations for wlaking and firing look correct.

PR #1212, opened 2023-11-30
sprites: address #1210; fix tripod offsets.

mc776:


selliott512:
I've confirmed that this change fixes the extraneous leg movement while firing missiles, but there is much more movement when in standby / inactive (when you sneak up from behind before firing a shot). Do you want that? Perhaps only some of the a/e sprites should be renamed?

Also, this PR adds non visible changes to some of the 10 sprites. Comparing the a/e sprites before the change to the e/a sprites after the change:

| Before | After | Same? |
|--------|-------|-------|
|cybra1.png|cybre1.png|N|
|cybra2a8.png|cybre2e8.png|N|
|cybra3a7.png|cybre3e7.png|Y|
|cybra4a6.png|cybre4e6.png|N|
|cybra5.png|cybre5.png|Y|
|cybre1.png|cybra1.png|N|
|cybre2e8.png|cybra2a8.png|Y|
|cybre3e7.png|cybra3a7.png|Y|
|cybre4e6.png|cybra4a6.png|Y|
|cybre5.png|cybra5.png|Y|

But they all look the same. Converting the PNGs to PPMs made them all match. Avoiding those changes so that the PR consists only of renames would make it easier to understand.

mc776:
I always run things through pngcrush. It seems incapable of being undone, though, and it all gets flattened when the wad is built, so as long as all the pixels are in the right place in the right colours I can't see how it matters unless there's like some malicious code in the metadata or something.

(Or if there's a better way to optimize the pngs so they're even smaller...)

As for the extraneous movement, it's always been extraneous and never looked right - the 3-4 angles are acceptable, 5 is janky but no more janky than the firing frames before this fix and it's really an issue with the art that will have to be dealt with on those terms.

....that said, I should take a look at some of these sprite offsets because they do *not* look right...

mc776:
Yeah, I didn't notice the tab thing until after I updated it - I had edited the offsets by clicking and dragging in SLADE, then exported a wadinfo file with Deutex and copypasted that, and that's how I learned Deutex uses tabs.

I'd personally like to make it all tabs going forward to be consistent with Deutex's default, which I would have done in this commit but I don't know what the prevailing preference is.

Issue #1161, opened 2023-10-25
Urban Jungle (Phase 2, Map 16) violates the ears upon finishing


Bracket-H:
Please, this wasn't too good for my headphones either, don't potentially wreck hardware please... 

That city explosion ...perhaps tone it down a little, or if its a prefab sound (then what's wrong with you recording/making such a thing?!) ...then change it out please.

Ow...

unithecutie:
This was addressed in the Freedoom Discord server! It's using the boss brain death sound.

Calinou:
I agree the boss brain pain/death sounds are really loud in general – they seem quite a bit louder and crunchier than their Doom 2 counterparts.

unithecutie:
https://github.com/freedoom/freedoom/assets/139153158/e7035932-c6e3-41f5-85f2-10234f7d3992

mc776:
I've taken the liberty of just making this issue about the boss brain death sound generally.

mc776:
Reverting that rename now that I've been doing more (normal) boss brain stuff.

The big issue here is that we've got the 3 boss brains under slow crushers so we're getting all 3 pain sounds stuttering once per tic before the actual death sound. Any source port that properly mixes all three is going to make this much worse.

I've made a variant that only has a single boss brain a bit further up, which itself is free from the crushes and only takes damage from the barrels. There's also a hidden barrel that creates an explosion sound to foreshadow the main event a bit.
[map16.wad.zip](https://github.com/freedoom/freedoom/files/13417301/map16.wad.zip)

mc776:
Variant with a sequence of explosion sounds designed to draw the player's attention to the right.
[map16.wad.zip](https://github.com/freedoom/freedoom/files/13417415/map16.wad.zip)

mc776:
Right now the map's lone bossbrain death sound with no pain sound is the minimum possible amount of bossbrain sound without mods or completely ruining the ending of the map. I'm taking the liberty to close this.

PR #1213, opened 2023-12-01
levels: address #1161.

mc776:
After some experimenting with different explosion sequences I think this one strikes the best balance while being compatible with both vanilla and limit removing.


selliott512 (approved):
It looks like you've replaced four boss brains with one. It's not obvious to me why we need to have this special boss brain ending for this level, but I think this is an improvement.

PR #1214, opened 2023-12-01
More track shuffling.

unithecutie:
Same reasoning as #1211, to accommodate for maps moving.


mc776:
Leaving your last comment in the commit intact but I personally think "Tunnel" with its darkness and dissonance works way better for E2M3 than "Workout" (and "Workout" works better for DM06).

PR #1215, opened 2023-12-01
Restore "Panic".

unithecutie:
"Panic" gives me the same vibe as the current track, but in a better way.
This track was also replaced mid-development so for it to be never used anywhere would be a shame.


mc776:
I actively dislike this change but am fully aware that this is a matter of personal taste.

I would, however, say that the vibe is different - the current track by BobFromReboot is much more fast-paced and actiony, this is... some kinda sprawling... rock.... thing.

Would like a third, fourth, etc. opinion before this is committed, rejected, or another more deserving track replaced.

I note that plenty of resources have been added and then replaced between the last release and now.

unithecutie:
> or another more deserving track replaced.

What tracks do you feel **need** replacing? My second thought was DM14.

mc776:
I don't think anything right now "needs" replacing exactly, but I can try going through the DM tracks now...

DM17 (d_stlks3.mid) especially sticks out as something I'd consider a candidate for replacement. It has some interesting ideas but as a whole it feels like a jumbled, needlessly dissonant sequence of unrelated things.

Wouldn't object to this replacing DM14 (Munt Battle), but I'm starting to suspect that this one could actually be saved with some re-instrumentation.

DM04 (Rush) seems like a potential candidate, at least I wouldn't mind if this were replaced with Panic but I think it's definitely a much more solid track than d_stlks3.mid.

(Playing chocolate and going backwards one map at a time... stopping at one where I'm all like "hey this one's cool we should definitely keep this also why does it sound familiar" and then realizing I'm on DM10, heh.)

unithecutie:
Accidentally closed this...

mc776 (approved):

mc776:
I'll leave this open for a bit in case anyone's got a second opinion but I'm not really expecting much of a contrary opinion at this point...

PR #1217, opened 2023-12-04
demos: replace P1/2 DEMO1s with singleplayer demos.

mc776:
> For Phase 1 and 2, do not have a multiplayer demo in the DEMO1 slot.
> Vanilla Doom (and some source ports) do not allow using the “load
> game” option while in a network game, and network game demos count as
> such.  The other three demo slots are fair game, since they do not
> near-immediately block out the load game functionality when starting
> Freedoom.
> 
> FreeDM is exempted since it is not intended to be a single-player
> game.

What I've got here are pretty ~~short and~~ hastily done so if anyone wants to do better feel free to suggest.


mc776:
The P1 demo here is 45 seconds while the P2 almost 2 minutes... should've been the other way around given how much sooner Doom 1 transitions into the demo from the title screen.

mc776:
P1 now almost 3 and a half minutes, which might be a little *too* long but should definitely give everyone enough time to decide on which of all 6 slots to open.

selliott512 (approved):
Looks good. Single player feels like a good introduction the technical issue aside anyway. I watched them both in Chocolate Doom. Action packed. Here's how long until I got the prompt back for each demo.

| Level | Time |
| --- | --- |
| E1M6 | 3:26 |
| MAP03 | 1:31 |

mc776:
Taking one last look to make sure I didn't get the Map03 versions confused... I like how my shooting is so abysmally bad at the start of this that it initially actually looks like a desync.

PR #1220, opened 2023-12-07
sprites: make hatchling/octoball explosions snappier.

mc776:
Something that's bothered me for a while now: the hatchling dies, it slowly oozes blood and crumples, the fireball explosion sound plays in the middle of all that bleeding and crumpling, then it slow-motion explodes in total silence.

Part of this the problem with the way the lost soul death is hard-coded - 6 tics per frame is just way too long for any decently snappy explosion. At best we can have like 2 actual "exploding" frames and the rest is just the bits floating in midair slowly fading away.

This PR skips the bleeding frames entirely in favour of what looks like a failed attempt at getting in one last attack, then a quick explosion right on the A_Scream frame followed by a few frames of that ring of glowing gore bits fading into nothing.

A similar issue to the hatchling death animation exists with the octaminator projectile as well. The lines coming out of the current B frame still imply an explosion rather than a lingering cloud of debris.


selliott512 (approved):
I like it.

At first the difference was not obvious to me since I tried it in MAP32 in the room that has a bunch of matribites in it, but there was too much activity to really understand what was happening. Once I added a lone hatchling to a level the improvement was obvious.

PR #1223, opened 2023-12-08
Add replacement Episode and Difficulty strings to dehacked file

superdave938:


mc776:
Can I ask what uses these? Just wondering if there's possibly any good reason why these haven't been added a long time ago...

superdave938:
Mainly source ports like gzdoom
Helpful for very specific circumstances where they might come in handy

mc776 (approved):
Just tried this with Hideous Destructor's episode select in Phase 2, works like a charm (and obviates something I had to do in HD for it to look right)

PR #1225, opened 2023-12-10
manual: Fix warning in tutorial steps 1-15

selliott512:
asciidoctor-pdf outputs a warning when ordered lists are numbered entirely with "1.". It wants actual consecutive integers. Prior to this fix there were 14 warnings such as:
  asciidoctor: WARNING: manual.adoc: line 234: list item index: expected 2, got 1
in the English and French manual. Interestingly it amkes no difference in the PDF produced, but let's minimize the number of warnings in the build.

----

I know we talked about this, mc776. I'm hoping you don't object on principle (the principle being that asciidoctor-pdf shouldn't have such a warning because it makes lists harder to maintain, like you said) since it adds 28 warnings to `make dist` and `make install`. And there's something to be said for being explicit as well.

### DIFF
--- a/manual/manual-fr.adoc
+++ b/manual/manual-fr.adoc
@@ -182,48 +182,48 @@ Ce tutoriel vous présentera toutes les actions de base dont vous aurez besoin p
 
 Commencez une nouvelle partie dans Freedoom Phase 1, épisode 1 sur le niveau de difficulté facile et suivez les étapes. Ignorez tout ce qui vous ennuie ou vous embrouille, et refaites tout ce que vous trouvez difficile aussi longtemps que vous le souhaitez, avant de passer à l'étape suivante ou de refaire une étape précédente.
 
-1. Essayez de vous déplacer en avant, en arrière, à gauche, et à droite.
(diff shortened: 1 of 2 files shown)


mc776:
Can we just leave it in there? After this change if anything else needs to be changed then explicitly numbering everything like this just creates another way for some stupid needless errors to come up.

No end user will ever see those warnings.

mc776:
If it really bothers you that incredibly much I think the best solution is to get rid of the numbers entirely.

selliott512:
I did a [bit more research](https://github.com/asciidoctor/asciidoctor/issues/3252) and `.` in place of `1.` seems to do what we want. It appears as a numbered list, but without the warnings. If that's ok I'll update this PR.

selliott512:
Done.

I converted the PDFs to text before and after my change:
```
pdftotext freedoom-manual.pdf freedoom-manual.txt
pdftotext freedoom-manual-fr.pdf freedoom-manual-fr.txt
```
and diff'd to confirm that it makes no visual difference.

mc776 (approved):

PR #1229, opened 2023-12-12
New brick patches

uhbooh:
Two brick textures I made, sources are heavily edited freedoom patches and flats.


mc776:
Every objection I had to the previous attempt to replaces these textures applies to this one.

Even ignoring all that though this is definitely worth deferring given the drastic change to the brightness and apparent age of the bricks and how that would affect the look of many maps.

uhbooh:
I'm alright with postponing these if they cause any problems. When making them I had PWAD levels in mind, so I tried to imitate the originals while having a distinct appearance. In the Freedoom 2 levels the slime bricks are used a lot, but I can hardly find any walls that have the red bricks besides mapa16.

mc776:
Screenshots, after and before:
DOOM00
DOOM02
DOOM01
DOOM03

DOOM07
DOOM04
DOOM06
DOOM05

The red brick makes me realize why it looks like that: in low light and low res the bits of the mortar stick out in a bad way.

The green brick looks a lot better than I was expecting - the way the algae spills over really helps mitigate the effect I just mentioned about the red. It still feels too bright though, like unnaturally so, and it does mitigate the gloomy dismal vibe the current green brick is often deliberately used to evoke. I suspect if it could just be darker it could work a lot better.

Was it intended that the flat not be replaced? (If it is to be replaced I'd suggest tweaking it a bit so the green goo looks more like blobs instead of lines)

uhbooh:
Oh I forgot there was a flat version of that wall, I'll get around to making that after RW41_1 is finished. As for the colors being too bright, I shifted the greens down one index. Looks closer to the originals colors but I'd like to know what you think.
DOOM21

mc776:
Could you try taking the green just a bit darker still?

As for the red mortar, yeah I think trying brown could mitigate the aliasing effects... as much as I'm not a fan of making things more Doom-like without any technical necessity I suspect that this was why they made it brown in the first place...

uhbooh:
DOOM23
DOOM22

Here it is with the mortar a bit darker

mc776:
Here's an attempt at brown red brick mortar, darkening the green a bit further, and replacing SLIME13:
slime13
rw41_3
rw41_1
DOOM02
DOOM04
DOOM03
DOOM01

mc776:
At first I wasn't happy how the wall blended into the floor like that, but it's no worse than current:
DOOM01
DOOM00

uhbooh:
RW41_1darker
Slime bricks look pretty good, here is the one I made. It's a little bit lighter than yours to mimic the original texture's palette.

RW41_3darker
As for the red bricks, I used a lighter red range than the original because dark reds don't look very good in software render. In your screenshot the texture is mostly black in dark lighting. Brown mortar looks nice but you missed some gray pixels.

mc776:
yeah no that's *way* too bright

mc776:
As for reds, though, yeah my edit is definitely worse, though it really should still retain at least *some* of the moldy grimy look of the original.

uhbooh:
RW41_1darkerer
This better? Looks almost the same as yours but just a tiny bit lighter.

mc776:
What I had already seems slightly too light - it's the best I could do without losing noticeable amounts of detail.

uhbooh:
Well the original texture is a rather bright green, that's kind of what I was going going for.
Screenshot_2
That's in pretty dark level too

mc776:
.....oh. So *that's* what all this is about.

If another maintainer wants to merge this I wash my hands of it.

mc776:
"the original"

Why do you put it that way? Can we see some WIP stuff to be sure this is an original and not based on the id texture?

mc776:
> If another maintainer wants to merge this I wash my hands of it.

That said, even if other maintainers are in favour of this I think we should definitely defer this and any other conscious efforts to make Freedoom look more like a Doom derivative until after the next release, so that that release can be a fallback in case we run into any further troubles later on.

uhbooh:
I’m alright if you want to delay these until after the next release. As for originality, I used gimp to edit freedoom textures and flats. The red brick is a recolor of flat5_4 and the slime bricks are edited from one of the aquatex walls. If being too derivative of Doom is an issue though, I can redo these textures to be more unique. I’m definitely aware of that problem in this project, so I made the mortar gray intentionally to look different from the Doom 2 texture.

uhbooh:
Also in the discord server you can see most of my WIP stuff in the art dev room

MerrittK:
> "the original"
> 
> Why do you put it that way? Can we see some WIP stuff to be sure this is an original and not based on the id texture?

I know I use the term original to refer to ID software stuff, even when the work is entirely based on other freedoom assets.

uhbooh:
Hello dropping in my current opinions on these textures. I hope everyone had a merry Christmas.

Looking at them again I do think the new RW41_1 looks a bit too close to the original patch. Although there's a lot of patches in Freedoom that are like that, I understand why its better to avoid imitating ID software too closely for a project like this. When I made these I matched the shape of the bricks in the originals, because PWADs are usually made with those in mind. For RW41_3 I chose the colors intentionally to look distinct from the Doom 2 bricks which had dark brown mortar and dark red bricks. The gray mortar and lighter red looks more like a real brick wall, with software rendering being much kinder on it in dark rooms as well. In vanilla Doom there is a strange absence of very dark reds in the palette, which makes darker reds turn black in dim lighting.

To make these I used gimp and heavily edited Freedoom's AQRUST01 and FLAT5_4, so they are 100% compliant with the free software license. In the discord I got advice on how to make them and you can see the work-in-progress stuff there. If someone has any advice on how to improve these and make them more original then let me know. Thanks.

mc776:
> RW41_1darkerer This better? Looks almost the same as yours but just a tiny bit lighter.

ok I'm feeling way more chill about all this now that we've got a release out.

I made a flat based on this:
slime13

It looks fine on its own and tiles correctly, but it looks comically dumb on the eastern part of Map11.

Would you mind if I edited this PR to do the following:
- add the flat
- add back the old green bricks as SLIME13A and RW41_1A/BRICK13A
- add an edited Map11?

uhbooh:
Do you have a screenshot with that flat in map11?

mc776:
It looks fine as long as the floor's edges are orthogonal.
DOOM00
DOOM01

But once you start trying to make arbitrary shapes the fact that the grain of the bricks has nothing to do with the shape of the structure it starts looking kinda ridiculous:
DOOM02

It's still like that with the current but it's barely noticeable because the texture is so mushy already:
DOOM03

uhbooh:
Yeah I agree it doesn't work very well in that instance. Maybe we could just swap the flat in that sector.

mc776:
OK for me to edit the PR then? Or would you rather you do it on your end?

uhbooh:
Sure you can edit the PR. Do you want to change the other texture as well?

mc776:
My intention is to use your new textures as the replacements, then add the old ones as a new thing on their own. The edited Map11 will use both.

(The new flat actually doesn't look that bad in those lopsided octagonal rooms now that I'm editing the map and looking at where to fit everything - the big issue seems to just be the misalignment with the connecting wall texture)

uhbooh:
Sounds good, I like the current slime bricks in freedoom so having both versions is nice. My original intent was just to replace the red bricks and I did the slime bricks for fun.

mc776:
...and it turns out I can't. Possibly related to this being the master branch on your end?

This zip has everything needed. Just unzip it, open up https://github.com/uhbooh/freedoom/tree/master and drag all the unzipped files and folders into it.
[newbrik.zip](https://github.com/freedoom/freedoom/files/14135990/newbrik.zip)

uhbooh:
It should be updated now

mc776 (approved):

Issue #1230, opened 2023-12-13
Miscellaneous Phase 2 QoL issues


dftf-stu:
Greetings!

With things starting to slow-down now for the Christmas period I've time to play some games, and thought I'd do a playthrough of the latest _Freedoom_ "Dev" builds, to see how-far things have come since I last played them.  I usually play them in-order, so thought I'd start with _Phase 2_ this-time!

On _MAP03 "Crude Processing Centre"_ I do find the exit to the map rather abrupt.

In the "Area Map" pickup-room, you go-down the left-hand corridor, call the lift down and the moment you step on it, the map ends!

![DOOM0001](https://github.com/freedoom/freedoom/assets/83965343/ee9cf0db-70ca-437c-a778-bbdde358f5f5) ![DOOM0002](https://github.com/freedoom/freedoom/assets/83965343/44942ca9-8ffa-4939-b639-8e7a4bb07900)

This seems a bit odd, especially given once you collect the "Area Map" and then look on the Automap-screen, you can see a whole area which is inaccessible through normal means -- but which has a proper, modelled exit:

![DOOM0004](https://github.com/freedoom/freedoom/assets/83965343/1c0f7ad3-47c4-4bb0-a6d2-280b6d970822)
![DOOM0005](https://github.com/freedoom/freedoom/assets/83965343/3415b08b-71eb-459a-9b5b-47ae54c98b47) ![DOOM0008](https://github.com/freedoom/freedoom/assets/83965343/0e8198f2-9ff5-4eff-bf88-4b545b183f50)

If you "IDCLIP" to that exit and use it, you still get taken to MAP04, so it doesn't seem like a forgotten secret-exit.  And while I've heard of "multiplayer-only start-areas" (that can sometimes show-up on the Automap), I didn't think "multiplayer-only exists" were a thing?

So... seems a shame not to use that area to me! 🤷🏻‍♂️

(P.S.: apologies if there is any existing-ticket I'm supposed to tack this onto, but I could only find 1181 , which is for Phase 1.  And in the Closed tickets list, I could only-find 101, but that is over nine-years old so doubtful it even applies to the current MAP03!)

mc776:
The lift takes you out of the level; the other exit is for the deathmatch containment area.

I wonder if the deathmatch area could just be marked so that it doesn't show on the automap. It's small enough to memorize and nobody has time to check the automap in DM anyway.

I like the lift and the way it implies you're moving to a different place, but it really could be better marked as a map exit.

[...]

PR #1232, opened 2023-12-14
levels: minor fixes.

mc776:
Addresses #1230. Includes replacement Map03 demo.

Map03-4
- Map03 exit is now very clearly marked as one. The lift is given an appearance that implies that it is what you're standing on at the start of Map04.

Map09
- The entryway switch into the ruined storage administration area is gone and the "inside" switch now opens the bars directly.


mc776:
(no "bug" tag since these are QoL fixes.)

selliott512 (approved):
Looks good. I like marking the exit in MAP03, and matching it to MAP04. I could not quite figure out what you did in MAP04, but it must be minor and safe. I agree that the MAP09 change reduces the risk of non-fun switch hunts.

I played the demo. Nice. It was 1:19 for me.

PR #1238, opened 2023-12-16
music: new d_e3m9 by @LolaHThicc.

mc776:
Redo #1237 from an up-to-date branch without the conflict.


selliott512 (approved):
Seems good. I took the liberty of squashing the commits to one and keeping the first commit message.

I listened to it for a while and it's nice, or at least it doesn't have any obvious errors. 

As long as this `d_e3m9.mid` has everything you want then I think we're good.

PR #1239, opened 2023-12-17
graphics: use smoking wreckage for BOSSBACK.

mc776:
Always thought it kinda weird why the victory end credits shows the final boss intact instead of successfully destroyed.


selliott512 (approved):
That's a neat idea.

Also, if you don't mind I'd like to send you a private message. I'm at `selliott512@gmail.com`, and I've been logging into Discord daily.

PR #1240, opened 2023-12-17
Removed Vents.

MerrittK:
Hey, here is a change to the aquatex panels, it deletes the vents, fixing aliasing issues in software render.  

It might seem less detailed, but it does a lot to help the texture look good in complicated environments.


MerrittK:
Before
After

MerrittK:
After 2
Before 2

mc776:
I am definitely not going to approve this until I've seen a second or third or fourth opinion.

(I actually think it's a good idea but this seems it can get controversial)

MerrittK:
Main thing was the fact the vertical lines resolve weirdly at low resolutions. I can't really demonstrate in a single screencap, but what ends up happening is the vents start flashing white-gray-black when you move side to side in software render.

The single pixel lines get rounded differently for every position, creating a nauseating strobe effect. I understand if we wanted to keep them as "vents" but that would still require redoing them as something that doesn't use single pixel lines.

Xindage:
seems good for me.

mc776:
> Main thing was the fact the vertical lines resolve weirdly at low resolutions. I can't really demonstrate in a single screencap, but what ends up happening is the vents start flashing white-gray-black when you move side to side in software render.

I am definitely familiar with the issue, my only reservation is how the Aquatex patches are some of the best patches in FD right now and I'm hesitant to change them.

That said, right after my initial reply I realized that one of my *very first ever* submissions to Freedoom was to correct one of the Aquatex pipe textures so it would tile correctly, so I'm going to approve this and remove the deferred tag and see if anyone objects to the latter decision (or just merges it) before I hit Merge.

mc776 (approved):

selliott512:
This seems reasonable to me. It's not obvious to why the metal panels would have to have raised vertical lines (heat syncs?), so should be ok.

Incidentally, the phenomena where one spatial frequency is close to another, so it looks like the entire thing switches from bright to dark is a "Moiré pattern".

mc776:
Just some screenshots for reference - no "before", this is just to check to make sure the new textures don't look out of place where they currently are before this is merged:
DOOM03
DOOM02
DOOM00

dftf-stu:
I was just playing Phase 1's E1M2 in _Crispy DOOM_ a wonder if a similar-change might be considered for this texture (found in the red-door area, to the left of where you start that map; seen just-before the small outdoor area before the exit)?
C1M1_001

As you move towards it, or backwards from it, it really "strobes", with the lines becoming wider and thinner:
C1M1_002

mc776:
DOORTRAK looks horrible but the way it looks (uniform alternating lines that provide a lot of "detail" and "texture" without needing any effort to align it properly) is so heavily relied on by mappers for both id and FD that we really don't have anything to replace it with.

That map could probably just use something else there.

mc776:
...unless we were to greatly reduce the contrast so from that distance it just looks grey-black.......

mc776:
Worksheet:
newdoortrak

Current:
DOOM00

Attempt with a regularly spaced horizontal variation:
DOOM01
(note that this looks like ass up close, it's just a proof of concept)

...low contrast:
DOOM02

id, for comparison:
DOOM03

I think the current Freedoom DOORTRAK is as good as it gets.

MerrittK:
Perhaps we just make it a rule to not rely on large doortrak spans? It works best as a highlight anyway.

dftf-stu:
I had a little go myself in _Paint_ to see what ideas I could suggest:

FDDoorTracks

The groups, from left-to-right:
- A cut-out in the centre, where the doors would have an insert into (to stop you ramming them open)
- Sunken cut-outs / grooves (where a motor-driven cog on each-side of the door would push it up against)
- Raised-rivet style
- Chain / ladder style
- Rubber tyres (which would push the doors up, similar to how some fairground rides work)
- Just a random pattern
- Zipper-like teeth around a central bar

(Assuming the texture-size is 8x8 pixels, then there isn't a lot that can be done, no.)

Xindage:
i think that we must keep door track similar to what id uses, even with the problem that it have it must be close to the original, and its not everywhere that this texture is used that way in 32wide walls, the most used case is 8w and 16w max, i'm suggesting to keep doortrack out of this rework. if you want to suggest and new textures to be used as alternative ok, but do not replace doortrack.

MerrittK:
I moved this discussion to a new issue, so as not to extend the pull log. #1245

PR #1241, opened 2023-12-17
"Fun is Infinite at AGM" as MAP07

unithecutie:
MAP07's current track is currently way too short for the level, and this seems like a fitting replacement.


mc776 (approved):
Gave this a run in Chocolate, yeah it definitely fits this map.

~~Kinda circusy but the absolute 🤡🤡🤡 deaths I was getting this run were on point.~~

PR #1242, opened 2023-12-17
Change MAP07 name

unithecutie:
MAP07 looks almost nothing like a warehouse as of the map update. It also somewhat resembles an "AND" logic gate (blue key + red key = exit, 2 signals to output); hence the name.

Might not be a fitting name along with the rest of Cluster 1, but this is all I could think of. :P


mc776:
Was "Outer Storage Warehouse" even the original name of this map, or just the name of the previous Dead Simple-like?

unithecutie:
"Outer Storage Warehouse" has been the level name since 0.10; when the majority of level names were added.

mc776 (approved):
I'm looking at https://github.com/freedoom/freedoom/commit/2de1ff1bce864523dfc6a1f8a526db6a76f5fea1 and #667 and it seems this Map07 never had its own title. Approving this PR now.

PR #1244, opened 2023-12-19
demos: correct readme.

mc776:
See #1243.

### DIFF
--- a/lumps/README.DEMO
+++ b/lumps/README.DEMO
@@ -23,7 +23,7 @@ DEMO4: E4M6 Cooperative by Xindage, Erick and Korp - v0.13.0
 Phase 2:
 DEMO1: MAP03 Singleplayer by mc776 - v0.13.0
 DEMO2: MAP10 Cooperative by Xindage and Korp - v0.13.0
-DEMO3: MAP22 Cooperative by Stratos - v0.13.0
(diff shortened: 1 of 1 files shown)


mc776:
But before this gets merged, though: would it be better to put both singleplayer demos at the start, to maximize the amount of time the player has to load a game?

selliott512 (approved):
Looks good.

selliott512:
> But before this gets merged, though: would it be better to put both singleplayer demos at the start, to maximize the amount of time the player has to load a game?

I didn't see this in time, but that's a good question. It raises another question - is there any issue with having out of order level numbers for demos? Probably not, but  for FD2 I did a quick experiment where I changed all four demos to short reverse order single player demos (MAP04, MAP03, MAP02, MAP01 in the demo slots respectively). I let it run and Chocolate Doom played the first three demos as expected it played (MAP04, MAP03 and finally MAP02) but for whatever reason it started over without playing the last demo, which was MAP01.

I don't know if this is a glitch with the way Doom demos work, FD demos in particular, or just this reverse order experiment. I do know that my last demo, MAP01, is playable when requested explicitly:
```
chocolate-doom -iwad wads/freedoom2.wad -playdemo DEMO4
```

 I can look into it more when I'm back at the appropriate computer.

Xindage:
Some version of doom actually look for demo 4 and may crash if not present

PR #1246, opened 2023-12-20
textures: add backwards DOOR3.

mc776:
For maps that do things that might actually give an impression of an ostensibly horzontal-moving door that can be viewed from either side, or a pair of double-doors at the entrance to a level.


Xindage:
Can we see how it looks?

mc776:
DOOM01
DOOM00

MerrittK:
Looks like a good idea.

unithecutie:
That looks quite nice. :)

Xindage:
that could be useful even for normal doors if the mapper want to make "lock" be on the same sides of the door by using some offsets.

selliott512 (approved):
I like it. 

I verified that `DOOR3L` is a mirror of `DOOR3`, and that the dimensions match.

PR #1248, opened 2023-12-24
sounds: new DSDMPAIN.

mc776:
I've been messing around trying to replace the current sound which feels like an attempt at (not saying it is, but an original recording *trying to imitate*) a near 1:1 copy of the id sound.

Turns out this squishy hiss sound works surprisingly well for the wide variety of monsters that use it.


uhbooh:
I don't think that sounds very good IMO. Besides I don't see what's wrong with the current demon pain.

mc776:
It's not good, I admit, but the opinion of someone who's posted a PR of something that intentionally makes FD look *more* like an id knockoff is of limited value.

mc776:
That said, this is a very minor thing and under no circumstances should it be the cause of any further delay in a release.

I'll mark it as "deferred" unless someone in the process of doing the release really wants to add it and can remove the tag and merge it themselves.

uhbooh:
Sorry if I came off as a bit harsh. I think making the current DSDMPAIN sound less like the original is fine, but I think the current one we have is pretty distinct enough. By the way how did you make this sound?

mc776:
Making hocking loogie noises into the microphone and mixing them.

It was originally just a lousy placeholder for my own project but then I realized it was generally more plausible on average for any particular monster than the current thing which basically doesn't fit any of the FD monster designs other than the painbros (which other vocalizations don't sound like that anyway).

Since there's no demand for it though I'll close this.

PR #1250, opened 2023-12-29
sprites: pistol zombie isosceles.

mc776:
Just posting this as a suggestion, to be honest I'm not even sure if we should do this.

Given the recent change to make the player fire two-handed (#1193), it seems to make sense that the zombies would also have similar training to fall back on.

(And yes, this does provide a rock-solid excuse to make at least the E/F sprites flipping - frankly I'd be okay with making flips for the rest too.)

DOOM30
DOOM28
DOOM23
DOOM20
DOOM15
DOOM06


mc776:
(I hope it goes without saying that I would prefer, if this were to go in, that this PR be squashed and merged instead of simply merged)

unithecutie:
This looks quite odd, considering they're undead.

uhbooh:
Well in Doom the zombies are formerly trained soldiers. Not sure about the Freedoom lore though.

mc776:
I don't think there's anything suggesting that any of the zombies or the player are military. That's just "how you hold a gun" for a lot of people.

For me the biggest inconsistency in the PR right now is the fact that the zombies would be coordinated enough to *switch* to a two-handed grip that quickly. The original idea that I didn't yet have time for is to make them *constantly* clutching their guns, as a variant on the "zombies with arms extended" motif, which would allow for fully internally consistent flippable sprites (instead of the newly introduced full-8-rotation bloat) - that would make their movements a lot less coordinated-looking while keeping the pistol grip consistent between then and the player.

Come to think of it, on the other hand, that inconsistency also seems like a thing with the new player two-handed grip as well...

MerrittK:
They might have been soldiers, but the aliens controlling them are not. The zombies are puppets, and their aim sucks.

selliott512:
You have this marked as a question so I'm not sure you even want it approved. But anyway, the thought of making the sprites flipable is compelling. That would be a maintenance win. But other IWADs have not bothered with that optimization. 

Zombies holding the gun one handed in some sort of unthinking undead manner seems like what you might expect from them. Given that and the fact that we have screen shots, including the manual, that should be updated if this were to change my inclination might be to leave it.

PR #1251, opened 2023-12-30
sounds: softer bossbrain pain and death.

mc776:
Quiets down the original sounds, delays them a bit giving the impression of a delay in the sounds reaching the player, and adds a bunch of explosions.

A slightly longer delay on the pain sound would have been enough to address the awfulness of the old method of the Map16 ending, but at this point it doesn't really matter.

I don't know how to get the explosions louder without clipping - I've still got the aup3 file so any adjustments can still potentially be done.


uhbooh:
Sounds much better, nice job.

selliott512 (approved):
I like it. These sounds can absolutely be quieter. I did listen to them, and that seems to be the most obvious difference.

Issue #1253, opened 2024-01-01
patches: W108_1 is just a blank wall.


mc776:
so I tried loading Doom maps into Freedoom and:
![DOOM00](https://github.com/freedoom/freedoom/assets/24984517/ee0e9979-df7d-46f5-91c5-be024332bca4)

Expected: *some* kind of decoration or marker here, even if it's not necessarily a dead pain lord.

Marked as "deferred" since I'm willing to bet that somewhere someone has used SP_DUDE6 as a blank wall in some Freedoom IWAD map and it will take a while to check.

Talon1024:
> Marked as "deferred" since I'm willing to bet that somewhere someone has used SP_DUDE6 as a blank wall in some Freedoom IWAD map and it will take a while to check.

This is what I found:

    $ grep -Eri 'SP_DUDE6' levels/
    grep: levels/e3m6.wad: binary file matches
    grep: levels/map26.wad: binary file matches
    grep: levels/e3m3.wad: binary file matches

PR #1256, opened 2024-01-01
levels: remove inappropriate uses of SP_DUDE6.

mc776:
See #1253.

Also addressed #996 for E3M6.


uhbooh:
Screenshot_11
Actually I think this was an intentional use of SP_DUE6 in E3M3.

selliott512 (approved):
Looks good.

Locally I combined this PR with #1257, since they seem related, and I gave the changed levels a quick UV-MAX play through with IDDQD and IDFA without incident. I didn't notice any out of places "dude"s.

One small thing - for E3M6 you removed SP_DUDE6 from the front of linedef 388 which makes Eureka unhappy about a missing texture. Could you replace it with anything? Eureka defaults to GRAY1, but whatever. It's completely not visible.

PR #1258, opened 2024-01-06
Add more uncredited people.

unithecutie:
Credit Hayden49 for E2M2, and credit SUNPYG Senpai for E2M3.


selliott512 (approved):
The formatting looks nice. I'm going to approve this, but please make sure that that is how Hayden49 wants to be in the credits.

Also, I'm having trouble confirming these changes since we don't set the `Author` field in git. If someone could double check that as well it would be helpful.

PR #1260, opened 2024-01-07
Add jump and ouch sounds for Doom Legacy

Mr795:
I was just randomly thinking about this when i looked inside of the doom legacy wad and seen the sounds "DSJUMP" and "DSOUCH" then i just thought about making a PR for this, the jump sound is just a reversed version of DSPLPAIN and the ouch sound is just DSPLPAIN. (i'm wondering if this'll cause problems with vanilla compatibility or not)


mc776:
the DSJUMP sounds a lot less bad than I thought it would be.

Is it confirmed that Doom Legacy actually uses these though? I haven't used that port in 20 years.

mc776:
If we add these, I think BUILDCFG should be updated to add right after line 326:
```
; Additional player sounds used in Doom Legacy
DSOUCH	= dsplpain
DSJUMP
```
with no separate actual dsouch.wav file.

mc776 (changes requested):
needs buildcfg update; also since we're just reusing the existing pain sound we might as well link to it in buildcfg instead of reduplicating the file.

mc776:
I'm looking at legacy.wad and there is a *shitton* of stuff that's extremely specific to id in there. It would require a *huge* amount of stuff to get everything up to speed that bloats the download for the sake of what is now a relatively obscure sourceport.

I'm not sure we should proceed with this.

Mr795:
> the DSJUMP sounds a lot less bad than I thought it would be.
> 
> Is it confirmed that Doom Legacy actually uses these though? I haven't used that port in 20 years.

i've used it before and i do confirm doom legacy uses it, this is optional but ZDoom uses it as well if its in a wad.

Mr795:
> If we add these, I think BUILDCFG should be updated to add right after line 326:
> 
> ```
> ; Additional player sounds used in Doom Legacy
> DSOUCH	= dsplpain
> DSJUMP
> ```
> 
> with no separate actual dsouch.wav file.

after or on 326?

Mr795:
i think i might need to fix how the sound plays in the wad, christ it was put in a microwave.

mc776:
Had to look up where DSOUCH is used:
http://www.doomlegends.com/emporium/tutorials/specs03.html

player hitting head against ceiling... maybe the usual pain sound isn't that useful, how about an impact like the new DSOOF?

buildcfg.txt
```diff
@@ -436,6 +436,10 @@ DPSKESIT
 DPSKEATK
 DPRADIO
 
+; Additional player sounds used in Doom Legacy
```
mc776, line 439: this needs to be outside of the "if Phase 2" block.

Mr795:
> Had to look up where DSOUCH is used: http://www.doomlegends.com/emporium/tutorials/specs03.html
> 
> player hitting head against ceiling... maybe the usual pain sound isn't that useful, how about an impact like the new DSOOF?

i think that just gave me an idea.

Mr795:
> > Had to look up where DSOUCH is used: http://www.doomlegends.com/emporium/tutorials/specs03.html
> > player hitting head against ceiling... maybe the usual pain sound isn't that useful, how about an impact like the new DSOOF?
> 
> i think that just gave me an idea.

so i mixed dspunch and dsplpain to make a separate dsouch sound

mc776:
I notice the sounds were uploaded in 44100, I've downsampled them to 22050 here:
[grunting.zip](https://github.com/freedoom/freedoom/files/13852551/grunting.zip)

buildcfg.txt
```diff
@@ -324,6 +324,10 @@ DSSKEATK
 DSRADIO
 #endif /* #ifdef PHASE2 */
 
+; Additional player sounds used in Doom Legacy
+DSOUCH
```
mc776, line 328:
```
; Additional player sounds used in Doom Legacy and ZDoom
DSOUCH ; hit head on ceiling (Doom Legacy only)
```

mc776 (approved):
Approving this for now but let's see if anyone else has any input.

mc776:
Given the discovery of #1261 I take it people have been trying this out already, so I might as well merge this.

PR #1263, opened 2024-01-07
levels: make E2M3 skylight area exitable if entered.

### LINKED ISSUE #1261: E2M3: Out-of-bounds jump possible through sector 205
You can jump out of this window into a void.
image
One of its lines (maybe the inner one, so that you can't even jump onto the window) should be set to block players.

mc776:
Closes #1261.


Talon1024:
I looked in the editor, and there are some monsters in that void area. What are you going to do about those?

mc776:
Well, shit.

The lost soul height is 56. There is nothing that can possibly be done in vanilla that would guarantee the hatchlings would be able to get out but the player unable to get in, at least without making the entry point much, much higher.

In all the times I've played this map I've never actually been able to watch them coming out of that window, but the delay certainly affects the pacing of that encounter.

It'd be a lot of extra work to create a whole new connecting area... but I suppose an arbitrary teleporter down there, invisible from the main play area, wouldn't hurt...

mc776:
[e2m3.zip](https://github.com/freedoom/freedoom/files/13865808/e2m3.zip)

Will rebuild nodes later tonight but here's the revised fix with no block but teleport.

selliott512 (approved):
An improvement, but did you mean to make linedef 936 (the one between the window and the hatchling box outside) impassable?

I did verify prior to this fix it's not possible to rocket jump off of the non-floor. With this fix the teleporter worked as expected, but only after using IDCLIP to get passed linedef 936.

Needless to say the impassable makes it so the hatchlings are not able to leave their box, but it's still possible to get 100% kills by waiting at the window until they are exposed.

I've approved it for now, but if you wait a bit, like a day, I can do a quick play test of your final version (assuming you didn't mean to make the linedef impassable).

mc776:
I most definitely had not.

This latest version should be good now, tested both jump and hatchlings.

selliott512:
Looks good. I verified that the hatchlings come out now, and the teleporter still works. I did a quick play trough UV-MAX with IDDQD and IDFA.

PR #1264, opened 2024-01-10
Change a few map names

unithecutie:
Feel free to plug your own (much better) map names (or close the PR), these are all I could think of and I find them quite bad. :P

MAP06: Was changed in 0.12 to a new map.
MAP09: Space out "data" and "center".
MAP23: Have it go with the rest of Cluster 3 with this commit.
MAP24: Was changed in 0.11 to a new map - reusing mid-development MAP11's name.
MAP25: Changed because this name was somewhat based off of "Bloodfalls" from Doom II, according to the person who named it. It's quite bland as well. - reusing mid-development E2M1's name
MAP29: "Last Stand" is very bland and sticks out like a sore thumb against the rest of Cluster 3.


mc776:
I'm in favour of new names only where the map itself has changed or it's positively misleading, or it *really* sucks. Otherwise I'd much rather have more continuity so people can refer to the maps coherently when comparing them.

(That said, looking at the history of Map06 I think it's too late to worry about consistency for that...)

So, that said:
<nil> = I think it should stay the same
~ = I propose an alternative
+ = I approve of this change


~Map06: The name becomes clunkier without being distinct, and also doesn't address the big issue I personally have with this name for this map - that it doesn't look like a lab in any way.

Map04 of Nexus Credo, the original of the current Map06, is called Jade Tumult - which doesn't fit the naming theme of this cluster at all. The map overall feels like some kind of control center for teleportation logistics, somewhere necessary but not very busy and generally neglected.

I want to suggest something with the same initials but is techbasey - probably something J something Terminal. I kinda want to use a name of a real-world city but that's either too non-remote to be plausible or too restrictive about the implied story. The only other name that comes to mind is Janus, because of the two-faced look of the starting area.

So... Janus Terminal?


Map09: I like the single word, it's way more memorable than splitting it.


Map23: Strongly against changing this. Proposal is not a bad name but this is one of the few maps that has *always* been here under the current name.


+ Map24: Mild concern about repeating the name but "Whiplash" never made it permanently into anything so as the person who suggested that for Map11 to begin with I approve.


Map25: Not good because it was the final name for the old E2M1 before it got replaced, which creates more confusion if someone talks about the map later - especially if this were to be ultimately replaced.

This has already been changed at least once - "Blood Church" was even worse for the id-like-ness, even before going through all the concerns about using a similar name for the current Map27. I never made the Bloodfalls connection as far as *the name* was concerned until this PR, and the issue (and I think it's a minor one) is basically the entire aesthetic of the map.

Am strongly in favour of keeping this "Red Works".


+ Map29: **Yes.** (As this is the *pen*ultimate map, and you're on the offensive, and there's no central circle of defenders you're dealing with but guys spread out all over the map, "Last Stand" was always a terrible name for this.)

unithecutie:
The whole idea behind my Cluster 3 changes was to give them more unique names, for it being the final cluster.

MAP06: "Janus Terminal" is quite a nice name! I'll add it to the PR.
MAP09: I wanted this to stay somewhat more grammatically correct, but meh. :P
MAP23: "Underground Facility" is really odd among Cluster 3's level names, it sounds like it belongs in Cluster 2.
MAP25: The map itself is fine in my opinion, "Red Works" also sounded a bit odd to me - was "Surreality" previously known as "Ruins" or was it a completely different map? Somewhat the same situation as "Whiplash" here.
MAP29: Yes, this really needs to be changed.

unithecutie:
Hmm, again. I wanted Cluster 3 to have a more distinct naming vibe to it (coming from things like "Remanasu" and "Grim Redoubt").

"Aberration" is the fact the base there looks very out of place (syn. _deviation_) among everything else in that environment, it's also somewhat "corrupted" in a sense.
~~"Whiplash" came from the fact that MAP24 is very back-and-forth.~~
"Surreality" for MAP25 was coming from the MIDI in that level and how it has quite a strange layout.

"Abandoned Teleporter Lab" sounds a lot like it would belong in Cluster 2 again.

mc776:
I can kinda intuit what you mean but that "vibe" is, if you look at the entire cluster, actually an exception to the names list

(and frankly I hate it and it's one of my least favourite things about Freedoom)

Blood Ember Fortress - descriptive name of place
Under Realm - descriptive name of place
Remanasu - nobody knows where this name comes from, it was from the donated map, deemed to be "the vibe"
Underground Facility - descriptive name of place
Tertiary Loading Bay - (not very) descriptive name of place
Red Works - descriptive name of place
Dark Depths - descriptive name of place
Palace of Red - descriptive name of place
Grim Redoubt - descriptive name of [place](https://www.merriam-webster.com/dictionary/redoubt)
Last Stand - "the vibe", but also we both agree it needs to be changed
Jaws of Defeat - originally "the vibe" but everybody parses this as JoD being the name of the final boss, so it's retroactively transformed into a descriptive name of place

But the problem is that the actual source of "the vibe" is not in Freedoom's progression in and of itself, but the expectations we've been trained on thansk to id: https://doomwiki.org/wiki/Doom_II#Levels (and even there we have a distinct exception in Map26)

Personally I'd rather we got rid of "the vibe" entirely and renamed Map22 (along with my proposal for Map24 as "Abandoned Teleporter Lab" and your proposal for Map29 as "Melting Point" which don't affect the status of "the vibe" but merits renaming for other reasons).

unithecutie:
> Redoubt

_Didn't even know it meant that until now, lol._

I honestly have no clue where this PR is going at this rate.

mc776:
I think at least we both agree re: Janus Terminal and Melting Point.

And given the history of the former, I'm tempted to open up the possibility of renaming Map22, but that seems utterly extraneous and unnecessary.

The two "unders" in Map21 and Map23 bother me when I look at the list though. Gonna check of Map21 ever had a different name or the name was for a different map...
[Answer: yes *but* the current map author specifically asked to keep the name](https://github.com/freedoom/freedoom/commit/fa5e756bc1adaaf344c9ef904adcb4e89b15a290). I'd say leave them both as is.

unithecutie:
MAP21's name was always some form of "Under Realm".

unithecutie:
More people commenting on this PR would be appreciated. :P

MerrittK:
Logic gate could be changed to Logistics Hub

unithecutie:
> Logic gate could be changed to Logistics Hub

Keeps part of the original name intact while having a more fitting name for the cluster.

MerrittK:
It was Outer Storage Warehouse until recently, so Logistics hub makes sense without having to look like anything in particular

unithecutie:
I renamed it to "Logic Gate" as that's all I could think of.

The main point of this PR was Cluster 3, going for a more obscure, unique set of names for the final cluster.

MAP20: Blood Ember Fortress
MAP21: Under Realm
MAP22: Remanasu
MAP23: Underground Facility --> Aberration
MAP24: Tertiary Loading Bay (old MAP24 level name) --> Whiplash
MAP25: Red Works --> Surreality
MAP26: Dark Depths
MAP27: Palace of Red
MAP28: Grim Redoubt
MAP29: Last Stand --> Melting Point
MAP30: Jaws of Defeat

MerrittK:
I don't particularly know what under realm is supposed to be, in the map editor it seems pretty disjointed.

unithecutie:
The level was replaced, but the name wasn't. Load 0.12.1 and that's the map it's based on.

MerrittK:
Phlegethon?

MerrittK:
Or Counteroffensive, because it's the first hell level.

mc776:
I strongly dislike where this is going.

mc776:
Logic Gate is a great name for the structure of the map, which "Logistics Hub" does not refer to in any way that could be derived if you didn't know the previous name already, and which the map doesn't resemble more than a warehouse.

Map21's name had been specifically requested by that mapper.

I've said my piece about Map23-25.

Honestly the longer this goes on the more I'm inclined to just close this.

mc776:
> but this would call for Cluster 2 to be renamed as well.

Who even are we to be doing this

PR #1267, opened 2024-01-11
Rename levels as discussed in previous PR

unithecutie:
Use @mc776's names for MAP06, MAP24 and MAP25 and use the name I suggested for MAP29, as discussed in #1264.

"Persistence of Memory" started to sit well in my head - it gave me the same feeling as the name of the old MAP27, "Warped Elementality".

MAP06 and MAP24 were never renamed when the maps were changed, and as discussed before, MAP29 is more on the offense than the defense.


mc776:
Yeah, my closing that PR wasn't a super-hard-no, just my response to the scope creep that was happening in the discussion.

I was thinking about the renaming last night and I'll throw this out there:

Map29: Lake of Fire

It's literally what it is, is appropriately apocalyptic/catastrophic for a map that looks like a bunch of old buildings slowly crumbling into lava - and oddly enough as an explicit biblical reference *not* that evocative of id:

- [D2, cluster 3](https://doomwiki.org/wiki/Doom_II#Levels) is a mishmash of different spirit stuff (Spirit World, Nirvana) that seems almost deliberately avoidant of anything in Christianity - with the sole exception of *Icon* of Sin, but even that is very clearly the product of an iconoclastic American Protestant cultural Christianity that thinks of icons as a pagan corruption of the faith.

- In [Episode 4](https://doomwiki.org/wiki/Thy_Flesh_Consumed#Levels) we see an explicit set of Biblical references, but what they really are is a bunch of not even phrases so much as mere *sequences of words* arbitrarily plucked out of the King James to give a doomy vibe. Not a single one even remotely describes the physical place; the closest it gets to actually referencing the content of the scriptures is "and hell followed" being an obvious reference to the four horsemen

I still prefer "Red Works" in part because it and the fluids theme reminds me of [Steamworks](https://steamworks.com/) but that one's just me.

unithecutie:
> explicit biblical reference

Weren't we trying to take references of religion/_Hell_ out of Freedoom or is my memory just bad?

mc776:
I think some folks are way too zealous about that when it comes to certain things in FD.

Like, the inverted crosses and doomy symbols, yes, but sometimes I see people objecting to things that are mere religious references that would work independently in their own right.

This level is a large pit filled with lava and filled with monsters near the end of the game.

unithecutie:
> a map that looks like a bunch of old buildings slowly crumbling into lava

I see something more like a set of temples.

mc776:
The image of the old gods being finally destroyed in a lake of fire plays to a lot of Christian tropes, though I guess that could actually tie it back to the presumed-iconoclasm of Doom's mythic tropes.

The image of many gods being destroyed in the final fire made me think of something Norse but nothing really fit.

Looking at other pagan ideas... ["Ekpyrosis"](https://en.wikipedia.org/wiki/Ekpyrosis#Ekpyrosis) might be worth considering:

> This *ekpyrôsis* is not so much a catastrophic event, but rather the period of the cosmic cycle when the preponderance of the fiery element once again reaches its maximum. All matter is consumed becoming completely fiery and wholly soul-like. God, at this point, can be regarded as completely existing in itself.

mc776:
While we're here, can we change E3M2 to "Geothermal Tunnels"?

It (back when it was E3M3) was changed from "Tunnels" to "Temple" here https://github.com/freedoom/freedoom/commit/7f10eb9ff3f909e3f32588f98be6262d43bb52b8 but no rationale was ever given nor anything remotely resembling a tunnel added to the map.

mc776:
E1M7 was also [changed](https://github.com/freedoom/freedoom/commit/2cc7a36fa89cb8a419191919d90730ef803a4be2) for no apparent reason in a way that denigrated from the name corresponding with what the level actually resembled.

The open-air pits don't make much sense for a virology lab either, so it should be something relatively inert.

Anyone like something like "Paraphysical Materials Lab" or "Xenobiotic Materials Lab"?

mc776:
One more for consideration: reverting Map21 back to "The Under-Realm".

I have no idea why this was changed, because the intention clearly seems to be something referring to some kind of underworld, but with the missing punctuation and article it's always easy to initially misparse as "under a place called Realm".

unithecutie:
> I have no idea why this was changed

Note that it was also changed along with "The Blood Ember Fortress" to just "Blood Ember Fortress" - seems to be a stylistic choice to have it match the other levels.
Would an entirely new name for this map (MAP21) be warranted, despite the map author's directions?

mc776:
There was never really any principled reason to get rid of all the "the"s in the first place, though at this point this would make Map21 unique. (Note that it's not unique in LibTech1 where I've already named Map29 "The Lake Of Fire")

The loss of the hyphen, however, just seems like one of Voros' odd little ESL hypercorrections that slipped through only because no one thought it was worth reverting.

mc776:
> map author's directions

Oh, right, Mort only added this map years after the change. Might as well leave it for now...

mc776 (approved):
I'll leav

mc776:
It's been a couple hours.

Just a heads up for future reference, unless there's an actual conflict there's no need to re-sync your branch - once the PR is merged it will merge only the changes you had made.

unithecutie:
> Just a heads up for future reference, unless there's an actual conflict there's no need to re-sync your branch - once the PR is merged it will merge only the changes you had made.

Ah. :P
I'm still fairly new to GitHub, thanks for telling me.

PR #1268, opened 2024-01-11
Swap DM23 and DM27's tracks.

unithecutie:
"Shells" is a lot faster and more suited for the small, duel map that is DM23.
"White Lights" is more grandiose (why it was removed in 0.12 in the first place, didn't fit E1M2 at all) and fits the very large, arena/fortress-like DM27.


mc776 (approved):

mc776:
I'm inclined to agree but any thoughts from @Ospaggi?

mc776 (approved):
Oops, forgot I still had an old build installed. That wasn't "Shells" I was listening to the first time I approved this.

I am very inclined to agree.

Ospaggi:
ah I'm fine, go ahead

PR #1269, opened 2024-01-12
Non-Hellish/Alien themed SPINE texture

UltimateCraneo:
Instead of a exposed spine, it's now a Trilobyte nest.


mc776:
On the one hand, I do appreciate the effort to de-Doomify assets.

On the other, I've got multiple concerns with this one:

1. This texture isn't necessarily associated with trilobites/cacodemons and might not necessarily make sense in their absence.
2. The big, wide, sparsely-placed bubbles would make very minor misalignments extremely jarring where they wouldn't be noticeable or would even be a correct-looking variation when using the current texture.
3. The bubbles wouldn't make sense on a convex or concave surface.
4. The bubbles are large and are shaded/filled as to look like they're supposed to stick out enough that the curve should be visible from the side, which makes it jarring when they're not.
5. It looks a little too plain compared to what it's replacing.

UltimateCraneo:
Yeah, that's fair, kinda wanted to see if this was good to be added, but those reasons I can't argue with.

uhbooh:
I think these textures could be included separately from the spines. They do look really nice and fit the alien theme.

mc776:
It definitely has promise as its own texture, though I think it would work better with more detail - wrinkles and veins in the skin, more smaller densely packed bubbles, maybe they form patterns in the skin... which could sometimes overlap or have little flaps and imperfections so the bubbles aren't perfectly round...

UltimateCraneo:
how do I delete the spines lmao, wanted to just upload the Cascade props idk how to use github still

unithecutie:
Quick word of advice: making branches of your fork will make pull requests a lot easier.

PR #1270, opened 2024-01-12
levels: make E1M7 yellow key platform obvious.

mc776:
Ran past this and got lost again.

The grey of the pillar just blends in too well with the wall behind it, while the height takes it out of the player's visual area of attention.

The pillar is given a texture that stands out and looks like something that should possibly move, while its base has been given the same markings as the switch that lowers the pillar. The pillar is also extended so that the yellow key is visible from the blue key (and if you can somehow jump to reach that key then more power to you - it should be an even harder jump than the old one.)

Before:
DOOM00

After:
DOOM01
DOOM02
DOOM03 

The switch, for reference:
DOOM04


selliott512 (approved):
Looks good. I'm all for preventing non-fun key hunts. It's both aesthetically improved and more functional. 

I did a quick UV-MAX test. Eureka is happy.

selliott512:
And there are no obvious structural changes other than the key.

PR #1271, opened 2024-01-12
Flip E2M3 and DM06's tracks

unithecutie:
As I said when I initially shuffled E2's tracks to realign them with the map slots Korp had intended;

"Tunnel" feels out of place among E2's tracks and a lot more suited for the sandy, rougher, Egyptian temple-styled DM06. The feeling is coming from the harpsichord and the more trashy... metal... thing that it is.

"Workout" is a lot more upbeat and fitting (techno-type beats) for the first few E2 tracks - [Korp had originally intended it for E2M1 also.](https://www.doomworld.com/forum/topic/124732-korps-music-thread-or-smth/)


mc776 (approved):
I was not expecting this to improve my experience of E2M3 as much as it did.

PR #1272, opened 2024-01-12
[Draft] E1M5 tweaks, pull request for discussion.

MerrittK:
Here is a version of E1M5 with a slightly less miserable start, the rest of the map is unchanged.


MerrittK:
Also changed the .gitignore to ignore .dbs files,

mc776:
Miserable start? Are we talking about the same E1M5 (5, not 1), Armory, where even with a pistol start you get a shotgun, backpack, ripsaw and a bunch of healing?

MerrittK:
Yes, the one were you spawn down the hall from hit-scanners, in cramped quarters.

MerrittK:
You don't get armour, there is barely any texture or height variety, the level is both too cramped and has too much empty space, and it makes continuous play of E1 an incredibly unfun experience.

mc776:
There's plenty of cover there if you don't charge right down the middle. The health bonuses are there to give you a reason to go that way.

I opened the map thinking I'd see something that would cause me to reconsider and reopen this, but I'm sorry but there's nothing to fix here, and the fixes added on this PR are full of misaligned and missing textures and things sticking out looking very out of place.

mc776:
It's one of Phase 1's easier maps if you actually treat hitscanners as a threat and learn to enfilade with the shotgun.

MerrittK:
I'll fix it and try re-submitting. It wasn't completed, I just needed some feedback.
Frankly though, the level is bad. The only way to complete it is to camp doors and wait for shot-gunners to toddle out, the elevator doesn't give you any good options, and a third of the level can, and should, be skipped.

mc776:
It's one of my favourite levels in Phase 1.

mc776:
The biggest change you'd made to that first area, strictly from a gameplay point of view, was to add more cover. That could have been done with literally one more crate down that open area. Instead, it's this ham-fisted mutilation of the entire architecture of that area for a massive aesthetic downgrade.

By the time you reach the elevator you have the minigun. You need to think fast, act fast and optimize cover.

The wide open corridor in the start allows a lot of distance to minimize your chances of getting hit while you line up your shot. For the most part the guys are hitting each other rather than the player.

Zombies take a certain amount of time to shoot. You can use that time to move between two sources of cover, or even pop in, shoot, and pop out while your shotgun plays its racking animation while the surviving zombies with line of fire on you finish flinching in pain.

The biggest flaw in this level is the boredom of backtracking to the red door once you're "done" - there are ways to fix that (but I would advise in favour of marking any such proposal as deferred).

A lot of map balancing is a judgment call between whether something is actually pointlessly artificially difficult and boring, or if it's just geared towards a different play style. There's lots of stuff in FD that doesn't match mine and I might have overdone a few, but I've at least tried to see what I've been missing.

unithecutie:
I don't find it at all difficult, I like the sort of "shooting gallery" style.
Play slow and learn to hit multiple things at once with the shotgun, and you'll be set.

On the odd chance someone plays Freedoom before the original Doom, this will also teach then about how to avoid hitscan damage more reliably.

MerrittK:
I'm fine with it being turned down, but I don't like how quickly the pull request was closed, then locked. My work was sloppy, but I marked it as a draft and was planning on updating it again tonight.

I've also seen other contributors have their work immediately binned based on only a single opinion.  

This'll be my last attempt at a submission for a while, I need to move on to less stressful projects.

PR #1274, opened 2024-01-14
New MAP11 track.

unithecutie:
Korp wanted "Inside the Ruins" replaced.

In their words: 

"im thinking of submitting this one actually, i havent checked current map09 so im not sure where it would fit best
im thinking map11
i never liked ruins"


mc776:
I gave this a run on both Map11 and Map09 in chocolate.

It *eventually* started to grow on me as far as Map11 was concerned, but it still feels like a noticeable downgrade from Ruins.

I'm surprised how much better I liked it on Map09 than the current "Terrible Secrets" (which was already decent for that map).

KorpKat:
It's too "Jungle-y" on the instrumentation side to be on MAP09 IMO, That's why i've requested it to be on MAP11 since the map already looks ancient and full of vegetation.

unithecutie:
I don't think this should be moved to MAP09, "Terrible Secrets" fits the grungy warehouse map that it's in.

For future reference, listening to these tracks with a GM (specifically Roland GM) more will give you a much better idea of what the authors of those tracks are going for.

mc776 (approved):
Giving this a longer listen and playing Map11 in both again.

Four points stand out:
1. Map11 is a very long map.
2. Melodically Ruins is far more instantly memorable.
3. The strongest thing about Ruins is the intro.

Point 2 actually kinda works against Ruins because point 1 means anything repetitive about it will sound a lot more repetitive.

If anyone really wants Ruins to be somewhere else, please speak your mind *ideally* before this PR gets merged later, although I don't suppose there's anything stopping someone from proposing this for another map in a future PR.

mc776:
> For future reference, listening to these tracks with a GM (specifically Roland GM - the default Windows soundfont) more will give you a much better idea of what the authors of those tracks are going for.
> 
> [Dropbox folder with these tracks converted to MP3 in Roland GM](https://www.dropbox.com/scl/fo/lk91dnv4g3zz2xj94z8jy/h?rlkey=mviln4fwqg8qqlsh3a6r4ahgp&dl=0)

My normal setup is Timidity++ with a really old copy of SGM-V2.01, which doesn't sound too off from these MP3s (except the guitar for Terrible Secrets is way better on SGM). The new track *definitely* sounds better this way, though I had intentionally avoided it until you reminded me just now because we should initially check these things at their worst because if there's any problem that's going to be where it's at.

(That said, listening to both the old and new Map11 tracks properly now doesn't fundamentally change anything in my assessment when I approved this.)

PR #1282, opened 2024-01-17
patches: get rid of BFALL black pixel.

mc776:
DOOM00


selliott512 (approved):
Looks good. I did compare them carefully. You only got rid of the black pixels.

PR #1283, opened 2024-01-18
Looping for "The Raging Sun"

unithecutie:
This is much better from a gameplay standpoint.


mc776 (approved):

mc776:
Bit of a side question, but does anyone know how to get Timidity to create an OGG without a huge silence at the end?

unithecutie:
This isn't a bug, the original MIDI is just like that.


I use Foobar2k to convert MIDIs to OGG.

PR #1284, opened 2024-01-19
Restore "Pit" by Ajanddino as DM04

unithecutie:
MFF7 doesn't work well as a deathmatch track, it's _way_ too slow.
I chose "Pit" as it's fast-paced and was pretty short-lived.

(remind me to fix the draft thing)


mc776 (approved):
I have no strong feelings either way though in general I like "Pit" slightly better. Will leave open for a bit in case anyone begs to differ.

PR #1285, opened 2024-01-21
New tune for E1M7

Josephus-DH4050-Astartes:
my new song "Totally Surrounded" for E1M7.


Josephus-DH4050-Astartes:
> This does fit E1M7 a bit better than my tune (which was intended for a different map that was, rightly, removed.)

Why thank you! Yeah, it would've been nice in a labyrinthian map, i didn't replace it because i didnt like it, I love it! I just wanted to make another tune for freedoom :>

mc776 (approved):
If even the author of the current track has given their blessing I suppose I should approve this.

mc776:
@northivanastan can we add "S: continuum.mid" to the credits file (given you've still got a couple other tracks in FD)? I didn't even realize you were the same person until now 😅

unithecutie:
Yeah, about that: I intentionally didn't place his real name there as it was removed on other platforms ([see here on Doom Wiki](https://doomwiki.org/wiki/Doom_Wiki:Information_removal_requests#continuum.mid))

northivanastan:
> @northivanastan can we add "S: continuum.mid" to the credits file (given you've still got a couple other tracks in FD)? I didn't even realize you were the same person until now 😅

oh yeah, definitely should do that. I have a ton of different usernames online and it would be impractical to include them all, but in the Doom community I'm going by continuum.mid now.



> Yeah, about that: I intentionally didn't place his real name there as it was removed on other platforms ([see here on Doom Wiki](https://doomwiki.org/wiki/Doom_Wiki:Information_removal_requests#continuum.mid))
> 
> Speaking of that page; I don't remember what Picklehammer said on LinkedIn in full but I'm pretty sure he was fine with keeping in Freedoom's text files, just not Doom Wiki.

My real name is already out there, it's on this account, and when I said the thing that prompted that DoomWiki removal I was going through some stuff. Not that I want to reverse that removal request, I still have complex feelings about that, but I'm not demanding that my name be removed everywhere. It can stand here, and we can add S: northivanastan, continuum.mid.

PR #1286, opened 2024-01-21
graphics: use dehacked for skill+episode names.

mc776:


mc776:
One thing that's missing from this is the ability to specify which dehacked lump to use.

How do you pass that IWAD environment variable to the script? Or is it impossible and that's why we never see it done here?

mc776:
Removing the "deferred" flag since I've confirmed it does build correctly and this PR by itself has no effect on the final IWADs that are produced.

selliott512:
I'll try to look at this later today or tomorrow.

selliott512:
I'm all for reducing redundancy by having one copy of data, so I like the idea. That being said I did diff the WAD files after making the version static and I saw that they were different with this change. Digging deeper lump `M_HURT` was different. In master:
```
  graphics/text/config.py
    "m_hurt": "Bring on the pain."
```
In this fix you get it from the lump, but in that case some of the letters seem to be upper case. In this fix the above is effectively (once the `read_bex_string(freedoom_bex,"SKILL_NORMAL")` is resolved):
```
  graphics/text/config.py
    "m_hurt": "Bring on The Pain."
```
Or, to put it another way, grepping for `Bring on The Pain.` case insensitive in current master:
```
git grep -i 'Bring on The Pain.'

graphics/text/config.py:    "m_hurt": "Bring on the pain.",
lumps/dehacked/dehacked.txt:SKILL_NORMAL = Bring on The Pain.
manual/freedoom-manual-fr.adoc:| 3 | **Bring on the Pain.** | Niveau de difficulté par défaut.
```
Is one better? Do you want to change the case that way?

I have not looked at the other differences.

mc776:
It's only the difficulties not the episodes.

Given how I pronounce it in my head "on the" in "bring on the pain" should be either both capitalized or neither. Since what was in `config.py` is what the end user actually sees I've changed the deh to match that.

selliott512:
Good. Closer. One more though. For nightmare if you want them to match by continuing to edit `lumps/dehacked/dehacked.txt`:
```
-SKILL_NIGHTMARE = Mayhem!
+SKILL_NIGHTMARE = MAYHEM!
```
Once that is done I've verified that the WADs built are exactly unchanged.

selliott512 (approved):
Looks good. The WADs match perfectly now. 

One concern I had was that it might break if someone were to try to build the graphics independently without first building the dehacked stuff, but that dependency is explicitly noted in `Makefile` as is: 
```
subdirs:
        $(MAKE) -C lumps/dehacked # graphics/text depends on generated dehacked files
        $(MAKE) VERSION=$(VERSION) -C graphics/text
```
So that's not a problem.

I can't think of any way that this would break. Feel free to merge it now, if you want (for 0.13.0, whenever that actual happens).

mc776:
Might as well not leave this hanging.

The text graphic generation script already relies on dehacked for the level names, which is why it occurred to me to do this at all.

PR #1287, opened 2024-01-22
Fixed Octaminator Sprites

superdave938:
removes leftover red pixels on the tan parts of the Octaminator


selliott512:
For future reference most Freedoom files are all lower case. As it was it looked like to GitHub that you were adding new files instead of modifying existing ifles. I took the liberty of modifying your commit so that it has lower case file names, but I did not otherwise change it. I also set the commit message so that it wasn't just `Add files via upload`.

superdave938:
Yeah, I tried using GitHub Desktop this time

It was confusing to say the least

mc776 (approved):

PR #1288, opened 2024-01-25
flats: remove pure black bit from FLAT1_1.

mc776:
flat1_1_oops


selliott512 (approved):
You indeed fixed the black pixels, but it looks like you made it a bit darker, particularly the brown regions, a change that persists after running both the before and after through deutex (it's a real change). I've approved it, but please check if you meant to do that.

mc776:
Fixed.

mc776:
And yes I did buld and run it this time:

Original
DOOM00

Built with my initial "fix" which might have gone bad after I'd run it through Irfanview and might have accidentally hit a gamma setting somewhere
DOOM01

Built with a re-do based on an extracted original flat from SLADE, done in Kolourpaint as usual
DOOM02

ShovelCode:
Great.

uhbooh:
Nice job

PR #1289, opened 2024-01-25
Retune MAP29

unithecutie:
Might be a few errors, I haven't edited a map before.

This map is still stupidly difficult, even with the prior edit. Amarande made sure to make this map infuriating, it seems.


mc776:
Can we get a summary of what was changed?

(And should we add a "gameplay tweak" label for PRs and issues like this?)

unithecutie:
> Can we get a summary of what was changed?

The difficulty is tweaked from infuriating to challenging. I placed specific details in the commit message, but I'll add them here (it also seems I forgot one tweak)

```
Halved the damage of the damaging floors in the map 
Removed the damaging floor in the room where you lower the yellow key completely.
Removed the torches around the first teleporter to enter the red key arena to make running into it a lot easier.
Replaced the Baron on the initial lift with a Hell knight on hard, a few chaingunners on medium, and completely removed it on easy.
Replaced the pain elementals in the cramped blue key room with cacodemons.
Made the blue key lift not lower when you step on it.
Removed the damaging floor in the blue key room.
Lowered the yellow key platform to make it easier to hit the arch-vile.
```
> (And should we add a "gameplay tweak" label for PRs and issues like this?)

I think `enhancement` would suffice?

mc776:
Was this map originally intended to work with ports that allow jump and mouselook? I wonder if these changes might actually take the map back closer to what the original mapper intended...

unithecutie:
Let me ask Amarande real quick.

unithecutie:
> jump and mouselook

The map is certainly a lot easier with these, but this map was submitted ~2001-2003 if I recall correctly.

selliott512:
That feels like a lot of last minute changes (we want to get 0.13.0 out very soon, like Sunday), but I see that you rebuilt the nodes, and you didn't introduce any Eureka errors, which is great.

mc776, would you like to do this for 0.13.0? If so I can probably give it a quick play through tomorrow. Depending on how much time I have I may or may not use the IDFA and IDDQD cheats, so a UV-MAX check rather than for difficulty.

Xindage:
I'll gameplay review it tonight then ill come with a commentary about the gameplay.

mc776:
**elevator**

Called the lift and got promptly flattened by the *four* minigunners lol

I think it should have a painlord on hard, painbringer on medium and a couple serps on easy - just all standard projectile monsters.

The monster's original purpose was to create a delay that exposes you to more pinkies and fireballs so you can't just hide out in there - this changes the entire encounter from a variably-timed area denial to a front-loaded single high-risk encounter with a fairly quick payoff (if you survive it) of substantial safety from being surrounded by the previous enemies.

(Changing the nature of this encounter isn't a problem for me in principle, but what is a problem is that this encounter is a jumpscare with almost no telegraphing - which works for the slow baron with its slow attack, but when it's more quickly lethal like this it just feels like one of those things where the expected play is to die and then learn the trap and breeze through it - not good for replay. It should also be noted that the minigunners will start shooting and hitting you long before you can see them, whereas a projectile monster would be blocked, so you really have no visual clue until you're already taking damage (and the clue is the taking damage).)

**jump platform**

This definitely plays better but it looks kinda plain (especially with the new grey rock texture in contrast with the other high-contrast rocks) - and *mostly invisible* from the tower especially once you're holding the minigun or the polaric.

I would like to suggest:


**blue key**

Agreed with not making the blue key platform move - that's exactly the sort of untelegraphed gotcha I was just complaining about with the minigunners. The giant platform looks really excessive though, and aesthetically feels really out of proportion for that spot and that item.

I think I like having a matribite here to create urgency and draw fire away from any surviving serps instead of just "oh look it's more enemies. yay" (but not so OP it feels like a shitty gotcha either). But having four of them really was uncalled for. Suggestion: make the eastern 'bite a matribite - that's a spot that's right on the edge of the player's vision when they warp in but it keeps the initial appearance of the room look nice and symmetrical.

**necromancer and yellow key platform**

Is it just me or is it *way* more aggressive now? Before this change it would often not attack even despite having line of sight but now I'm seeing flames pretty much anytime I get in LOS. Is height factored into monster targeting?

mc776:
One possibility: just make it really easy to reach the blue key without even fully exiting the cavern:


(my original idea was to make the platform so high you can't even go out the window but it didn't look as good)

mc776:
Did the SSG hurtfloor get neutralized as well? Good call.

unithecutie:
> Called the lift and got promptly flattened by the _four_ minigunners lol

My idea was to nuke them with the super shotgun, but I'll change them to imps.

> jump platform

Looks a lot better, I'll add it to the PR.

> necromancer and yellow key

Strange, it was always this aggressive with me before the change. 

> I think it should have a painlord on hard, painbringer on medium and a couple serps on easy

The Baron was stupidly hard to kill in the tiny space of the lift, that's why I changed it to a Hell knight.

> make the eastern 'bite a matribite

Will do.

mc776:
I forgot about the archvile maximum range. Disregard my comments about aggressiveness.

I just noticed that if you've got a missile launcher at the start of this map you can take out the necromancer before you even leave the starting room. Should this be allowed?

unithecutie:
> I just noticed that if you've got a missile launcher at the start of this map you can take out the necromancer before you even leave the starting room. Should this be allowed?

This was always possible, but you don't get a rocket launcher from pistol start.

mc776:
I'm inclined to approve at this point but let's wait for Xindage's review first.

selliott512:
I've give this a quick cheat code completability (100% kills and secrets) later today. Xindage review will probably be more interesting.

Xindage:
this map is still hard af, i cant beat it yet, what i dislike most is the stupid fights that i'm given, like a monsters full of chaingunners that i can completely leave and, telling me to use bfg because i cant stay outside and them die because it cant clear all at once.

Also after few tried (like 10) it gets boring i really hate to have to punch 50 or so demons because the ammo is kinda tight at the start and we have no safe zone to rest a little, there's always something coming at you and the archvile... is the cherry on top just to him i died like 4 times.

selliott512:
I just ran through it today, with commit 39f8b52e22dbabee5d941e39f71e91e010b33d81 and I was able to get 100% kills and secrets with god mode and unlimited ammo.

No Eureka errors still - good.

Here's an animated GIF showing the changes

map29

I don't have strong feelings about this. I don't have a good sense of the change in difficulty given the way I played it. I'll approve it to make it easier for one of you to merge if you want.

Keep in mind that we (hopefully) have a release tomorrow. Try to be extra safe with any further changes, or defer it.

selliott512 (approved):

mc776:
> this map is still hard af, i cant beat it yet, what i dislike most is the stupid fights that i'm given [rest of post omitted for brevity]

I don't think this can really be addressed without *thoroughly* revising the map beyond recognition. (Personally I appreciate the existence of totally optional ultra-hard stuff)

I don't think there's anything further worth tweaking that isn't definitely of the not-safe-and-should-be-deferred type (cf. the various changes I'd made on LT1), so I'm going to go ahead and merge.

PR #1290, opened 2024-01-26
musics: Restore "Paranoia" by Picklehammer as DM10

unithecutie:
The current DM10 track isn't held in too high regard and was suggested to be replaced at one point.
"Paranoia" is a fantasy-action-metal thing and it fits deathmatch really well.


mc776 (approved):
I have no strong opinions either way (actually split - I like Paranoia more overall but the current DM10 I think works well for deathmatch) so I'll leave this open pending a possible second/third opinion.

selliott512:
I listened to both and I also don't have strong feelings. They are both nice songs to me.

In `CREDITS-MUSIC` IMHO we don't need `(moved to DM10, not removed)`. We don't need the edit history in there. Other music has been moved around, which we don't note.

mc776:
The other moves have been noted, actually - it's only noting the changes made since last release.

PR #1292, opened 2024-01-27
Replace E1M1's track

unithecutie:
Continue what was done in 0.10 and replace all of the vanilla Doom remixes.
This track should've been replaced a while ago.

While Ralphis' remix is a good E1M1 track and a good remix at that, the "remix" part is the problem.
"Crawl" is a medium-paced metal/rock track that helps set the mood at the start of the adventure.

This should be left open for further discussion, as I can't think of anything that fits E1M1 better other than "Crawl".


mc776:
On the one hand, I'm the last person to disagree with the need for this in principle.

On the other, "Crawl" fall far, far short of eitherthe current E1M1 or "Not My First Rodeo". It's not particularly inspiring of anything and feels like something from the middle of the first cluster.

Other possibilities:
- Varis Alpha's old Map29 track can work really well for a ramping-up sort of song.
- ..."Warpath" (the old P2 drum solo intro) could also work for the same reason.
- I've used "Stanky Leg Specialist" in LibTech1 and frankly I think it works better for the upstream E1M1 than LT1's heavily modified, more prison-like one.*
- I'm probably only thinking this since it had just been decommissioned and it's still fresh on my mind, but "Derelict" might actually work as E1M1 is sprawling and relatively quiet and you're supposed to be looking around trying to work things out still.
- ...but "Animal Skin" might work better for an episode start if we go in that direction.
- I'm almost tempted to seriously suggest "Fountain Dance" since it would create a leitmotif in common with the intro.
- ...ok but what if we used "Paranoia" so that the very final map of the entire combined Freedoom singleplayer campaign shares themes with the very first one*

*If we swap the tracks the current E1M1 track shouldn't be too noticeably id-like-in-a-bad-way if it's heard in some random deathmatch level.

mc776:
Here's a zip of all the tracks I mentioned plus Crawl for reference:
[d_e1m1.zip](https://github.com/freedoom/freedoom/files/14073212/d_e1m1.zip)

uhbooh:
I think "Stanky Leg Specialist" is the best song in that zip for E1M1. Crawl.mid is a nice doom style song, but sounds more fitting as a mid or late game level.

unithecutie:
"Stanky Leg Specialist" could work, actually!
I'll need to replace DM11's track though from doing that.

mc776:
On the one hand, Ralphis' track is one of FD's "iconic tracks" that should get some use somewhere.

On the other, I am surprised at how *well* "Death's Embrace" works for that DM map.

unithecutie:
Picklehammer's "fantasy+rock/metal-thing" style fits deathmatch really well, that's why I chose it.
My only reason for calling for Ralphis' track's removal is what we did in 0.10 by purging the Doom remixes.

Some of the Doom II remixes were in Phase 1 too, and they were purged.
The tracks as they were in 0.9:

E1M1: "At Doom's Gate" (a different remix at this time, still by Ralphis)
E1M5: "Suspense"
E2M4: "Deep into the Code"
E2M6: "Suspense"
E3M1: "Bye Bye American Pie"
E3M2: "Suspense"
E3M3: "Deep into the Code"
E3M6: "Evil Incarnate"
E3M8: "Waiting for Romero to Play"
E3M9: "The Healer Stalks"
Phase 1 intermission: "Bye Bye American Pie"
Phase 1 bunny: "Bye Bye American Pie"

mc776 (approved):
Would be good to have a few more pairs of ears on it but I'm not aware of any reason why this shouldn't be merged before release...

selliott512:
The existing track for E1M1 sounds pretty good to me, but I don't have strong feelings about musing, so let's wait, unless someone else wants to do it.

mc776:
My original intent had been to merge this shortly after midnight PST which should have given folks enough time, but I was unusually tired and went to bed at 11 having forgotten the whole thing.

We've probably given more time for this one than any of the recent shuffles and no one's suggested anything else or insisted that the current d_e1m1 is sufficiently dissimilar, so I think we might as well merge it.

selliott512:
> so I think we might as well merge it.

is good enough for me. Done. Now it will be in 0.13.0.

Mr795:
I remember a use of ralphis' remix in dm13, is that still used there or is it replaced?

unithecutie:
It was replaced by "Shut it" by Korp.

Mr795:
Ok, just checking. The remix was actually used as dm13 in 0.9, before e1m1's use of it

PR #1294, opened 2024-01-29
New sounds for the BFG and plasma rifle

uhbooh:
Made by MoonDeLaAxel and I, from CC0 sounds.

Sources
https://opengameart.org/content/space-laser
https://opengameart.org/content/doomsday-laser-cannon-sound-effect
https://opengameart.org/content/synthesized-explosion


mc776:
I like them and they definitely work well, though they need to be much louder to sound right ingame.

Poking around on bart's OGA profile and they look legit so far... i'm not missing anything, am I?

mc776:
If we're about to post a release, my opinion is that this PR should be deferred and the release should not be delayed because of this.

However, I'm not going to add the deferred tag because if the release is delayed for totally unrelated reasons and all outstanding issues are addressed on this PR, then it should be merged.

mc776:
While testing #1296 I'd accidentally used a wad that had both that sound and these ones and yeah they are definitely not loud enough...

uhbooh:
Yeah I'm aware that the sounds aren't very loud. My skills with audacity are not the best, but I'll try to fix them sometime soon.

Snoucher (changes requested):
I just noticed that in chocolate doom dsrxplod.wav doesn't play to it's end, which creates a noticeable cut off in the audio. We should probably do something about it before we merge it.

Xindage:
so we anyone else wanna point out if you like this one or not? i'm inclined to close this one.

mc776:
> I just noticed that in chocolate doom dsrxplod.wav doesn't play to it's end, which creates a noticeable cut off in the audio. We should probably do something about it before we merge it.

In vanilla the sound only plays as long as the actor emitting it exists.

The BFG ball animation has six frames at 8 tics each, so that's 48 tics or 48/35= 1371 milliseconds.

This sound is over 3 seconds long, about 2.5 seconds before it's too quiet to hear.

uhbooh:
Well it has been over a year since I worked on this, but if enough people like these sounds I can finish them. My Audacity skills have improved quite a bit so I can make these sounds fit in Freedoom better.

PR #1296, opened 2024-01-29
Hopefully resolve #1122.

### LINKED ISSUE #1122: Octaminator not telegraphed enough
Because the Octaminator is shorter then it's doom equivalent, it doesn't have to lean forward to attack, this change makes it easer to miss it's attack in the heat of battle, making it harder to dodge.

unithecutie:
#1122 - "Octaminator not telegraphed enough"

This issue's been open for longer than it should've and this should hopefully resolve it.

Reuse 0.9 and before's rocket launcher sound to make the revenant attack sound more distinct and impactful, allowing it to be heard easier than the quite weak "swoosh" it is currently.


mc776 (approved):
At first I thought this was on the loud side but it works fine ingame. (Maybe still a *bit* on the loud side but not jarringly so IMO.)

I can now dodge octaminator missiles without looking from the starting open area in Map13.

selliott512:
I like it. I agree with mc776's assessment of the volume, but it's fine. It's not louder than the impact sound, unless you are petty close. It's also reassuring that it was once in Freedoom.

PR #1297, opened 2024-01-29
New d_map32 music - "Choice of Life and Death"

rostuhan:
I wrote this music for the map32.


unithecutie:
Eh, no offense - but I prefer "Black Mountain" leagues over this.

rostuhan:
> Eh, no offense - but I prefer "Black Mountain" leagues over this.
Yeah, "Black Mountain" is a good music, but I feel like "Choice of Life and Death" fits more to the amount of monsters you have to fight in the map 32.

unithecutie:
"Not Sure" is a arena-type thing (according to C5TEXT - which still refers to MAP31 by the old map, "Safety in Numbers", but MAP32 was never replaced).

The very dark, grungy "Black Mountain" fits this level's aesthetic perfectly.
"Choice of Life and Death" is in the same situation as ["Low"](https://github.com/freedoom/freedoom/blob/f6e8744fc1198d858a8931f29672a33383dd756a/musics/mellow/low.mid) was, where it doesn't fit the map all too well.

It could be repurposed as a FreeDM track, though!

rostuhan:
> "Not Sure" is a arena-type thing (according to C5TEXT - which still refers to MAP31 by the old map, "Safety in Numbers", but MAP32 was never replaced).
> 
> The very dark, grungy "Black Mountain" fits this level's aesthetic perfectly. "Choice of Life and Death" is in the same situation as ["Low"](https://github.com/freedoom/freedoom/blob/f6e8744fc1198d858a8931f29672a33383dd756a/musics/mellow/low.mid) was, where it doesn't fit the map all too well.
> 
> It could be repurposed as a FreeDM track, though!

You're right, I will think of a better music to replace that will fit the map.

rostuhan:
I'm closing this for now.

PR #1298, opened 2024-01-29
Remove reference to old MAP31 in DeHackEd

unithecutie:
The DeHackEd text for the MAP31 to 32 secret exit still mentions "Safety in Numbers" - 0.10 and before's MAP31.

> hell-rock that strange prison had been on.
Screenshot_Doom_20240129_160025

This could most likely be worded better.

### DIFF
--- a/lumps/dehacked/dehacked.txt
+++ b/lumps/dehacked/dehacked.txt
@@ -589,7 +589,9 @@ C5TEXT = You step into the teleporter. You feel a\n\
          You'd rather not find out.
 # Before MAP32 (secret level #2):
 C6TEXT = Forcibly uncaged again. Good job.\n\n\
-         But where are you? This... this isn't the city?!\n\n\
+         But where are you? The air and gravity\n\
+         still feel like whatever planet that\n\
+         strange prison had been on.\n\n\
          You look around and the layout triggers\n\
          some old memories from history class.\n\n\
          This is an arena.\n\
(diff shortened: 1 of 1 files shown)


mc776:
I wrote that line as a reference to the current Map31 (the one where you start in a cage in a row of cages), having no knowledge of the previous Map31 (in fact this PR is how I learned that this other map ever even existed). I had absolutely no idea what you were talking about when you mentioned he "Safety In Numbers" reference in that other PR so I think someone who's reading this for the first time wouldn't think about that other map that they've never heard of before.

At least in sourceports that fix the persistent-sky bug, the secret levels share the same sky. The reference to "rock" is derogatory for the planet, a reference back to the frequent references in Phase 1.

I find the narrator's wide-eyed *shock!* that this random teleporter would not immediately send the protagonist back to the city more jarring than anything else.

It seems more reasonable to misread the "hell" part to be a leftover Doom reference than to misread the "rock" part as being about the old preceding secret level from almost 10 years ago.

unithecutie:
Well, that certainly confused me. "strange prison" and "hell-rock" reminded me of "Safety in Numbers" more, since there's nothing really "rocky" in "Be Quiet".  Should I close the PR?

mc776:
How about replacing "hell-rock" with "planet"?

mc776 (approved):

PR #1300, opened 2024-01-29
New d_dm29 music - "Before the end"

rostuhan:
Previously known as d_map32 - "Choice of Life and Death" and d_map16 - "Night City".
The title is a reference to the map 30 which is the final map, maps 31 and 32 are technically between maps 15 and 16.


mc776:
I definitely like the current d_dm29 better (or any of the other Crucified Dreams tracks - I'm not sure why uni suggested those in #1297).

I should note you're already in the credits for two tracks in FreeDM so it's hardly like your contributions are going to be erased in the release (and tbh I like both of those tracks a lot more than this one).

At this point I'm not aware of any particular track that "needs" replacing... everything should be unique now (except d_intro and d_introa because screw that) and I don't think anything is an obvious id remix at this point.

rostuhan:
> I should note you're already in the credits for two tracks in FreeDM so it's hardly like your contributions are going to be erased in the release (and tbh I like both of those tracks a lot more than this one).

Correction: it's actually 3 tracks.

mc776 (dismissed):
Approving this for now as I suspect the overall feedback will end up tilting in its favour (and currently does in the absence of anyone chiming in).

MoskovchenkoD:
I'd also give a small extra doze of preference for the current dm29 track

PR #1301, opened 2024-01-29
New DM02 track

unithecutie:
I retract my suggestion of replacing Ralphis' Crucified Dreams tracks, this is the only one that needs replacing in my opinion.

"Untitled Jam" is really weird and just... doesn't work at all.


mc776:
I think both tracks are good, to the point where this seems to be a horizontal change. Untitle Jam is dissonant at first but it works out as it progresses.

Can we get a third opinion on this?

mc776 (approved):
Approving this for now as I suspect the overall feedback will end up tilting in its favour (and currently does in the absence of anyone chiming in).

PR #1309, opened 2024-02-01
Wip Rework of map22.

Xindage:
I've been working on this map slowly and i'm looking for feedbacks about the actual map22 and my wip rework attempt.
Please i want to finish this rework by myself so, only feedbacks for now.

**Changes:**
The difficult was increased a little early on.
Some new areas keeping the thematic of the map.
Additional enemies on coop on all skills.
Reduced some more the amount of linedefs in the main area of outside.
Reinforced the theme of the map.
Lava does more damage.
Secrets gives better loot.
Good weapons are introduced early.
There is no way to do skips by using jump.

**Todo**
Create an additional area near the yellow key, this place is not done yet.
finish the area where you release the revenants swarm and the room after you get teleported.


mc776:
One of my complaints about that map has been it's way too ridiculously hard at the start...

mc776:
On another note, if the result is going to be significantly different from the original Remanasu, can we bring back the name Lava Canyon?

unithecutie:
I would strongly prefer to keep the original name "Remanasu".
It was already heavily modified before being put in Freedoom anyway.
image

mc776:
ok but now that I look at that I gotta say whoever got the idea to add that lava aesthetic was a genius...

Xindage:
> One of my complaints about that map has been it's way too ridiculously hard at the start... then again, my main problem is the lack of weapons and ammo (at least before the changes I had introduced but even those were fairly conservative)

Now I made the ammo tight again in singleplayer, but you're granted a rockets and berserk (secret) early on the map so it may be not an issue in long term, after that near yellow youll have backpack and then you gain acess to the plasma where the red key is, so you'll have weapons to choose.

Xindage:
@mc776 i would like you to review my newest changes so far, the map is still now a beta, can be played from the start to the end, but there's a area near yellow key that it not done and its blocked, still you can progress, tell me what you think of this version of the map so far.

mc776:
Some very initial impressions, haven't completed and haven't tried without god mode yet:

- I know people hate the funky switch but please keep it a switch not a walk line. The elevator is clearly coded as a starting "safe room" that requires deliberate interaction before getting into LOS with the enemy.
- *Really* not a fan of the redecoration at the bottom. The flat featureless shiny white metal look so bad! 
- Not a fan of the new octaminator by the yellow switch. It just feels really gratuitous and arbitrary - would rather just have the old minigunners in the bunkers since that's what that part of the map looks like.
- I *do* like the minigunner behind the door though.
- The stairs leading up from that are not an improvement.
- The crate hall feels way too cramped now, I'm basically just constantly backing out and SSGing the crowd inside.
- ...please just stop with the AQPIPE11 steps :sob: 
- The new door leading to the yellow key: please remove the door and just make the guys on the other side ambush, *or* set it up so that all the previous silver doors on that side of the hallway are all very clearly blocked by crates. I didn't even realize that was supposed to be openable until I noclipped past it by mistake.
- I hate these minigunners hidden behind these impassable lines. They're every bit as obnoxious and had to see as the old bunker guys but in a much more cramped space, a lot less fun to engage, and they don't even make any internal sense. Best to just have them free to move about.
- What happened to the spectres by the yellow door? They were perfect for that dark area.
- I ran around the map three times looking for that blue key. **why is there a wall there blocking the view of the blue key.**
- The sky has fallen around that area with that elevator that was uglified. I think a sector tag got accidentally misused somewhere.

So basically: I'm really not liking these changes. Most of them seem to be aimed at making the map harder, but not in a natural way but in a tiring slaughtermappy way, and the aesthetic changes are utterly inexplicable.

Honestly this map was already stressful and *way* too long and complicated before and this is making that so much worse...

mc776:
That all said: Remanasu is one of my least favourite maps in FD to begin with. If someone who's a huge fan of Remanasu likes these changes they may have something going for their opinion...

mc776:
Now re-reading the OP and realizing that a good deal of what I'm reacting to is intentional.

For what it's worth, this is going in the exact opposite direction of #1289, which together is going to make FD a lot more inconsistent in its difficulty scaling.

mc776:
For what it's worth, an extended remix of this map could work as a neat Freedoom-based project, I'm just concerned the extension *as part of the main Freedoom IWAD* is going to bring back a lot of what people were complaining about in the difficulty balance of a lot of levels.

Xindage:
Being honest, your commentary hit me like a brick, and I haven't been in the mood these past few months. I was almost giving up on this, but... okay, let me read that again and try to process it in a more mature way.

1. "The Starting lift switch." God, I hate this thing with my live but ok, if it has to come back it must be the last thing that i'll add in this map when its done to release.
2. "The decoration and the hp bonus" I can see something else to apply there, and now about bonus healths, i'm a completist kind of guy, i do like to reach 100% items and walking around there was annoying as well, while I go by the thinking that "We are on map22 we don't need to babysit the player into where is the only path that he can go, unless else walking on lava (death). So do you have a better suggestion?
3. "Various comments on different places." Since i cant really figure everything i'll go by what i could think you meant:
- The octaminator AT the yellow switch was already there, i dont changed it, if you are talking about the one behind the red rocks, yes this one i added.
- The minigunners most of the times just killed themselves if you ignore them due its good accuracy and they could not move as well, that's why i replaced by the octaminator, but if you have a better suggestion than having trapped gunner i would like to know. or i can just think on something else later.
- The only spectres that i remember is the ones down there in the underground, they're already there, but was moved around so they actually anbush you instead of just rushing at you on sight and getting stuck at the stairs, if i'm not wrong the others were the ones who come at the teleport, that was replace by barons because i give you access to plasma and you have rockets as well, before you get plasma on this map only on easy skill.
4. "Minigunner at the door" thanks but its purpose goes beyond that, make more sense in multiplayer were you can actually lost all the miniguns chances up to this place.
5. "Stair rework" This one took a while to pullout without looking bad, nut i managed to keep the theme.
6. "Cramped crate area" Aside a single crate at the start and another on the end, its the same layout from before.
7. "Aquapipe steps" The green tekwall just cuts and wont fit in a 8px walls, thats why i went to this and to keep theme i had to do it on all doors.
8. "The door on the U-shaped path" Well the door had a clearly had different textures, but i'll found a way to block the others doors so, since i want that area to be a little darker. (I still dont know how you can noclip thought a different textured door by accident as you say)
9. "Minigunners closets" I wanted some static threat, i was actually wanting to korp made a texwall texture for me whit a hole on it, similar to that brown used there but i dont think that korp will do it, so i'll make it to work different somehow.
10. "Underground spectres" They're still there as i said, but hidden behind the walls so they can properly engage on you only after you really go into there.
11. "Why there's a wall" because there was a place already exploding in linedefs limits causing a visual homs, the easiest solution was to hide the interior of that room since it reduced lines from 290 to 260, it's still not perfect but already saved it to be under vanilla limits. you can make it happens when you open the red door near the spider and stand on where ssg is and look in the direction of that windows. but I had a small idea to help you out.
12. "Switch blending" Of course i can consider some lights.
13. "VPO" Give me the coordinates and i'll double check it.
14. "Mistagging" The location of the glitch would be nice.
15.  "Difficult" The only location where i really died on this map was on the revenant slaughter fest, but this was already there before it was the hardest area or had any other, i an tone it down a bit maybe by changing few revs by spectres and removing few others.

mc776:
To be blunt, my biggest problem with Remanasu is that it's incredibly long, meandering and confusing and complicated and tedious, and a major effort to make it even *more* so - in such a way that *major progression points have to be obfuscated to prevent rendering errors cause by all the complication* - isn't going to get an approval from me.

Xindage:
Its done the final version, with deathmatch and coop properly set and totally different from my previous attempt:

Changes:
Visplanes drastically reduced, there is no overflow present, even if death or doing archvile jumps.
Difficult ramped down slight. (there less archviles)
There is way more ammo and health. (more plasma and rockets as well)
There's just one monsters teleport ambush, everything else is geometry.
The navigation of the map is more clear now, its not as easy to be lost.
The new area for the yellow key is now done.
Reworked the lava river so you could not go on the red door there anymore. (you can after getting the key)
The outside got a major rework and textures issues was fixed.

The difficult is still high, but you're not forced to fight anything to progress, still leaving things behind can haunt you later.
A 100% run of the map can be done at a normal pace in 16 to 20 minutes.
Open for final feedback now.

PR #1310, opened 2024-02-01
lumps: Reworked ENDOOM screens

Anonysussy:
I've been working on this for at least probably 2 days, i'm also going for the more stylized look with this one while also trying to mimic the original ENDOOM look to it.
Screenshot (887)


selliott512:
That's nice, but I have to warn you that a significant amount of effort was spent on the `ENDOOM` lumps a few months ago, so it may be a hard sell.

Also, "mimic the original" is mostly not the direction that Freedoom wants to go. It more wants to distinguish itself as its own unique game.

And finally, it's easier if you create a branch in your fork rather than creating pull requests with your `master` branch since your `master` branch is now tied up. It's fine though. You can create a branch for your next pull request.

unithecutie:
I think "the original" in this context would be the old ENDOOM? The one that existed from v0.1 - v0.12.1.

mc776:
I agree wth selliott512's comments.

I'm guessing "original" means the previous FD ENDOOM because of the old green and blue branding, but even then that green and blue branding is pretty much totally gone from anything else in FD now, so it just sticks out inappropriately.

The word "Free" in this logo is actively painful to read.

We've discussed the text before.

PR #1313, opened 2024-02-02
levels: improve Map29 aesthetics (mostly).

mc776:
The less-safe stuff that I dared not suggest in #1289. (It's always a little scary to touch anything involving the lines of a sinking-stone sector that's attached to anything else.)

Checked with Visplane Explorer before uploading.

- The crumbling land has more linedefs now. The oddly straight geometric shapes especially around the SSG never sat well with me - it doesn't appear to be meant to be the remains of anything artificial, and I can find no visplane overflows or other vanilla rendering issues resulting from these additional linedefs.
- The torches around the jump teleporter are green again, fitting the original theme of the map where they were all green. I'm guessing the red was intended to better telegraph that this is a special thing the player is expected to reach, so the platform itself has been retextured as to distinguish it from the surrounding rock formations.
- Gave the serps near the cloak secret more room to move.
- Moved the cloak powerup a little bit deeper inside the secret sector while expanding the wall to make sure there are no pickups from the other side.
- Realigned the metal textures inside the doors in the starting hub.
- Set the remaining 3 sectors in that teleporting window to the blood pool to non-damaging. (This is the reason for the bug tag on the PR.)
- Aligned the elevator and exit window heights with the wood textures.
- Used MARBLE2 for the backs of the face panels.
- Resculpted the crumbling SKAG path to make it look a little bit more convincingly like broken pieces of a path.
- Widened the butt end of the southwest arrow so it looks less like someone drew it with a shaky hand.
- Addressed #996.

Before/after:
map29_before
map29_after


unithecutie:
> red was intended to better telegraph that this is a special thing the player is expected to reach

That was my intention, yes.
Your implementation here looks a lot better against the rest of the map, though.

selliott512 (approved):
Nice changes. Eureka passed. IDDQD + IDFA 100% UV-MAX no problem.

Your before and after is great. For those that like blinking things here's my version as an animated GIF:

map29-aesthetics-mostly

Also, I've always thought the secret in the minigun zombie room (linedef #933) was a bit non obvious. Should it have misassigned texture, or something?. Maybe it's supposed to be an extra secret secret:

linedef-933

mc776:
DOOM01
DOOM00

selliott512:
Looks good. For your latest changes I only verified the minigun zombie room switches, but I'm sure it's fine. Eureka is happy at least.

PR #1314, opened 2024-02-02
levels: improve E3M5.

mc776:
- If you miss that *one* shellbox accessible from the start, the entire first 1/3 of the map becomes exponentially, inappropriately more difficult. While it is fairly visible once you're down that corridor, you basically have no other reason to go there if you don't have the yellow key, so if you're distracted the first time you see that area (or perhaps you miss it altogether) you might not think to visit it again until you have the yellow key. A small trail of armour bonuses visible from around the corner has been added.
- Moved the tripod gate opening switch a bit so that the surrounding textures align and the fake contrast doesn't make it stick out like a sore thumb.
- The tripod fight sucks. It'd be an amazing arena fight if this were Phase 2 and you had the SSG in those tight quarters, but (a) without the SSG it takes *way* too long and (b) since that door opens permanently like that and there's no wall to follow away from it, the tripod spends 95% of its time blocking your entrance into the arena. Now it's an instant-raise ambush jumpscare - which gives you a moment to take evasive action since it won't start shooting until you're well inside the room unless you intentionally back out.
- That all said, that tripod fight as is is *still* completely inappropriate for just the 5th map in the episode. Compare the equivalent in Map27, where you not only have the SSG and possibly the SKAG, and way more ammo either way, but the entire area is *much* more open and you can easily retreat to a spot where the tripod takes several seconds to reach you instead of one second. The monster blocking line has now been unflagged and you are now able to lure it out into the hall; even if it never leaves that juncture just outside the door, at least it gives you a lot more space and time to work with.
- Made the soul sphere secret just a little bit easier to find with the fake contrast, because expecting the player to figure that a *misaligned rock texture* is a secret when everybody accidentally misaligns rock textures everywhere else is totally unfair.
- The green T room with the teleporter leading to the red slime maze now has a door leading to the hallway going into the final arena, which opens at the same time that said arena does. This should cut back on some backtracking without disrupting gameplay too much.
- Gave the two corridors by the start different marbfaces.
- The cross on the platform by the blue key looks a bit janky even if we weren't trying to get rid of the real-world religious imagery here. It is now a glowing infinity symbol.
- Tekwalled up the back side of that weird not-door to the southwest.
- Mixed up some of that BRICK9 with some BRICK8 where things seemed adequately unexposed to the elements. Split up an over-1024 linedef in the process.
- Flipped a bunch of "ingrown" linedefs.


mc776:
That super-long linedef must have caused some problems somewhere... might as well tag this bug too

unithecutie:
Might also want to point out the other awkwardly placed cybie, in E3M6 - the one near the switch that lowers the bars.

mc776:
Once you get past that initial bottleneck with the monsters that come out as soon as you hit the switch, the tripod can follow you out into the expanded red cave area - you actually have more room to work with than on E3M5.

That initial bottleneck is brutal though (it already is on medium where you have a small gang of painlords instead), but I'm not sure if it's *notably inappropriate* for "Freedoom Hard" hard.

Xindage (approved):

PR #1315, opened 2024-02-03
levels: tweak E3M6.

mc776:
Prompted by the discussion [here](https://github.com/freedoom/freedoom/pull/1314#issuecomment-1923782305). After a couple runs it seems the big issue is the huge traffic jam after you hit that switch making avoiding the tripod much less feasible than it is on any other non-E2M8 level that has one.

- Changed the red switch arena to make it a bit more cyberdemon-fight-friendly. The switch is now flush against the surrounding wall so even if you run right into it to hit it you won't get stuck in an alcove. The space on the sides of the big green pillars has been expanded. The red skull pillar has been moved so you can move around it on every side. We also have 20 more rockets because hey why not.
- Lines 939 and 946, the two very ends of the red rock corridor leading into the red switch, are now marked as sound-blocking. This prevents the small horde in the star courtyard from creating a traffic jam blocking the player's escape even longer.
- Lines 931, 962 and 963 are also sound blocking. If you can manage to take on the tripod staying entirely within the "arena" part, after the initial wave of serps you can pretty much solo it.
- Added more texture variety to that arena (different marbles, more tekwall, etc.) and gave that very lonely serpentipede up there some friends.
- Made a couple doors look a bit better (taking some cues from the E3M5 doors) and added visible lights to the tentacle-star-side of the main red door to explain why it's lit.
- Prettied up the bridge courtyard a bit. Hopefully the stairs are a bit more visible now.
- Raised the hatchling fireplace so that you can't get inside without jumping. Reshaped a good deal of the rest of the room to get it to work.
- Moved those four shellboxes so it's not so easy to run into all of them at once.
- Restored the old rising-from-the-lava effect on the final painlord.
- Skewed some orthogonal lines in rocky areas to avoid unwanted fake contrast.


unithecutie:
The secret exit also looks very slapped-in-there and out of place, coming from the fact it wasn't there when the map was created.

An experimental concept of mine is to make the current secret exit switch reveal a soulsphere at the start of the map, and have the secret exit just be non-optional (changing the normal exit lines to secret exit lines) - extending E3.

mc776:
Looks fine to me and there's no reason for creating a new inconsistency.

mc776:
I tracked down the PR #622 and the more I look at it the more I appreciate how it's like this budding *growth* protruding from the path, like it's some shifting organic thing deciding or helping you decide your fate, it's not clear which.

If anything I'd almost want to lean into that by making that room visible even when unlocked...

selliott512 (approved):
Nice. My usual blast through no problem UV-MAX, but I did pay attention to the red switch and I see how that helps.

If you don't mind adding some texture to the upper part of of linedefs 1052 and 1319 it would make Eureka happy.

If you feel like it, and you agree that it's an issue, I wish the secret exit and the main exit had the same size rooms and portals. it doesn't matter much, but it does seem like something that could match.

mc776:
I think they should be clearly visibly different so that you can tell one of them is the secret exit.

On that note, there should be something *marking* them as exits, so I propose:
DOOM00
DOOM02

selliott512:
That's great.

PR #1316, opened 2024-02-03
textures: add tileable TEKWALL3 variant.

mc776:
128x128 and should align more or less without glaring seams with either itself or TEKWALL1.
tekwall7-preview
In this preview the new texture is where the marbface usually is on E3M6, and is surrounded on both sides by TEKWALL1.

Proof that there is no tutti-frutti:
DOOM00
Aesthetically the vertical tiling leaves a lot to be desired, but this is the best I can do without creating a new patch.


mc776:
Here's what the tutti-frutti looks like with TEKWALL3 on that rock column:
DOOM01

selliott512 (approved):
Looks good to me. I viewed in in a level editor briefly. Having 128x128 versions of textures seems like a good idea. I wonder if there are others like this, close to 128x128, but not quite.

PR #1318, opened 2024-02-03
levels: use old slime bricks for Map11 west entry.

mc776:
This is actually a perfect example of why I wanted to keep the old slime bricks in #1229. This is supposed to be a very poorly maintained, badly worn-out, poorly lit, barely structurally functional section of the building that the new bricks just look far too solid and *intact* to convey properly.

(I did also take the liberty of putting more ashwall into the more irregular lines)
DOOM04
DOOM03


uhbooh:
When I made the new slime brick texture it was intended as a better fitting replacement. For example pwad maps that are made with the vanilla doom 2 textures in mind. Keeping the old freedoom bricks as a separate texture is a great idea for situations like this, because these maps were made with a freedoom texture that was noticeably different from the original doom.

mc776:
Superseded by #1324

PR #1319, opened 2024-02-03
Tweak E1M5

unithecutie:
Change the sofas to resemble actual sofas instead of flesh, using COMPTILEs with texture offsets to emulate chair legs without having to make new textures.

Restore Ralphis' old E1M1 track as E1M5 - taking Matt's suggestion [here](https://github.com/freedoom/freedoom/issues/1292#issuecomment-1913358481_) into heart.
"The Raging Sun" - although I like it - is quite bland. Ralphis' old E1M1 track is a fast-action-styled track that fits the shooting gallery aspect of E1M5.


mc776:
The shape and actually adding parts to the couches are certainly an improvement, but is there really no red or warm colour option?

It really makes that room look a lot more... bleak? sombre?
DOOM05

And (though this could just be the old flat not quite matching the shade of the old patch) with the blue the shapes seem to get lost at a distance:
DOOM08
DOOM07


Also, a bit unrelated but I just noticed this (it was already like this your change, I just never really gave this any thought until now):
DOOM00
Like it's not at all clear where these bands of light are coming from... did this room used to be much darker and whover changed it didn't change all the sectors? Are those dark ceiling bits supposed to be *radiating darkness* onto the floor below? The lighting is just so odd...

mc776:
That said, I'm going to suggest the changes in this zip:
[e1m5.wad.zip](https://github.com/freedoom/freedoom/files/14151111/e1m5.wad.zip)
It fixes the lighting and swaps out the flat used for the seat of the couch to a darker blue, allowing the couch's shape to be visible even in low resolution.

I've also addressed #996 for the crate area and simplified a bit of geometry (no visual changes, just merging sectors).

DOOM01
DOOM00

unithecutie:
Added your tweaks. :P

mc776 (approved):

mc776:
Taking a look again this morning and the blue couches are fine... especially given this is supposed to be an *armory* and the tacticool vibe definitely fits.

PR #1320, opened 2024-02-03
bunch of Cascade stuff

UltimateCraneo:
idk a bunch of sprites by Cascade

. . .

I HATE GITHUB, I HATE GITHUB


mc776:
My thoughts:

Columns: 
greencolumns-old
greencolumns
Not 100% ready to agree it would be an improvement on the balance - they're a bit simple-looking for a decorative pillar, especially the red ones.

I think all of these columns could look a lot better if the column parts tapered a little so they're a couple pixels wider at the bottom.

Col5:
columnguy
I can see the rationale behind every decision behind this image, and I can't disagree in principle in my capacity as a maintainer of this community project: but with all respect *aesthetically* it's a massive regression in every way. We took a weird-looking but clearly ornate and old decorative sculpture and replaced it with what appears to be the new green column (which I initially mistook for a garbage can until I looked closer) wrapped in a ribbon with a severed head on it. The visual weight is very top-heavy and the lines are awkward.

Suggestion: make the tentacles wrapping the column look more like roots, and the ??severed impaled head?? on top be instead the pulsating main body of the creature. It could be a bit more red, or be oozing something red, to better match the colours of the Doom sprite (something that the current lizardbaby does).

Col6: Same comment about the red columns generally but I do like the skull.

Ripsaw: Hell *yeah*. But I would like to suggest the following offsets:
```
SAWGA0 -150 -59
SAWGB0 -150 -60
SAWGC0 -160 -81
SAWGD0 -162 -81
```
This would give a better sense of thrusting the saw into the target.

Techpillar: We've discussed this before. I approve.

Fist: No. At first I was just going to object that the sprite is too oily-shiny and the left fist is too weirdly small, but after trying it out I realize this is also regression from when Anonysussy made the current fist animations much better in #1193. (Also the shrunken left fist is misaligned, there's a huge empty space below it if you disable the status bar.)

Trees: No. They just feel worse in every way - less alive, less tree-like. They just feel dead and old rather than alien.

Dust, ash, death.
newtrees

Life, verdancy, *visual contrast*.
currenttrees

I do like the *shape* of the small tree, though - really if we can just have this but in the colours of the current tree I'd merge that (assuming it didn't come in a massive omnibus PR that contained other things that shouldn't be merged)


As a general comment, the sharper perspective on the bottom of everything is going to look really bad in GL ports (which is why the current techpillar's base looks like that).
billboard

unithecutie:
I did suggest this should be split up into multiple PRs, it's a bit whacky.

mc776:
> I do like the *shape* of the small tree, though

After giving E1M1 a try with this palette-swapped version
SMITA0

I'm going to retract this above comment - the details actually make it look worse at low res in vanilla while you're moving, like they form a grid pattern that creates a Moiré effect like the one removed in #1240.

unithecutie:
@mc776 Could you close this in favour of 3 seperate PRs for the chainsaw, tech pillar and columns respectively?

mc776:
I was just thinking about doing that later, thanks!

Cescade:
The thing with the trees is that despite with the 'alien tree diversity' they just do not look good together because of said different artstyles. I pretty much went with the gothic vibe you've said back then on discord. 

> They just feel dead and old rather than alien.

Of course they feel dead, The burnt branches backs it up.

I'll try to follow-up from these aswell soon.

PR #1322, opened 2024-02-03
New green and red columns by Cascade

unithecutie:
Split #1320.


mc776:
My comments here: https://github.com/freedoom/freedoom/pull/1320#issuecomment-1925425994

mc776:
Changing the title because I looked at the PR list and did a double-take when my brain somehow parsed "column" as referring to the techpillar despite distinctly recalling that I'd merged that PR earlier.

Cescade:
> They're a bit simple-looking for a decorative pillar

Coming up with something to resolve the simpleness of the columns, like ancient hand carvings on the upper part. But I'm still looking for recommendations if you have any on mind. 

> It's a massive regression from the lizard baby fountain in everyway.

I've no idea why we should keep up with the lizard baby stuff honestly.

As for the stuff I've unmentioned, I'll keep them on mind and will be posting a follow up from all those.

mc776:
As far as general decor goes I imagine something like tentacle and fish themes for the green and fire and star themes for the red.

mc776:
I think I like the variant with more branching and no upper tentacles... I'm partial to the blob rather than the teeth but I can see the teeth working too depending on the animation

Cescade:
Ingame:

mc776 (approved):

mc776:
*Yes.*

(Leaving open for a few hours in case last minute etc.)

Issue #1317, opened 2024-02-03
textures: some id definitions are changed which may affect compatibility.


mc776:
Here are the FD maps I can find that currently use BROVINE2 that need to be checked once the texture is corrected:

```
E2M1
E2M3
E2M6
E2M9
E4M2
E4M8
E4M9
Map11
DM09
```
[rest of original post follows]

While doing #1316 I noticed there were two BROVINE2s: one Doom 2(?) texture with the green WALL62_2 and one custom Freedoom texture with the brown WALL62_1.

Was the FD one supposed to overwrite the D2 one? Or was this supposed to be a new texture and the "2" was a mistake?

Here's D2 Map01 (loaded as a PWAD in FD) for comparison:
![DOOM01](https://github.com/freedoom/freedoom/assets/24984517/8a213e1e-d15e-4a7a-a571-853a9f6254f5)
![DOOM00](https://github.com/freedoom/freedoom/assets/24984517/730a5d8f-490d-4324-9810-bae728bdd7a1)

And if I rename FD's BROVINE2 to BROVINE3 here's the "old" BROVINE2 revealed:
![DOOM02](https://github.com/freedoom/freedoom/assets/24984517/704396ab-0aee-476c-a7cf-5bde7318664f)

mc776:
I've also noticed that a lot of the GRAY texture definitions depart from Doom's:

FD:
```
GRAY1       		64	128
*	W33_7       	0	0
*	W33_9       	0	64
GRAY2       		64	72
*	W33_7       	0	-8
*	W31_2       	0	56
*	DUCT1       	21	16
GRAY4       		64	128
*	W33_5       	0	0
*	W33_6       	0	64
GRAY5       		64	128
*	W33_7       	0	0
*	W33_9       	0	72
*	W33_8       	0	64
```

id:
```
GRAY1       		64	128
*	W31_1       	0	64
*	W31_1       	0	0
GRAY2       		64	72
*	W33_7       	0	-8
*	W31_1       	0	56
*	DUCT1       	20	16
GRAY4       		64	128
*	W33_5       	0	0
*	W33_5       	0	64
GRAY5       		64	128
*	W33_7       	0	72
*	W33_7       	0	0
*	W33_8       	0	64
```

Has it been decided that there's no longer any functional need to follow id for these, as long as the underlying replacement patches and the resulting textures are the same dimensions?

Xindage:
[...]

PR #1324, opened 2024-02-03
textures: rename FD's new BROVINE to BROVINE3.

mc776:
See #1317.

~~It's easiest to update the affected maps after this is committed.~~ yay scripts


Xindage (approved):
Good work.

PR #1328, opened 2024-02-05
levels: rename Episode 2.

### LINKED ISSUE #1327: Rename Episode 2
https://github.com/freedoom/freedoom/pull/1326#issuecomment-1927531629

I'm so glad someone else brought this up. These haven't been "military" and scarcely "labs" in forever.

As far as FD's existing plot goes, this area needs to:
- be known for having a starport
- be connected to the prison outpost and the protag needs to know about it

Of course, even these aren't necessarily set in stone - the texts can easily be rewritten so that the protag takes that teleporter and finds themselves in an unknown location.

As far as E2's aesthetic goes, it is:
- techbase with lots of alien monster stuff enmeshed into it
- covered with water and vines, very overgrown

Maybe it's another outpost on the same planet, but more equatorial?

mc776:
Closes #1327.

### DIFF
--- a/CREDITS-LEVELS
+++ b/CREDITS-LEVELS
@@ -11,7 +11,7 @@ Episode 1 - Outpost Outbreak
     E1M7: Xenobiotic Materials Lab by Will "Kid Airbag" Hackney
     E1M8: Outpost Quarry by Luiz Henrique "Xindage" Gasparin Jeronimo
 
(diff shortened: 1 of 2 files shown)


selliott512 (approved):
Seems like a nice name. It's probably more unique to Freedoom than the previous one. I like the text. No spelling mistakes or typos as far as I can tell.

PR #1329, opened 2024-02-06
levels: fix minor E2M3 lighting change issue.

mc776:
After the lights come back on in the blue room the lights in there appear to cast darkness instead of light. 
DOOM00

A bunch of stuff has been rearranged slightly so that the brightest adjacent sectors to the affected ones are not brighter than these light sectors.
DOOM02

As a side effect, this means that the gradient of the big techpillar light by the entrance is less affected by the in-game shutoff than it used to be, and remains brighter than its surroundings after the main light is restored.
DOOM03
Unless the player has enough minigun ammo to camp there on purpose I don't think it'll affect gameplay for most people.

I've also simplified that room to use much fewer sectors - none of the lines between them were sound blockers and there were no changes in floor height or texture or lighting.


selliott512 (approved):
New carpet. Nice.

I tried just that room in game (but not the entire level). Eureka is happy. FWIW here's an animated GIF showing changes in the expected location:

e2m3-1329

PR #1330, opened 2024-02-06
credits: remove doubled Mort.

mc776:
Keeping the 19 in case someone does only a string search for "mortrixs19" and not "mortrixs".

Got rid of the time stamp in the invitation to update info.


selliott512 (commented):
Nice. I take it you know that `Mortrixs19` is the more correct one. It seems to be the most recent one in `git log`.

I made some effort to merge duplicates in the past, but in that case I only caught exact duplicates with names and email addresses.

Should we sort this file by `S`? It would make duplicates more obvious. I could probably do it with a bit of scripting (not manually).

mc776:
I kinda like how it's naturally sorted roughly chronologically myself.

selliott512 (approved):

selliott512:
No problem regarding the sorting. Just a thought.

PR #1332, opened 2024-02-07
scripts: For "makepkgs" replace "is" with "=="

selliott512:
The "is" keyword is no longer supported, which result in a warning in "make dist" as well as unintentionally having a "v" in the resulting ZIP files and directory within those ZIP files.

---

This solves an 11th hours 0.13.0 mystery I had with my ZIP files looking different than previous ones - "is" support depends on the Python version whereas this  fix does not.

This should be backported to 0.13.1. Should we have a "backport" tag?

### DIFF
--- a/scripts/makepkgs
+++ b/scripts/makepkgs
@@ -24,7 +24,7 @@ version = os.getenv("VERSION")
 if version is None:
     sys.stderr.write("Version not specified for release\n")
     sys.exit(1)
-if version[0] is "v":
+if version[0] == "v":
     # Strip the leading "v" from versioning
     version = version[1:]
 


mc776 (approved):
0.13.1 isn't out yet so I don't get what backport means here?

unithecutie:
> 0.13.1 isn't out yet so I don't get what backport means here?

Tagging is set to say `0.14.0-alpha` currently.

mc776:
Would be great if we could have a script that just automatically sets all the numbering stuff that we could run just before a final release...

Anyway this looks like a needed fix but @selliott512 please merge when you're done because I do not know what I'm doing.

selliott512:
In the future when we are ready for 0.13.1 we may decide we want some subsets of the fixes that have been done since 0.13.0. In that case I'll create branch `freedoom-0.13.y` initially pointing to the 0.13.0 commit. Then, whichever subset of fixes in the master branch that have been done since 0.13.0 will be cherry-picked (copied) from master to that branch. I'm inclined to call that cherry-picking backporting (back to an older version from a newer one).

Another approach is to just tag PRs with the version when they will be first released, so this would be tagged "0.13.1", but then a new tag is needed with each release. A tag named something like "backport" is reusable, but possibly more confusing. Either approach is fine.

selliott512:
I could try to create such a script. It's really not very many places.

PR #1333, opened 2024-02-08
dist: Use non-thumbnails for screenshots.

selliott512:
Switch to using the full screenshots in the metainfo files. These URLs should not change.

----

I've checked them, and I've added a comment that should help with the confusion as a result of their being both PNGs and JPGs. My impression is these should be pulled infrequently such as when flatpak and Linux repo packaging is done each Freedoom release, or Linux distro release.


mc776 (approved):
I suppose if they ever did change going forward there'd be the commit on the website as an advance warning.

PR #1335, opened 2024-02-12
Add a "previously featured in 0.12.1" section to CREDITS-LEVELS

unithecutie:
Dunno why I didn't do this in the initial PR...

### DIFF
--- a/CREDITS-LEVELS
+++ b/CREDITS-LEVELS
@@ -7,7 +7,7 @@ Episode 1 - Outpost Outbreak
         E1M9: Nutrient Recycling by Mortrixs19
     E1M4: Supply Depot by GeekMarine
     E1M5: Armory by Oleg "Wereknight" Vovk
-    E1M6: Training Facility by Luiz Henrique "Xindage" Gasparin Jeronimo
+    E1M6: Training Facility by GeekMarine
     E1M7: Xenobiotic Materials Lab by Will "Kid Airbag" Hackney
     E1M8: Outpost Quarry by Luiz Henrique "Xindage" Gasparin Jeronimo
 
--- a/CREDITS-LEVELS
+++ b/CREDITS-LEVELS
@@ -125,3 +125,33 @@ Cluster 3
     DM30: Last Man Standing by Dean "deathz0r" Joseph
     DM31: Desolated Fort by Luiz Henrique "Xindage" Gasparin Jeronimo
     DM32: Fourplay by Brett "Mechadon" Harrell
+
+
+--Previously Featured in 0.12.1--
+
+E1M5: Main Control by Miguel "Z0k" Suarez Gomez
+E1M7: Transportation Bay by Will "Kid Airbag" Hackney (renamed, not removed)
+E1M9: Armory by Oleg "Wereknight" Vovk (moved to E1M5, not removed)
+E2M1: Ruins by Luiz Henrique "Xindage" Gasparin Jeronimo
+E2M2: Power Plant by Andrew "Malinku" Rehberger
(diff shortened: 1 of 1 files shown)


mc776 (approved):

mc776:
I've approved it for now, but I do have two comments:

1. The current Map26 is actually just heavily modified from the previous one. Should it really be included?
2. I don't think we should include maps that have been moved but not renamed. Unlike the music we can see all the level names in-game (or looking at the dehacked) so you can see where things went, whereas the music credits are the only place in the package where song titles are stored, so this additional record doesn't seem to be worth the cost to brevity and readability.

unithecutie:
> The current Map26 is actually just heavily modified from the previous one. Should it really be included?

I can tell, but **heavily** modified is the key here.

> I don't think we should include maps that have been moved but not renamed [...]

I think that should still be pointed out, would be best to see what others think about this but barely anyone talks on the GitHub page...
/me shrugs

CREDITS-LEVELS
```diff
@@ -45,7 +45,6 @@ Episode 4 - Double Impact
     E4M8: Launch Bay by Matt "RottKing" Cibulas and Ralph "Ralphis" Vickers
 
 
-
 --Phase 2--
 
 Cluster 1
```
selliott512, line 50: I know it's outside of the scope of your current change, but we have episode names, but not cluster names? Should we have cluster names? There is probably nothing for you to change here now, but maybe something we can talk about here or elsewhere.

CREDITS-LEVELS
```diff
@@ -125,3 +123,33 @@ Cluster 3
     DM30: Last Man Standing by Dean "deathz0r" Joseph
     DM31: Desolated Fort by Luiz Henrique "Xindage" Gasparin Jeronimo
     DM32: Fourplay by Brett "Mechadon" Harrell
+
+
+--Previously Featured in 0.12.1--
+
+E1M5: Main Control by Miguel "Z0k" Suarez Gomez
+E1M7: Transportation Bay by Will "Kid Airbag" Hackney (renamed, not removed)
+E1M9: Armory by Oleg "Wereknight" Vovk (moved to E1M5, not removed)
```
selliott512, line 132: This is a flat list where the secret levels (`E1M9` and `E2M9` in this case) are not indented under their respective levels. But how would you do that when the level in question is not mentioned because it was not changed since 0.12.1? As I previously mentioned my inclination would be to have a simple flat sorted list (like this), and this may be another reason to do so. There may not be anything for you to change here, just my thoughts on the topic.

selliott512 (approved):
Looks good.

I didn't review the actual content (how each map was replaced renamed or moved), but I did check the indentation, spacing, spelling and order. That all looks good.

selliott512:
I think this will be helpful for people trying to trace maps through FD history, something that is otherwise hard to do.

CREDITS-LEVELS
unithecutie: Cluster names have no use.

CREDITS-LEVELS
mc776: > But how would you do that when the level in question is not mentioned because it was not changed since 0.12.1?

Isn't that what the rest of the file is for?

CREDITS-LEVELS
mc776: They don't appear anywhere ingame and wouldn't be alluded to in the intermission texts, and in vanilla the sky doesn't even transition to let you know it's a new cluster. Frankly the idea of D2 clusters confused me when I first encountered them... which was maybe sometime around 2010? and I've been had Doom II since 1996.

I would be against even mentioning them, except that they help break up what would otherwise be an awfully long wall of undifferentiated list items.

PR #1336, opened 2024-02-13
New PLAT2_1 by Korp

unithecutie:


mc776:
I like the shiny metal. At first it seemed maybe a bit *too* shiny but it's a regularly moving thing so it probably wouldn't have as much gunk on it as other map geometry anyway.

*Really* not a fan of the fucked-up top stripes. These imperfections might seem cool but on longer things the repetition will be really obvious... and they clash with other warning-stripe stuff that this texture would have worked with before. (The bottom of it also gives less of a sense of this thing being a platform on top of the piston thingy.) I think if the old patch top part could be combined with this we can get the best of both.

Not sure how I feel about the corkscrew spiral. Would be cool if it actually animated but obvs that's not an option.

Looking at the two side by side I notice that the current one is much more assertive in shading and giving a sense of depth to the various components (cf. especially the area around the smaller piston things where the shafts stick out from their darker surroundings). Can we have that back?

Definitely more moiré jitteriness when moving but not intolerably so. (My subjective preference overall is for the current patch just because this new one is so visually busy but I do recognize the old one does stick out a bit with how clean and basic it looks and this addresses that.)

Can you try running this through pngcrush or something? The *uncompressed* Doom graphic is 17k and this is 30k, 6 times the size of the current compressed png.

DOOM00
DOOM01
DOOM02
DOOM04
DOOM03

mc776:
> Looking at the two side by side I notice that the current one is much more assertive in shading and giving a sense of depth to the various components (cf. especially the area around the smaller piston things where the shafts stick out from their darker surroundings). Can we have that back?
image

KorpKat:
I recall putting depth on the smaller pistons. I'm not entirely sure why i removed them but i don't think it needs it regardless, the texture already feels detailed enough, i don't want to overdetail it.

About the stripes i guess it does get a bit annoying, i could try to fix that. Not sure about the png compression, i don't want the texture to look crusty and lose detail

selliott512:
> > Not sure about the png compression, i don't want the texture to look crusty and lose detail
> 
> No need to worry about that; PNG compression is lossless, and tools like `pngcrush` and `oxipng` will not affect the image quality in any way.

That is true. `pngchrush` deletes irrelevant PNC chunks, and it tries to compress the same thing better.

A while ago I added make target `make fix-deutex-pngs` which standardizes the PNGs by extracting them from the `wads/freed*.wad` IWADs that are built, and copying them back to the build tree. I find that the compression is almost as good as `pngcrush` and it completely standardizes the PNG chunks so they are all the same (no confusing gamma correction that gets ignored, no incorrectly indexed images, or whatever). Do we want to consider that now that the release is out? I know it touches a lot of files. Maybe a discussion for #development.

mc776:
For what it's worth, I tried to combine the new shafts with the old panels and I liked the results - basically no moiré issues in vanilla though it will need to be tweaked since it looks a bit asymmetrical for some reason:

plat2_1
DOOM00
DOOM01

KorpKat:
That looks way too bland and i feel its not quite fitting with the current art direction, maybe i could just merge some of the plates so it doesn't look too busy.

mc776:
The main problem is the thin horizontal lines forming the highlights - they look terribly jittery in vanilla.

(A few aquatex textures have this problem - one was recently fixed and the rest most mappers know better than to put them on huge areas where it'll be noticeable. That is not the case with PLAT2_1, where there's no such issue in id (attached for reference) or current FD, so you'll see things like E4M8 where if you tried using one of the horizontal AQRUSTs it would look atrocious.)
PLAT2_1_id

KorpKat:
Should that really be a concern? it wont look that way in any other port so, as long as it runs and looks decent on vanilla it should be fine, i feel like this is more nitpicking and isn't something that i can really fix either without making the texture look blurry.

mc776:
we're ***explicitly*** aiming for vanilla

we've talked about this

KorpKat:
That is entirely correct! and the texture i've proposed works in vanilla with no errors nor graphical issues whatsoever,. What i understand about your criticism is that you want the texture edited to remove shading so that just looks lightly less choppy in the vanilla rendering (seen on your example above) which is something i would prefer not be heavily altered so as to make the platform stand out more as such, i am willing to hear your suggestions, but reverting it back to an undescriptive grey wall is (in my opinion) a really crude and lackluster way of fixing it.

mc776:
Try lowering the contrast on the highlights on the panels near the top.

mc776 (dismissed):

selliott512:
Maybe I'm overly concerned about the physics of things for this type of game, but what role do the apparent new threads play? They have to be stationary relative to the platform, so they fit into a threaded socket in the base that rotates?

Anonysussy:
I decided to test out the new PLAT texture via both Chocolate and Crispy, it got red from a certain distance.
Screenshot 2024-03-30 183719

Anonysussy:
Decided to come up with a quick fix for this one, i think it should fix the red texture when viewed on vanilla
plat2_1
Screenshot 2024-03-31 160237

Snoucher (approved):
It's a plain upgrade from the original. Nothing bad will happen if we merge.

Anonysussy:
Did this one accordingly to the new PLAT design.
aqf068
Screenshot 2024-04-06 201552

Xindage (approved):

PR #1339, opened 2024-02-14
patches: increase contrast of RW18_1 and 2.

### LINKED ISSUE #1334: patches: consider increasing contrast of RW18_1 and 2.
Posting this as an issue instead of a PR because it may well be just a matter of personal taste.

This has bothered me for months now. I've  tried to get used to it but it really does feel like the new patch with its *very* light shadows is out of place in FD:
lowcontrast
I know the cracks are intended to be shallow but my brain just Will Not Work With It and instead of a difference in depth it just senses an artistic style clash.

Here's an attempt to increase contrast and gamma while reducing brightness so the overall shade would sorta roughly match:
highcontrast

If folks agree, here are the altered images:
rw18_1 rw18_2

mc776:
Closes #1334.


Talon1024:
There's a [screenshot in the issue](https://github.com/freedoom/freedoom/issues/1334#issue-2128235329).

KorpKat (approved):

Anonysussy:
While I'm still at it, here's a slightly cleaned up version of RW18_2 I've worked on
rw18_2

mc776:
Before and after. (Using the first one - lightening the back is going to reintroduce the problem this was intended to solve)
DOOM00
DOOM01

DOOM03
DOOM02

mc776:
The lighter one for comparison - it's not too apparent from the static shot but as you move towards it the darkest parts go from black to grey in a way that makes you wonder if there's something wrong with your gamma:
DOOM01
DOOM00

mc776:
> the darkest parts go from black to grey in a way that makes you wonder if there's something wrong with your gamma

The issue for me is that these parts were completely black and if they're pure black they stay black, but when they're dark grey and lighten up as you approach you still see the solid non-black but it's all a flat colour with no additional detail visible.

Here's an attempt to add more detail inside those cracks:
DOOM01
DOOM00
rw18_2

mc776:
> This one looks better, I would like this to be committed

I did this shortly after this comment, so unless something seriously went wrong with the upload that I can't see I think this should be ready.

PR #1340, opened 2024-02-14
levels: jump-proof Map19 soulsphere secret.

### LINKED ISSUE #1337: MAP19 jump compatibility issues
At the start of MAP19, some mods such as Trailblazer and DN3Doom make it possible to jump onto this ledge to grab the ectoplasmic surge (Soulsphere):

Screenshot_Doom_20240213_114727

It is also possible to jump up there, even without mods, if you circle around to the other side. And once you're up there, you get teleported straight to the second last room when you try to jump off the ledge.

I recommend making both sides impassable, and add the A-RAIL midtexture.

mc776:
Closes #1337.


selliott512 (approved):
Looks good. I tried just the changed area (not a full play through) in both GZDoom and Chocolate Doom and it worked.

It may be a bit touchy if the player is super cautious in approaching the soul sphere, for some reason, but it should work fine most of the time. If you wanted you could have the teledest, soul sphere and then an explicit teleport, but that may be tight. What you have seems fine.

I didn't see any other structural changes with my usual animated GIF. Eureka passed.

mc776:
> It may be a bit touchy if the player is super cautious in approaching the soul sphere, for some reason, but it should work fine most of the time.

I did think about that... if that happens then the next thing that happens is that the player tries to find a way out, then they'll run over the line anyway.

PR #1341, opened 2024-02-16
Fix Picklehammer's credits

unithecutie:
Truncate last name per request, update email, remove space. MIDIs also modified to change metadata to also go by these.


Xindage (commented):
Seems good but why the mid files were changed as well?

unithecutie:
> Seems good but why the mid files were changed as well?

They contained Picklehammer's full last name in the MIDI metadata, and he specified to remove it from everything if possible; so I did.
image

mc776 (approved):

selliott512 (approved):
Looks good, but I have some observations about the metadata that I'd like to share.

Using `timidity` I converted the five changed MIDIs to WAV to verify that there are no changes to the actual music, which was the case - great.

For the metadata I confirmed that you fixed the last name and the email address for all five MIDIs - great.

However, based on the standard output of `timidity`, when either playing or writing to a WAV file, with this change there are now some blank `Track name:` fields. For example here's how the output of `timidity musics/d_dm10.mid` was impacted with this change:
```
--- b/d_dm10.txt	2024-02-21 08:16:47.503611566 -0500
+++ a/d_dm10.txt	2024-02-21 08:14:31.929900007 -0500
@@ -12,12 +12,12 @@
 
 Text: Synth Strings
 
-Track name: Midi's for freedoom
-Track name: Tyler <Last Name>
+Track name: MIDIs for Freedoom
+Track name: Tyler P.
 Track name: "picklehammer"
-Track name: pickle_hammer@hotmail.com
-Track name: Member of ACMT
-Track name: Atomic Chicken Music Team
+Track name: picklehammer@msn.com
+Track name:  
+Track name:  
 Text: Drum line
 Output musics/d_dm10.wav
 (Conductor Track)Playing time: ~189 seconds
```

Both before and after your change the metadata does not contain information that it seems like it should based on `CREDITS-MUSIC`. For example, `Paranoia` is nowhere to be found. It's not something you did with this change, so don't worry about it, unless you want to.

Contributors should be aware that we don't have a good way of changing how their name appears in git log messages that have already been merged to master (not that you have to do anything, just pointing it out).

mc776:
Which just means the faster this gets done the sooner the deadname gets buried... can someone merge this already?

selliott512:
> , why did you edit Matt's message instead of just commenting?)

Good question. I guess I hit "edit" instead of "reply" by mistake and preceded to act as if I was replying. It looks like you understand what I meant though.

PR #1342, opened 2024-02-18
levels: tweak Map08.

mc776:
The four serp platforms never needed blocking lines of any sort, let alone totally impassible. They are now unflagged for blocking anything.

The nearby soulsphere secret has 2 flicker sectors sharing a common apparent light source that go out of sync. This has been simplified.

Removed some useless linedefs that seemed to be leftovers from when this map had more detail sectors.

Fixed the mismatched flat below the bars in front of the exit.

This map has been notorious for being difficult to navigate. It's actually really clever in its layout but it's also a little too clever about trying to surprise the player by putting important things in places the flow of the map would naturally direct the player away from. There's no real curing this but at least the signposting could be better:
- The raised screens in the starting loop area now align with the bricks and can't be climbed by accident.
- The two big remote-only doors near the start now have different textures since BIGDOOR1 implies a directly usable door literally everywhere else. They also have those pump-piston thingamabobs that show they're affected by those switches.
- Made the courtyard secret a little easier to find (if not actually figure out), and made its panel different from the starting area one because that confusion after you teleport is fun only once. Lower unpegged those doortraks as well.
- Raised the window railings so they're more clearly not intended to let you run out.
- Brightened the area around that door with the underpillars since the flow of the horseshoe makes it very, very easy to run past and not notice until it's too late.
- The "door" pillars on the central switch hub are now opened by nearby switches instead of being directly used. As pillars they are coded very similarly to the remote-open pillars everywhere else and it's always this off-putting and demoralizing distraction to have to remember that they are not. The keyless door also gets the same treatment for consistency.
- The door leading to the teleporter out of the switch hub is now opened by the keyless switch, ensuring that you have to have done this at least once before you can proceed without backtracking.
- Exit bars are explicitly marked as needing the red key.

DOOM00
DOOM01
DOOM02
DOOM03
DOOM04
DOOM05
DOOM06


unithecutie:
> This map has been notorious for being difficult to navigate

Can definitely agree on this, I find myself running around like a headless chicken in this map; shocked to learn that this map was originally made for MAP02 as well...
<br>

<br>
Surprised it hasn't received any changes to ease navigation until now.

mc776:
Screens of what got changed in this commit:
DOOM01
DOOM00
DOOM03

Xindage:
ill review this one personally before aproving.

X64X2 (approved):

selliott512 (approved):
Looks good. I did a real play though UV-MAX (no cheat codes) to 100%.

I like the way the room works with the colored switch room works now (around sector 159). I also have always thought the soul sphere secret was too hard, so I'm in favor of the step.

One small thing, if you could fix the missing texture above and below apparent control sector 40 that would be great.

selliott512:
As usual I hope my blinking GIF that highlights the changes is not too annoying. Let me know if that's the case:

map08

selliott512:
@Xindage I'm happy with this. Feel free to merge after your review.

Xindage (approved):
While playing i dont noticed anything major... unfortunately i remenber the progress of this map so i know where and when to go, so i dont find it any different, and when i looked at the changes it was kinda minimal. so FOR ME because like i told i remenber the order of this map perfect i dont fell much change, but i agree the doors are more readable now.

PR #1344, opened 2024-02-20
Fix a few map authors

unithecutie:
These were symlinked under the wrong authors (mainly just very small fixes), _**why???**_

I'm honestly confused how none of this was organized when the project was conceived...


selliott512 (approved):
Looks good, but I can only comment on the formatting - no problem. As long as you know that it's accurate, and its how these folks want to appear in the credits, then great.

X64X2 (approved):

Xindage (approved):

PR #1346, opened 2024-02-24
patches: Revised SW3

Anonysussy:
Revised SW3 for the added consistency.


mc776 (approved):

Xindage (approved):
i noticed tht its a slight darker, and the png was compressed in pallete as well, ok commiting.

mc776:
It actually looks and feels like a lever you're pulling down now!
DOOM01
DOOM02

PR #1347, opened 2024-03-01
Better-fitting E1M1 track.

jadenquinn:
This track is a duplicate of Map07's midi, however I feel it fits much better as an E1M1 track than the current one this replaces does.


mc776:
I think the intention here is to move the current Map07 "Light 'Em Up, Boyz!" to E1M1. Presumably to avoid duplicates "Stanky Leg Specialist" would be moved to Map07.

I think "Light 'Em Up, Boyz!" doesn't quite start with the same kind of energy we want for an E1M1, and the good epic fight parts of it are really wasted on such an early level.

jadenquinn:
> Huh? We've specifically been trying to avoid duping tracks and replacing said duplicates for ages now. Why **_replace_** this track as well?

Because the current E1M1 track sounds like it needs to be in a terry wad. It's just way too god damn goofy, and won't fit in any normal Doom level, at most it would fit a secret level.

Can we stop getting bad tracks for Map07? I think someone just needs to make an entire new midi for E1M1 AND Map07. This is just a temporary patch so that can happen. Neither tracks match their map rn.

jadenquinn:
> I think the intention here is to move the current Map07 "Light 'Em Up, Boyz!" to E1M1. Presumably to avoid duplicates "Stanky Leg Specialist" would be moved to Map07.
> 
> I think "Light 'Em Up, Boyz!" doesn't quite start with the same kind of energy we want for an E1M1, and the good epic fight parts of it are really wasted on such an early level.

Are we even talking about the same midi? This is Map07's current midi, to be clear. There aren't any "good epic fight parts" in this track, it's a nice laid back midi, but has great fast paced+slow paced rhythms, which is great for an E1M1 map. Purely objective. The volume doesn't ever get too loud, it's a good pick for E1M1, or maybe E1M2, in fact I think it would fit better in E1M2, especially since the current E1M2 is crapola, maybe we merge this, and then move it up into E1M2 when an actual E1M1-worthy fitting midi gets made.

MerrittK:
Honestly, it was a bit jarring booting up the newest freedoom after a while not playing, and hearing stanky leg instead of the older track.

jadenquinn:
> Honestly, it was a bit jarring booting up the newest freedoom after a while not playing, and hearing stanky leg instead of the older track, though I guess the older one was a remix, huh.

Oh wow, I didn't know that one was a remix, but yea.

PR #1349, opened 2024-03-03
New e2m1.

Xindage:
This new map is more simple and linear, good for starting slots. There's no new monsters introduced, all monsters present there so far is present in episode 1. the secrets are simple but clever so players does not have to hunt for them so hard at episode start. Theme chosen is plutonia, and that fits for ruin introduction. Map can be beaten in around 1 to 4 minutes, if you hunt or not for secrets. and as final the map already has full deathmatch support.

Screenshots:
image
image
image
image

I can UV-MAX (Not professional speedrun) in around 2 minutes
image

By last i have a video showcasing the deathmatch support as well: (low quality because my laptop has low specs)
https://www.youtube.com/watch?v=6lBe6rMVpSk
NOTE: Its deathmatch layout does not represent its coop/singleplayer gameplay, both are on the same map but the deathmatch support works only in deathmatch.


mc776:
Gave this a try just now, with and without monsters. I do not consider this map an improvement over the current one in any way, but I admit I am biased.

It should have a different name - "Elemental Gate" fits this about as well as "Warped Elementality" fit the *current* Map27.

Aesthetically it feels like a FreeDM map.

Again with your maps, I have the hardest time finding anything - the with-monsters run wasn't too bad but vital stuff always seems to be tucked away in ways I always find counterintuitive. The level feels much longer than it should be. At no point did I feel I had any clear idea what is a wall and what is a door and what opens what where.

Xindage:
If you get 100% of the secrets before actually reaching thevexit then i consider it an actual success because it's intentional, i dont make hard to be find, and the way of you tell that you get lost is interesting to hear because there's just 1 single way to go, tou can't go anywhere else since it's a linear path, even if you go somewhere else you still have to go on the main route.
Now about your comments about the doors, is kinda interesting because like the map has just 2 doors and one of it being a secret that requires a backtrack. To be honest there's just one secret that does not requires a back track. So let's take exactly what you say "at no point did I fell I had any close idea what does what is a wall and what is a door and what does opens what where.". So there again yet 2 switchs in the map (WOW is that many that fell like a hexen pluzzle) joke aside, again there's 2 switch, one main and explicit visible and one secret that is lighted just to be secret, as soon you press the main switch a pillar will lower and the main route will be available, you can see the pillar lower from where you use the switch, it contain just ammo so you can exactly know what the switch does going there you can see that the main way is open. Like there's nowhere else to go, that lead direct to the exit where there's clearly a wood wall texture that is impossible to miss and tava exit.
Both secrets are on a point where you can see what it does if you pay attention, so again if you find them all you have been pretty much observant and not lost. Unless if you have a internal urge in visiting the starting area every time you use a one of the 2 switchs in the map to see what it does, instead going where you is clearly a main path.
I dunno, you play in a very uncommon way, I've send this map to both, Charlie reef, korp, Iredmc and uni all of them played it and told me that the map seemed strong and clear.
Now you say it looks a freedm map... that's kinds impossible since the main layout is terrible for deathmatch, I just make it possible but that definitely not ok for freedm in a slight.
But I should have already expected it from you, be it simple be it complex you always get lost while playing, maybe original doom e2m1 could be way confusing to you as well.

Anonysussy:
Imo I dig in the look of the map, my suggestion is to add more enemies, from the look of it, the enemy is placed pretty sparsely (I haven't fully played the map to finished it yet so this could be wrong)

Xindage:
Iredmc just posted a video as well: https://youtu.be/oZLoxe2WRUM?si=CHQFBNDJvXH-G-zN

jadenquinn:
iredmc here: Feels like a Plutonia map without the difficulty. I'd say add a few tougher monsters than tier 1s, and then it's short, solid, decent.

SarahMagnolia:
I think this map is an improvement over the most recent one in 0.13. Much more straight-forward and easy, yes, but the overall geometry and layout is more organic and entertaining while the current i feel is rather symmetrical and bland. E2M2 isn't particularly challenging either, albeit harder than this map, so being an easier map contributes to more of a sensible difficulty curve, which i find to be a good thing. Secrets are well hidden and helpful, allowing an early chaingun going into E2M2. 

     Current E2M1 I feel is excessively difficult for its slot, however, and not in the right ways. It borders on being a hot-start with the player surrounded by enemy resistance from every direction. Tons of lost souls, Imps, and shotgunners, with Cacodemons as back up, and nothing but a shotgun in the starting area. Inexperienced players can be easily overwhelmed, and even experienced players can waste a lot of time trying to kill all those enemies without heavier weapons. It's, in my opinion, a rather grindy and unfun combat experience, and too radical a difficulty spike, with an uninteresting layout. This is a much better introductory map by comparison.

Xindage:
Did some changes:
Increased the numbers of enemies in each skills and make few thoughter ones appears instead on hard plus some extra ammo.
Nunbers of enemies: E: 21 N: 33 H: 47 -> E: 26 N: 40 H: 57
Wooden barricade by a red keyed door, the pillars that lower also reveals a key, if this does not tell you the MAIN WAY of the map i dont know what could do.
The vines on the windows near the exit was changed by rails for consistency, since you can walk thought vines on the first secret.
The teleport to the secret soulsphere was made way more visible than before and is activated more easily.

Here's the revision 2.
[e2m1.zip](https://github.com/freedoom/freedoom/files/14512205/e2m1.zip)

MerrittK:
I love it. It feels a lot more organic, and it doesn't hotstart in the middle of a swarm.

Xindage:
Underground lava? The previous version does not have lava.

Anonysussy:
Screenshot_20240308_170928
Maybe this is the lava he meant?

SarahMagnolia:
No.

By the time the architecture and enemy placement is changed to be more beginner friendly it's an entirely different level anyway.

MerrittK:
For what it's worth, the current E2M1 is a mess.

PR #1352, opened 2024-03-10
New tracks for MAP20 and MAP22 

Josephus-DH4050-Astartes:
New tracks to replace MAP20 and 22.


Josephus-DH4050-Astartes:
then maybe we can swap the two? idk man, tbh I really don't like alien waltz, and my friend didn't either so I decided to work on a replacement.

Josephus-DH4050-Astartes:
But if I may give an idea, maybe swap my map22 replacement back to MAP20 and replace MAP22 with White Lights?

KorpKat:
I think his new track for map22 is a lot more "general doom vibes" than alieny and strange, being as objective as i can. I believe freedoom aims for a more synth-rock ost, not to mention it is kind of repetitive as it only has two chords for the most part. It's a solid track, but i don't think it particularly fits for freedoom nor cluster 3

Josephus-DH4050-Astartes:
in my personal opinion, freedoom seems more power metal during the more alien levels rather than synthy but that's just my personal perspective.

Josephus-DH4050-Astartes:
plus it seems like freedoom's trying to be "kid friendly" when it shouldn't be. But that's probably my Quake 2 and 4 talking

Josephus-DH4050-Astartes:
Anyway, i'll probably just close this and just make a PR to for remastered MAP20 qnd Remastered DM26.

PR #720, opened 2021-02-21
FreeDM overhaul by Xindage.

Xindage:
All edits are based on the suggestion made by Decay on Freedoom forum.
Please read dm_changelog.txt for details.

Do not merge yet. This PR was made for testing purpose.

PR #1355, opened 2024-03-14
Freedm partial #720 PR

Xindage:
This PR does not include map changes, just map modifications that were already planned for me for a long time, so i'm releasing them so they would not get lost in time.
I tested it and like i told no maps is replaced in this PR and can be safely merged, it just improve the mentioned maps, extra details area also available on each commit.

The remaining data from #720 still need to be worked on to be forwarded into freedm.


X64X2 (approved):

Xindage:
the other maps are coming next

PR #1360, opened 2024-03-26
New, better fitting E1M1 song.

Josephus-DH4050-Astartes:
Thanks to DAR for his help! I like the current E1M1 but in no way does it fit as an E1M1, unsure where to move that to., since it is a good song I'd hate to see it go to waste.


Snoucher (approved):
I checked both the proposed and current and I think the proposed is better. The current doesn't give the vibes that an e1m1 should give, instead it gives more of a "You've found a secret level!" vibe.

X64X2 (approved):

Josephus-DH4050-Astartes:
thinking on it, I definitely agree, stanky leg specialist does feel like a secret level track. but yeah! thanks!

Josephus-DH4050-Astartes:
> I think it'll be better if Stanky leg specialist being moved over to E1M9 while the current E1M9 song be moved to E1M3 rather than being removed entirely, i tested it out and i think it worked fine enough, (also i don't like the current E1M3, the intro is a derivative of E1M8 with a really long buildup and overall kind of forgettable)

Fair enough, but i still have bias towards "Alone with Colors" but fair point I suppose. I wish DM27 "White Lights" was moved to somewhere in Phase 1 or 2 though. Bit that's also a personal thing.

jadenquinn:
I want this closed. Lola has a new awesome track I want added instead of this one.

Josephus-DH4050-Astartes:
> I want this closed. Lola has a new awesome track I want added instead of this one. I thought I already closed this pull, but... it still says open and I've heard it's still causing problems.
> 
> Edit: nevermind, GitHub is so god damn confusing, this isn't even my request

Send me it in discord and I'll concider closing this one

PR #1365, opened 2024-04-21
Shorten DSCYBSIT (assault tripod wakeup sound)

KukoBerry:
Self-explanatory title. Only made sound slightly faster and cut some low-volume parts in the beginning and end.


Xindage:
ok i check and its better still i cant PR it since it failed, maybe something is not right.

GeorgePieVG:
Converting the sound to Doom format and back to wav via SLADE3 fixes the issue

KukoBerry:
on it! thank you for the guidance

KukoBerry:
Just commenting here that the file's already fixed, and it can be PR'd now @Xindage

Xindage (approved):

PR #1366, opened 2024-04-21
Replace Non-free DSDMPAIN

Donowa93:
The current DSDMPAIN is sourced from perkristian's high resolution Doom sound effects pack. Specifically, it's that pack's DSPDIEHI at half speed. This commit replaces it with an edit of me screaming.


fragglet (approved):

mc776:
Just heard this ingame for the first time. It's definitely an improvement over the old one!

PR #1368, opened 2024-05-01
Update playpal-base.lmp

Donowa93:
based off a palette by ugljesa11ugi in the freedoom discord. this palette is supposed to be doom compatible, while maintaining as few colors from the og doom palette as possible, as well as adding hueshifting based on brightness.

nugg0077
nugg0074
nugg0069
nugg0067

do note that this hasn't been tested with building just yet, as i haven't figured out how to build the repo.


Talon1024:
I think you should wait until #1108 is added. That script will automatically convert all of Freedoom's graphics from the old palette to the new palette.

Calinou:
This looks interesting, but wouldn't this break the appearance of some mods when run with a Freedoom IWAD?

Donowa93:
> This looks interesting, but wouldn't this break the appearance of some mods when run with a Freedoom IWAD?
the palette is laid out in the same way as the doom palette, and i've made many changes to try and maintain compatibility with the vanilla pal while still having it be visually distinct

> Other than that, I find the plasma color is a definitive upgrade for sure, but the contrast ratio for the overall scene rendering seems pretty high. This results in some maps like E1M1 looking quite dark in the starting area. I'm viewing the images on an OLED display in a dark room, so on a display with worse viewing conditions (e.g. LCD panel in the sunlight), it might look better due to the relative lack of contrast.
i mean i did most of the work on this palette on an lcd in a brightly lit room, but the contrast seems to match the doom palette
newpale1

oldpale1

fabiangreffrath:
To be honest, I don't think this palette is much of a win. It seems to me that it merely replaces some of the warmer brown/beige shades with some colder blue-ish ones. The whole scene feels a bit colder and blue tinted. I feel the need to enable cold light reduction on my screen to mitigate it, but this pretty much cannot be the intention of a new game palette.

Donowa93:
> To be honest, I don't think this palette is much of a win. It seems to me that it merely replaces some of the warmer brown/beige shades with some colder blue-ish ones. The whole scene feels a bit colder and blue tinted. I feel the need to enable cold light reduction on my screen to mitigate it, but this pretty much cannot be the intention of a new game palette.

it's probably due to how doom is a very dark game and the colormap having no ranges for ranges becoming brighter (unlike quake's colormap) while the palette i have here hueshifts to yellow in light tones and blue in darker tones
in earlier versions i tried making the yellow hueshift more extreme with bright tones, but i scrapped those after i realized that it turned the blood on zombiemen blue (since this is supposed to be doom compatible, i was testing it with different games than just freedoom)
pissblood

Anonysussy:
Screenshot_20240509_095703
I wouldn't say that I hate the palette, but I've play test this palette for probably a week, I think I agreed that the grey palette should be actual grey, they're the most basic colors you're gonna get, your palette wouldn't cut it if it didn't nail the most basic colors, considering how doom PLAYPAL and COLORMAP works, it just make everything bluish and slightly yellow tinted in few areas, it's either look cold blue or Sephia most of the time and it looks off, it made them look like they just tinted for the reason of being tinted and not much more.

X64X2 (approved):

Donowa93:
while i do admit that i might've made the hueshifting a little too strong for the grays, i think your edit only exacerbates the problem of the grays being tinged by the hueshifting
once i get back home i'm thinking of tweaking the palette though, mainly thinking of reducing the hueshift for the red ranges (or maybe even having the red ranges hueshift to brown at the very low end)

Anonysussy:
> while i do admit that i might've made the hueshifting a little too strong for the grays, i think your edit only exacerbates the problem of the grays being tinged by the hueshifting once i get back home i'm thinking of tweaking the palette though, mainly thinking of reducing the hueshift for the red ranges (or maybe even having the red ranges hueshift to brown at the very low end)

I whipped that edit up pretty quickly, yeah it looks worse in comparison, i think the green one is fine and colormaps better as far as from my experiences goes (e.g Reelism 2 HUD), i would still prefer normal greys though, other palette other than greys were pretty passable though.

Donowa93:
i'm thinking of also reducing the saturation on the gray tones, i still wanna keep the blue/yellow shift since i like how they make the light and dark gray textures more distinct while still being recognizably gray

Donowa93:
just updated, posting this so that it shows on the discord
nugg0120

Xindage:
the last screenshot seems more of an acceptable way, can you give us more screenshots?

fabiangreffrath:
> the last screenshot seems more of an acceptable way, can you give us more screenshots?

Yes, before/after screenshots with the original palette would be best.

Donowa93:
> > the last screenshot seems more of an acceptable way, can you give us more screenshots?
> 
> Yes, before/after screenshots with the original palette would be best.

by "original palette", do you mean the doom palette or the older proposal?

fabiangreffrath:
> by "original palette", do you mean the doom palette or the older proposal?

The Doom palette is the one this proposal needs to perform against.

PR #1371, opened 2024-05-26
New E1M1

CharliReef:
While the old E1M1 is fun, I feel it's a bit too much at once, and in no particular order, not to mention leaving the rest of the episode without much to surprise the player or shake things up. I find it to not be a very good introductory map, and so I made this. This map is simpler, with secrets to unlock your arsenal for players that look beyond those corridors on the surface. But the enemy variety is stripped back slightly, and introduced at a slower pace, that way pinkies, specters, and shotgunners later introduced can serve to keep the gameplay fresh throughout.

The old E1M1 is not gone. I found the level rather fun, and so I have created an expanded version, not found here, but to be potentially added later on once polished enough for me to put it up, hopefully to be used somewhere around the mid point of E1.


Josephus-DH4050-Astartes:
:D

Anonysussy:
> I agree with you, E1M1 is cool but it's a bit much for the first level. I just played through your version, it's definitely a much smoother introduction to Freedoom, and does a better job allowing a new player to figure out the game mechanics. I like this map a lot. I do love the current E1M1 though, so maybe we don't have to completely replace it, but just add your level before Outer Prison.

Imo the current e1m1 had to undergone massive changes if it were to be e1m2, it wouldn't transition much smoothly from this map to the current e1m1 and the current structure of current e1m1 just screams first level imo.

42manako:
> > I agree with you, E1M1 is cool but it's a bit much for the first level. I just played through your version, it's definitely a much smoother introduction to Freedoom, and does a better job allowing a new player to figure out the game mechanics. I like this map a lot. I do love the current E1M1 though, so maybe we don't have to completely replace it, but just add your level before Outer Prison.
> 
> Imo the current e1m1 had to undergone massive changes if it were to be e1m2, it wouldn't transition much smoothly from this map to the current e1m1 and the current structure of current e1m1 just screams first level imo.

I can see what you mean, the level does have that feel to it, to try to draw the player in
I still think this level would make a better E1M1 though (to me)

42manako:
[screen-recorder-tue-nov-05-2024-18-16-43.webm](https://github.com/user-attachments/assets/a1c56053-cf4f-4bf1-bc14-b7095006da0f)
I found some kind of glitch, not sure if this is a level problem or a problem with the application i'm using

PR #1373, opened 2024-06-02
levels: New map30

GeorgePieVG:
Sketched out a possible replacement for map30, now it's not just an Icon of Sin clone. Inspired by an older longer version of the current map30. Definitely not the final version and it needs edits such as balance changes. This is also my first map I'm posting publicly, so criticism is very welcome.


Anonysussy:
i like the idea of the new map, it felt like a mini version of the old map30, though the crusher felt like they just sorta came out of nowhere, the left room is more tolerable since it gave you some head start, the right room is quite punishing with it as the crusher just came out of nowhere and it didn't give any indication that it'll going down to crush the player, the double icon of sin boss fight is pretty cool though and put a unique spin to the icon of sin boss fight, though it may caught first player and players from the previous versions off-guard due to the different method of beating the icon of sin, still pretty cool though.

GeorgePieVG:
> i like the idea of the new map, it felt like a mini version of the old map30, though the crusher felt like they just sorta came out of nowhere, the left room is more tolerable since it gave you some head start, the right room is quite punishing with it as the crusher just came out of nowhere and it didn't give any indication that it'll going down to crush the player, the double icon of sin boss fight is pretty cool though and put a unique spin to the icon of sin boss fight, though it may caught first player and players from the previous versions off-guard due to the different method of beating the icon of sin, still pretty cool though.

Thanks for your feedback, I appreciate it. I agree about the crushers, however, if they should be replaced, I'm not sure what type of trial to make for the right room. I might do something like getting the blue key in Blood Ember Fortress (map 20), where enemies teleport into the room and you have to wait for the key to appear. Regarding the confusion about how to defeat the boss, that's why I made the starting room where you have to shoot the skull-switch with a pistol/chaingun.

Anonysussy:
> > i like the idea of the new map, it felt like a mini version of the old map30, though the crusher felt like they just sorta came out of nowhere, the left room is more tolerable since it gave you some head start, the right room is quite punishing with it as the crusher just came out of nowhere and it didn't give any indication that it'll going down to crush the player, the double icon of sin boss fight is pretty cool though and put a unique spin to the icon of sin boss fight, though it may caught first player and players from the previous versions off-guard due to the different method of beating the icon of sin, still pretty cool though.
> 
> Thanks for your feedback, I appreciate it. I agree about the crushers, however, if they should be replaced, I'm not sure what type of trial to make for the right room. I might do something like getting the blue key in Blood Ember Fortress (map 20), where enemies teleport into the room and you have to wait for the key to appear. Regarding the confusion about how to defeat the boss, that's why I made the starting room where you have to shoot the skull-switch with a pistol/chaingun.

i definitely dig the idea of having the player wait for the key to appear, you could also add up extra contrast for the room, like having the middle room more lit up while the stairway more darker.

Xindage:
This map really manage to make me angry, dude seriosly, i cheated fly to win, this map has some serious flaws that need to be adjusted.
First aside getting 200 armor i see absolutely no advantage in doing both sides of the main room and the final room, my RNG loves me so much to give me 2 archviles in a row in there, the 3th run i just went god and fuck.
Now about the sides... crushers... like there's no other options that we can go aside crushers, both side seems to rely on this + visibility limitations, i cant see a thing properly and the crushers wanna me rush thought the maze.
I dont had much fun on this map tbh, and is a forgetable as the original map30 is, oh and by a side note, hitting the brain in almost as impossible in gzdoom, its sprite height does not match its sprite and i shot thought it, i was using rocket and they fly thought it and i took sometime to figure why it was taking no damage.

KukoBerry:
Just played through it. I feel like the blue room and its rewards were a bit too small and few compared to its red complement. I agree that the crushers came out of nowhere and were a little cheap. Other than the archvile, I feel like the room was a little underpopulated... Maybe we could do that trick from Final Doom where there's a tower with a hidden archvile infinitely resurrecting a hitscanner of sorts?

I had fun with the red part, the maze felt like a hard but fun challenge. Though the textures were a little choppy, there were some parts where I kept thinking i could go between two red areas, but it was blocked off. The mega-armor followed by the megasphere I feel were good boons for the trial.

Now, the main boss arena was a pretty large difficulty spike. I love the platform being in the shape of an infinity logo, and the idea for two spawners is quite a neat innovation, but GOSH is it hard to deal with them both with Exactly One Single Healing Item (a large medikit). I wasn't quite sure what opening up the brains for the spawners was supposed to do given the brain wasn't behind either of them - took me a little while to realize they opened up the alienvat doors, which were also a smidge hard to hit given the vast influx of enemies. Finally, the bossbrain. It was nice to have it appear like that, but as it has a minuscule hitbox it's best if it has a sort of wall behind it, so rockets' splash damage can get it easier.

Very challenging map, maybe a little too much? Other than completionism it IS a little hard to make someone want to take the red or blue trials, maybe make the boons clearly visible from the start? Like say, instead of putting a bunch of ectoplasmic surges in the main door, put just one and make them part of the very end of each trial, maybe. I think the blue room could have some healing too.

KukoBerry:
image
also, regarding this area - are these parts supposed to be uneven? It adds interesting volume, but I don't think the textures mesh well with it, they're all cut off at the upper part.
If you intended to make it seem like they're slid in and crushed between the marble structure and the inside of the wall, maybe add a bloody marble texture right above it? I'm not sure if that'd work though

FarjFarj:
I find the end battle to be a really interesting concept!  (Note:  I have only played the older version so this might be inaccurate, also I played on hard difficulty for my blind runs)

However, the side rooms before the boss battle just don't have anything worth the trouble, and they are both annoying to go through.  They are both harder than pretty much anything else in FreeDoom, and, at least for me, required planning and throwing myself at the problem before I completed them.  The blue key room even felt like trial and error with a bit of luck before I found out about the secret room via the local area map from the red key area.  (For basically all other parts of FreeDoom on hard, I can handle them pretty easily, although that may be because most of FreeDoom is enemy encounters whereas the side rooms in this stage are more crusher encounters).  Eventually, I decided to just skip these rooms because they weren't fun and weren't worth it and go straight to the boss battle.

Here are my issues with the boss battle.  First of all, it is way too telefrasggy.  I know that this is an outlier, but there was one run where I got telefragged while I was stuck in the teleporting animation or whatever the teleporter fog effect is.  Perhaps this could be fixed by making it so that the teleporters to the switches require one easy jump to get to to prevent monsters from teleporting in.  I think that the many other times I got telefragged are due to both there being 2 spawners as well as the fact that they mostly target the same thin platform (By the way, the platform being an infinity symbol like the FreeDoom logo is really cool!).  

(Also, I should note that I played this map at 320x200 resolution, though I think that accommodating for this should be done not just for Vanillaness but also for people with poor eyesight.)  Another issue that I had was figuring out how the boss worked.  At first, my IoS instincts kept in until I realized that there were teleporters at the bottom of the arena.  After opening up the skulls, I again tried rocketing the skulls.  After a few attempts where nothing happened, I decided to go into GZDoom and use fly to see what was happening.  After figuring out that the skulls housed hitscan switches, I knew how the rest of the battle needed to go as the arena changed.  Even when I had it figured out, though, I sometimes shot the same skull twice in a row due to low resolution, wasting precious time and getting me killed.
Here is what I would suggest is to make the arena clearer:  
1.  Make the shutter (that blocks bullets and has to be raised via switch) bigger/ stick out of the wall more to make it stand out that something is blocking your shots.  
2. Make it more obvious that there is something at the bottom of the arena (like maybe make there be some decorative geometry and/or textures to make players more likely to think that there is something at the bottom and look down.
3. Change the texture of the hitscan switch.  It should be bigger and have a greater difference in appearance once it gets hit, and should not be covered up.  Actually, maybe you could make it so that once a hitscan switch gets activated, some "lights" (like, bright green textures behind mini doors on the wall) turn on (mini doors raise).  Also, here you mentioned that you wanted to use the room at the start to remind players about hitscan switches.  What I recommend is that whatever large, flashy hitscan switch texture you use for the boss battle, you also use in the start room so that players associate that specific texture with being a hit scan switch.
(Also, I should note that I think that the skull textures shouldn't be used as to not make people coming from DooM think that you need to deal with it the same way you would the IoS.)

This is a bit miscellaneous, but I think that the 2 outer pod things that you have to kill be lower so that it is easier to aim at the from the bottom of the arena.  (The player has to be quick in this fight, so the rad suit should still be fine at this point, though this is just an assumption)

Because the elevators speed up after one use, if I didn't know about the rocket's infinite splash height, it would've been very frustrating to hit.  The Boss Brain definitely needs to be easier to hit.  This could be done by just simply not having the elevators speed up (The speedup doesn't really make a difference anywhere else in the fight because you can use the chaingun to insure that you hit the hitscan switches at least I think:  I never missed the hitscan switches once I started using the chaingun so I don't know what it's like hitting the hitscan switches with at the faster elevator speed.)

Last thing I want to talk about regarding the boss battle is that it is way to difficult for FreeDoom.  I failed enough to the point where I started memorizing some timings of the monster spawners.  I don't think that it'd be possible to adjust how the monster spawners work, so I'll talk about adjusting the arena.  Some ideas I have are including other platforms so that the monsters spawned in aren't so concentrated, having only one monster spawner, and/or having a few more resources in the arena.  


Overall, once I ignored the side rooms and figured out how the boss battle worked, I had a lot of fun.  The side rooms need more incentives and need to be more fun, or maybe just scrapped / kept skippable.  I really hope that you continue working on this!  The boss battle felt really cool and climactic.  

P.S.  Maybe instead of 1 hitscan switch, you could have 2 or 3 hitscan switches right next to each other that all have the same function so that the player can more easily use any hitscan weapon in case they are stubborn and only try to hit the target with a certain hitscan weapon.

PR #1379, opened 2024-06-21
Add some texture variations

GeorgePieVG:
Some important UPD: 

the background of the tscrn14 patch is based on the sky pattern from GIMP. Here is what the GIMP license says about patterns: 

> Any data used in artworks, such as brushes, patterns and the like should be
>   under a CC0 license (Creative Commons Zero) or equivalent, i.e. "No rights
>   Reserved", because GIMP contributors clearly don't intend to claim any right
>   on anyone's work just because they used GIMP and its default data.  We cannot
>   clearly give proper licensing on all existing data, prior to the addition of
>   this note in the LICENSE file (for historical reasons, simply because their
>   contributors are since long gone), though we can say that an uncountable
>   number of users have used these for dozens of years and never had legal
>   problems as far as we know (it is anyway unclear whether anyone could really
>   claim any right for very basic brush or pattern usage). Since 2015, we even
>   have a clear FAQ entry to explicitly say the GIMP project has no intention
>   whatsoever to put restrictions on people's work:
>   https://www.gimp.org/docs/userfaq.html#can-i-use-gimp-commercially

So this patch may not be safe to include in freedoom and needs to be redone. The other patches are based only on textures from freedoom and attic. End of UPD.

1. Tried to recreate the big windows by Korp in the old artstyle:

2. Compblue in every color:

Red and yellow will be especially useful for identifying the area for keys of corresponding colors. 
3. Skin version of star set
Screenshot from 2024-06-21
4. Skinless skintek tubes:

5. Fredrik's crates have been restored as a new texture.
6. Red and yellow silver wall. For the same purpose as multicolored compblue.

7. New screens:

8. Warped snake textures via ANIMDEFS. Since this is the original freedoom texture, it won't break mods. 
Screenshot from 2024-06-21
9. Some new flats:

10. Cursor for GZDoom:
doomcurs
11. Polyobject-like door. Even though we only make maps for vanilla, we can use the texture for static decorative doors.


GeorgePieVG:
Added a few purple recolors

Xindage:
i think its too much, here is my point of view: (each number is each one of your topics)
1 - Maybe its ok, not sure if we will ever use but ok.
2 - I dont think we need compblue in ALL colors, red and yellow is kinda of acceptable still.
3 - Does not look that cool, I'm unsure about this one.
4 - :+1: This one could really be of use as patches for other textures.
5 - I remember these crates, it would be nice if these return.
6 - Same reason as 2.
7 - I dont think we need that much screens, the broken and static seems interesting, but since animadefs is not a good option (i'll explain bellow) a screen showing the TV pallette written "no signal" would be more effective EX: 
image
8 - No animadefs please.
9 - The red one lacks bright.
10 - Seems ok having one.
11 - Not sure, seems cool, but may end up unused?
Purple - Its a bad idea, purple would not look good in distance on vanilla because lack of pallette of purples it fade in brown, i would suggest against purples.

mc776:
I've been trying to avoid commenting on aesthetic issues and mostly just agree to disagree, but there is one functional concern here that if we've got these red and yellow variants of these blue silver things there's a natural temptation to use them to mark locks, which creates colourblind accessibility problems dealing with which had been a major selling point of the last release.

If someone absolutely needs variants of these (and tbh I don't know why) it might be better to provide green and purple or green and black/grey (which would keep the cool colour theme the originals were clearly meant for).

Issue #1380, opened 2024-06-22
Sprite lump names need to align with id lumps.


mc776:
First reported here: https://codeberg.org/mc776/LibTech1/issues/6

Some mods use specific sprite frames for certain effects, including custom textures. These break if the sprite in question uses different rotations.

The example mod that prompted this:
https://www.doomworld.com/idgames/themes/university/trinityu
https://www.doomworld.com/idgames/themes/university/trinity2

Where there are very obvious reasons for deviating from id (e.g., the archvile), the buildcfg should include a comment right there warning future contributors not to break this stuff again.

See https://www.doomworld.com/forum/topic/45009-list-of-all-doom2-sprites-needed/ for a convenient list.

*Can be done right away:* (and I'm going to post a PR shortly after this issue)

* **Cyberdemon**
  * Full rotations on all non-death frames (A-G). Revert #1138.
* **Demon**
  * Full rotations on attack and pain frames (E-H). Revert #1138.
* **Archvile**
  * G through Q and the funny characters are all full rotation. Edit buildcfg to refer to the existing 0 sprites.
* **Zombieman**
  * Standard 2/8 rotations throughout. Full rotations should be added to the attic.

*Will need some work - but it's just the grunt work of flipping and renaming:*

* **Revenant**
  * Full rotations on attack and pain frames (G-L).
* **Arachnotron**
  * Full rotations on I frame.
* **Minigunner**
  * Full rotations on *all* non-death frames.

*Will need **quite a bit** of work - non-trivial edits needed for the sprites themselves:*

* **Spider Mastermind**
  * *All* 1 and 5 walk frames are used for both the first and latter half of the loop. Freedoom's large technospider can be made to fit just by making the design symmetrical.
  * Standard 2/8 rotations otherwise, but this means LT1's sprites are backwards.
  * Full rotations on I frame.
  * J frame needs to be zero.

*Will need a **lot** of work - basically re-design the monster:*

* **Baron of Hell**
  * Standard 2/8 rotations on walk frames (A-D). Given the heavy emphasis on the asymmetrical parts a full redsign may be necessary.
* **Hell Knight**
  * Standard 2/8 rotations on walk frames (A-D) *except* A1C1 and B1D1. (For clarity, the baron does *not* flip the 1/5 like this.) Given that the difference will be visible *from the same angle*, this should not be dealt with without a redesign.

mc776:
For future reference here's my workflow for converting a bunch of _2_8 etc. to full rotation:

[...]

PR #1381, opened 2024-06-22
sprites: Match Doom lumps.

mc776:
Addresses #1380.

Reverts #1138, and imports the zombieman sprites from Lotan's Tomb.

Partially reverts #936, but does not restore the old graphics files and just points to the 0 frames in buildcfg instead.


mc776:
I don't know what conventions should be followed for the attic so if this gets merged someone please add this:
[freedoom-fullpossrotations.zip](https://github.com/user-attachments/files/15934868/freedoom-fullpossrotations.zip)

mc776:
Unless someone finds a typo in the offsets somewhere this PR is ready to merge. The nobles and mastermind will have to be dealt with in due course much, much later and should not be delaying a fix that addresses an actual failure to load a vanilla-compatible map.

mc776:
@fragglet since you seem to be looking at FD stuff can you take a look at this? I'd hate for my final FD legacy to be "made it crash on TRINITY.WAD".

mc776:
zip containing the old full-rotation BOSS and BOS2 walk cycles.
[old-asym-boswalk.zip](https://github.com/user-attachments/files/16819292/old-asym-boswalk.zip)

mc776:
I've got a working set of painbro attack animations but it's enough of a change that it might be better to treat that as a new PR.

Which means as far as I can tell this PR is ready to merge.

GeorgePieVG:
> I've got a working set of painbro attack animations but it's enough of a change that it might be better to treat that as a new PR.
> 

Very cool animations, looks more grotesque and less cartoony than it was before. Honestly, I prefer the symmetrical pestmeisters now. You should definitely PR them.

mc776:
I guess I shall once I've finished the rotations (and fixing little things like actual holes in the sprites 💀)... the inconsistency when you strafe rightwards is *much* more visible on these guys than on the gun-toting zombies, just because of how prominent that second head is on the sprite (whereas with the gun guys the animation is much faster, you're probably seeing it through a painflash or from behind cover, the minigunner's constantly facing you and the most prominent visual element is the gun flash which is pretty symmetrical and covers some of the most obviously inconsistent parts)

mc776:
And done.

I stand by what I said about the spider mastermind, though:
> we don't yet have a known map that breaks because of this and it's *really* hard to have any motivation for this when multiple people have done at least some preliminary work for a total replacement anyway, so there's no point in letting that get in the way of more critical incremental improvements.
> 
> Which means as far as I can tell or care this PR is ready to merge.

mc776:
For reference here is the entire sprite sheet I've used containing both the asymmetrical and the symmetrical painbro design, including both the FD painlord and LT golden painlord.
[bossym.png.zip](https://github.com/user-attachments/files/17055560/bossym.png.zip)

PR #1385, opened 2024-07-10
New E2M8 map + Improved readability for the ENDOOMs

Anonysussy:
Imo the current E2M8 is quite punishing for those who were pistol starting the level, i recall running out of ammo a couple of times on the map, it's also mandatory to make one sound to wake up the cyberdemon to obtain a new weapon and progress, this new E2M8 added more ammo so the player could stand a chance and player also fight one cyberdemon at a time instead of two, i also came up with teleporting gimmick for the new map to spice things up, this one also improves on the transition between levels (E2M7-E3M1).
Screenshot 2024-07-10 151355

For the ENDOOMs i was quite unsatisfied with my previous work, so i decided to reworked the entire thing, the new ENDOOMs features the more stylized title and improved readability for some of the text and the DM logo while adding up some color coded text in the process, this one also replaces the copyright year with the version info.
image_2024-07-10_151523179


Xindage:
please split this PR into 2 PRs each one with it own request. and close this when done.

To clarify accepting PRs with 2 different contents is hard to revert if something goes wrong as both changes are togheter.

PR #1387, opened 2024-07-16
Reworked ENDOOMs

Anonysussy:
Separates the ENDOOMs PR from the level PR
347319046-3205f882-d291-4133-a8c9-b15ded3c543f


Anonysussy:
Also im delaying the E2M8 PR, there are still some possible changes and improvements for the map further down the line.

Xindage:
can i see a preview of old vs the new?

Anonysussy:
> can i see a preview of old vs the new?

Screenshot_20240718_063030
349158063-48eda8d5-bf3c-4478-aff0-53c8172f6735

Xindage:
ok i think i found a small issue this endoom includes a versioning but its not automated and im not sure if this can be automated, this way we have to modify this endoom every release we do, this could be problematic later.

Anonysussy:
> ok i think i found a small issue this endoom includes a versioning but its not automated and im not sure if this can be automated, this way we have to modify this endoom every release we do, this could be problematic later.

Imo it's better than having a copyright year as copyright year requires changes yearly where as version isn't guaranteed yearly and could wait for 3-4 years or even more which give enough time for changes to be made for the ENDOOMs.

Issue #1376, opened 2024-06-08
DM09's music causes a crash in DOOM2.EXE


MelodicSpaceship:
When I load FreeDM as a PWAD in DOOM2.EXE under DOSBox 0.74-3, it crashes on MAP09 with the following error:

`Z_CheckHeap: block size does not touch the next block`

Oddly, Chocolate Doom does not seem to have the same crash which might be why this went unnoticed.

By the way, I was testing this on the stable version 0.13.0. If this was somehow fixed in the dev version, then feel free to close this.

MelodicSpaceship:
Just tested this in the latest autobuild, the crash still occurs there.

Anonysussy:
What version your DOOM2.EXE running on? As far as I'm aware, between versions there's some changes made to the engine itself (e.g raised vanilla/static limits or even removal of certain features).

MelodicSpaceship:
I was mainly using version v1.9 of the Doom 2 executable. I checked and this also seems to happen with the Final Doom and Ultimate Doom executables as well.

I'm pretty sure DOOM2.EXE v1.9 is the closest match to how Chocolate Doom runs FreeDM, and it's also the most commonly used DOS executable, so I think it would be fair to fix it.

Also, [according to DoomWiki](https://doomwiki.org/wiki/Error_message#Z_CheckHeap:_block_size_does_not_touch_the_next_block), the error seems to occur when there's bad sidedefs or linedefs with no sector, so the map should be checked for that.

MelodicSpaceship:
Okay, I did a bit more testing. Oddly enough, when I swap MAP08 and MAP09 in a PWAD, the crash still occurs in the MAP09 slot and the level that was in MAP09 no longer crashes. This means the crash has nothing to do with the levels themselves, but rather with the way the WAD is arranged.

In this case, I have no idea what could be causing the crash. It would likely require a change to the build script to fix, and at this point I'm not 100% sure if it's worth bothering with.

MelodicSpaceship:
Okay, I tried one more thing, and I think I found the source of the crash. It appears MAP09's music is what's causing the crash, because when I edit freedm.wad to replace D_IN_CIT with a copy of D_RUNNIN it no longer crashes, so that may need to be changed.

That's also probably why it didn't crash in Chocolate Doom, as that natively uses MIDI rather than MUS like the vanilla executable does.

mc776:
[...]

PR #1391, opened 2024-07-17
Edit DM09 music to not crash under DOOM2.EXE

MelodicSpaceship:
I managed to edit the DM09 music so that it no longer causes DOOM2.EXE to crash, thus fixing #1376. This was done by importing the MIDI in a MIDI editor and re-exporting it.


Xindage (approved):

Xindage:
still i have a question why such action reduced the file size in 98%?

MelodicSpaceship:
I don't really know. The actual file is only smaller by less than a kilobyte, so I think GitHub might've been misreporting the size difference percentage.

PR #1393, opened 2024-07-18
New DOBIGTV patches

SenhorFlibble:
I attempted to make more accurate replacements for TNT's DOBIGTV patches, based mostly on Freed∞m's COMP patches. The current ones are very obviously placeholders. I tried to keep the textures' geometry close to the originals, which was not very complicated for the most part as TNT based them on Doom's COMP patches as well.
dobigtva dobigtvb dobigtvc dobigtvd
Please feel free to make any tweaks if you need to, I'm not really an artist, and these can still be improved I guess.


Xindage:
you should add yourself at the credits.

SenhorFlibble:
Done!

SenhorFlibble:
Added new DOBLIP patches too.
doblip1a doblip2a doblip3a doblip4a

SenhorFlibble:
I've also pushed new DOSHA and DOTV patches for the TNT texture replacement set.

Xindage:
i'll check those when possible.

SenhorFlibble:
Updated the DOTV patches. I did not immediately recognize that TNT's originals were based on RW46_3, and that Freed∞m also has its own version of that patch, so I just transplanted the animated screens to it.
dotv1b dotv2b dotv3b dotv4b

Xindage (approved):
i reviewed all of them seems accurate to what tnt does, tbh i would like if the comps does not have these outlines but i think that the precision to what tnt does is needed, good work.

PR #1396, opened 2024-07-24
textures: add dangling version of big vines.

mc776:
For when you want an upper area of a wall to have vines but you don't want them to unnaturally cut off at the bottom.

It's something I've wanted for as long as I've tried to do any mapping for Doom.

The regular vines have also been tweaked to get rid of some excessive solid colour blocks.


Xindage:
I agree with this idea, I'll test out later.

mc776:
Here's how they look in LibTech1 E3M4:
DOOM00
DOOM0000
DOOM0001

PR #1402, opened 2024-07-30
Complete story revamp

Tiramisu100:
Freedoom's story is in dire need of a revamp because it still follows the old lore of the monsters being AGM experiments instead of aliens (which is the current consensus).

To this end, I wrote entirely new intermission texts for both Phase 1 and Phase 2, as well as a prologue that could be placed in the manual, and a document (story-guidelines.txt) detailing why the new lore looks the way it does (which could be used to maintain consistency and avoid plot holes going forward).

In my story revamp, I tried to achieve the following goals:

- Establish the monsters as aliens to match their current common perception
- Explain inconsistencies (f.eg. why the alien levels look the way they do) and plot holes in a satisfactory manner
- Make the story more dynamic through plot twists and a more active role for the main character
- Carry over as much from the old story as possible (f.eg. Karl being a would-be super soldier who longs for freedom)
- Keep things simple and familiar, avoiding decisions that could be seen as strange or controversial

I am open to feedback regarding both the intermission texts and the general lore detailed in the story document.

### DIFF
--- a/lumps/dehacked/dehacked.txt
+++ b/lumps/dehacked/dehacked.txt
@@ -427,62 +427,74 @@ NIGHTMARE = The game is not designed to be\n\
             really that good - or bored.\n\n\
             (Press Y to confirm)
 # End-of-episode texts, now with a consistent story. Feel free to edit them.
-E1TEXT = Free at last from that dreadful outpost!\n\
-         But you're still stuck on this rock, with\n\
-         AGM and all their monstrous "experiments."\n\
-         Best not to dwell on what evil horrors\n\
-         they were going to create from you...\n\n\
-         Damn alien teleporter ate your gear.\n\
-         The air is hot and muggy, and smells of\n\
-         strange, intoxicating vegetation. You see\n\
-         what looks like AGM corpo tech patched\n\
-         over old, overgrown alien structures.\n\
-         Nothing close to the tech needed for this\n\
-         existed at the outpost, meaning there must\n\
-         be a starport nearby. To Earth. To freedom.\n\n\
-         Just gotta hope that ship will launch\n\
-         from these routed, infested ruins...
-E2TEXT = The massive brutes collapse onto the\n\
-         ground, falling apart bit by bit.\n\
-         The starport is yours.\n\n\
-         But no ship: the launch bay sits empty,\n\
-         leaving only a massive teleporter pad.\n\
-         Limited inorganics. Fixed destination.\n\n\
-         The name suggests it's probably the new\n\
-         AGM ship hub in this remote star cluster.\n\
-         Bastards owe you for what they've done,\n\
-         so what even is interstellar piracy\n\
-         but a bit of D.I.Y. law enforcement?\n\n\
-         You board the platform and fade away.\n\
-         Next stop: Horizon.
-E3TEXT = The abomination explodes into oblivion.\n\
-         What the hell even was that?\n\n\
-         As the shadow sets over this red rock,\n\
-         you realize that you're now trapped here.\n\
-         At least you've got more space in this\n\
-         wasteland than your old cell.\n\n\
-         You kick the corpse in anger, and it feels\n\
-         good. You kick it again, and a remote falls\n\
-         out. You press the button, and the ground\n\
-         gives way to reveal a teleporter. You look\n\
-         at the gizmo - it says "To Double Impact."\n\n\
-         Wherever or whatever that is, you pray\n\
-         to any listening deity that there's\n\
-         a working ship on the other side.
-E4TEXT = Suppressing fire echoes above as the last\n\
-         survivors flee the ship. The launch process\n\
-         is irreversible: killing you now would only\n\
-         bring a slow death drifting through space.\n\n\
-         Fifty-nine. Fifty-eight. Fifty-seven. .  .\n\
-         You melt into the pilot's seat, lost in the\n\
-         voice of an automated assembly of angels\n\
-         delivering you from this wretched rock.\n\n\
-         The ship rumbles as she wakes up;\n\
-         you think of Earth as she lifts off.\n\
-         Hopefully AGM won't find you there:\n\
-         they've got the outbreak to deal with,\n\
-         so that'll give you some time.\n\n\
-         Destination: Earth!
+E1TEXT = As the last brute falls to the ground,\n\
+         you only now notice how much your whole\n\
+         body is quivering. Just what the hell are\n\
+         those things? Aliens? Did AGM really make\n\
+         first contact and hide it from the world?\n\
+         Well, whatever it is they did, it bit them\n\
+         in the ass hard. There's not a single human\n\
+         left alive on this whole rock - and no\n\
+         spaceships to take you home, either.\n\
+         You step onto one of those stupid\n\
+         AGM teleport pads that give you nausea,\n\
+         and arrive at the facility where they first\n\
+         brought you. The place is already in ruins,\n\
+         ancient-looking alien marble peering from\n\
+         the cracks. Ruins built upon ruins.\n\
+         Whatever - the spaceport better still be\n\
+         there, you're in no mood for more detours!
+E2TEXT = Now that these monstrosities are dead\n\
+         and done with, this whole place is yours\n\
+         for the taking - including a shiny new\n\
+         spaceship, and access to AGM data logs!\n\
+         And who would've guessed - these bastards\n\
+         really did make first contact, 10 years ago,\n\
+         and decided to privatize it. They dug up\n\
+         the ruins and found the alien bugs, which\n\
+         they've since bred for genetics and combat\n\
+         research. They also stole their teleport\n\
+         pad technology. But then they found those\n\
+         bird-faced bruisers a month ago. These guys\n\
+         somehow remotely hijacked the portals to\n\
+         route to the Horizon star system, which is\n\
+         where the invasion force is coming from.\n\
+         Your instincts are clear - Earth can wait,\n\
+         someone's got to clean up this mess ASAP!
+E3TEXT = Before you reached Horizon, you thought\n\
+         maybe the aliens could be reasoned with.\n\
+         But then you saw humans being tortured for\n\
+         sport. Or half-eaten while still alive.\n\
+         And slaves, thousands of slaves worked\n\
+         to death building their bizarre temples.\n\
+         No, these guys are not the victims here.\n\
+         Hard to believe, but they're worse than AGM!\n\
+         Well, at least it's over now - the thing you\n\
+         just killed was some sort of psychic beacon\n\
+         homed in on AGM's colony. As the hijacked\n\
+         portals slowly lose target, you make your\n\
+         way back. Thinking of Earth the entire time.\n\
+         Your ship's non-precise warp travel might\n\
+         force a pit-stop at Saturn's Double Impact\n\
+         moon base, but it's no problem, right?\n\
+         Surely they'll welcome you with open arms!
+E4TEXT = Well, this is not good at all.\n\
+         If Double Impact was overrun, it means\n\
+         the aliens have found our solar system.\n\
+         You try to contact Earth, but to no avail.\n\
+         As you try to calm yourself by blaming\n\
+         the damaged equipment, something else\n\
+         catches your attention - an AGM digital\n\
+         wanted poster, featuring your own face!\n\
+         They're still after you, after all this?\n\
+         You smash the console in anger and leave.\n\
+         Between AGM and the aliens, you're just\n\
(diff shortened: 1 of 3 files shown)


Anonysussy:
imo i dig this one, my suggestion would be to not lean too much on the word aliens as i find them somewhat overused, try to spice it up with something like 'extraterrestrial' or some other term you could come up with, otherwise pretty good so far.

Tiramisu100:
> my suggestion would be to not lean too much on the word 'aliens' as i find them somewhat overused, try to spice it up with something like 'extraterrestrial' or some other term you could come up with

That might actually be a bit difficult due to space limits, but I'll see what I can do!

Anonysussy:
image_2024-08-08_214016665
just found out that you could apparently change up the background texture for the intermission screen, while you still at it i'd suggest change up the AQMETL texture with some other texture, as those never looked good for the intermission screen and the way they looked made it look like there's something wrong with the texture.

SenhorFlibble:
Very well-written stuff, I like it a lot! A great take on the setting and story here!

x64x2 (approved):

SenhorFlibble:
You mentioned the idea to leave Jaws of Defeat as something ambiguous without explanation. I think an obvious solution -- although it would require to change the visual design -- is to replace it with a massive teleporter or "warp gate" from where the invasion is spearheaded. So shutting it down basically stops the invasion.

One could provide some lore that AGM assisted in its construction for the purposes of research or, as you say, because they were secretly in cahoots with the aliens, thinking humanity would be otherwise doomed (no pun intended). The reverse speech sound clips could be replaced with some machinery hums and/or those sci-fi teleport/warp effects. The textures are not animated, but I don't think it would be much of a problem overall. Part of the machine could still be skull-shaped -- for example, the portal could be located in its open mouth -- but the idea is that this is no longer some kind of creature or entity that "spawns" monsters, but rather a portal to the place they are coming from.

That would also be another step away from vanilla _Doom_'s design tropes, if only in the visual/concept department.

Ferk:
I agree that it's best to avoid the word "alien", at least when used as a noun. It sort of makes them look more cookie-cutter. I don't think they should be given motivations and logic as if they were human-like... the way I interpret it, the direction is more lovecraftian.
Some alternative terms to make them sound more scary:
* Cosmic horror
* Otherwordly being
* Ancient evil
* Spawn
* Abomination
* Abyssal horror 
* Eldritch entity

And here's a write up I had from some ecyclopedic-like entry about Horizon (though I like the idea of Horizon being the star system).

> Horizon is a remote, inhospitable planet situated in the [Star System] sector. Shrouded in an aura of ancient mystery, it has long been a source of both fascination and dread. The planet's surface is a desolate, irradiated wasteland, scarred by cataclysmic events of unknown origin. Its geology is as aberrant as its history, marked by formations that defy terrestrial norms.
> 
>Beneath its barren exterior, Horizon harbors secrets of a nature so profound as to defy human comprehension. Evidence suggests the planet once supported a unique ecosystem, home to life forms of an alien and disturbing character. These entities, far removed from terrestrial biology, exhibited behaviors and morphologies indicative of an evolutionary path divergent from our own.
>
> The Horizon Disaster, a catastrophic event that occurred in [Year], marked a turning point in the planet's history. The AGM Corporation, a prominent conglomerate of the era, established a research facility on Horizon to study the planet's unique biosphere. However, the corporation's experiments inadvertently disturbed an ancient, dormant power within the planet's core.
>
> The resulting cataclysm unleashed forces of a nature so alien as to challenge the very foundations of human understanding. The facility was destroyed, and the surrounding area was irradiated. In the aftermath, horrifying new life forms emerged, bearing the hallmarks of both the planet's indigenous biology and the disturbing genetic manipulations of the AGM.
>
> Horizon’s geography is equally unsettling. Mountains rise at impossible angles, their peaks often shrouded in perpetual shadow. Rivers flow in patterns that defy gravity, their waters imbued with an eerie luminescence. The planet's overall geometry suggests a world shaped by forces beyond human comprehension, hinting at a cosmic order that is both alien and terrifying.
>
> Despite the extreme conditions, reports of anomalous readings and sporadic disappearances in the vicinity of Horizon persist. Some believe that remnants of the original alien life forms, or perhaps even mutated descendants of the creatures spawned by the Horizon Disaster, may still inhabit the planet. These unconfirmed reports have fueled speculation about the true nature of the horrors that lurk beneath the planet's desolate surface.
>
> Horizon remains a quarantined world, its secrets cloaked in a shroud of uncertainty. The planet serves as a chilling reminder of the consequences of human hubris and the unknowable nature of the cosmos.

Tiramisu100:
>just found out that you could apparently change up the background texture for the intermission screen, while you still at it i'd suggest change up the AQMETL texture with some other texture, as those never looked good for the intermission screen and the way they looked made it look like there's something wrong with the texture.

This is outside the scope of this PR so I won't be doing this, but if you want to then go ahead. I personally actually rather like the black metal texture in E2, but as a whole I'd rather all the AQTEX intermission screens were replaced with vanilla textures (though I'm biased because I'm not a fan of AQTEX in general lol).

> You mentioned the idea to leave Jaws of Defeat as something ambiguous without explanation. I think an obvious solution -- although it would require to change the visual design -- is to replace it with a massive teleporter or "warp gate" from where the invasion is spearheaded. So shutting it down basically stops the invasion.

On one hand I like the idea, on the other I feel like it takes away a bit from the "final boss" feel that a big monster like JoD naturally has. Either way JoD needs an overhaul, whatever it ends up as will be used as a basis for updating the story (and the guidelines doc too, because that can be changed over time as well).

> I think the previous text was intentionally left ambiguous. It didn't outright state that all monsters were experiments. And I believe that even the current consensus is that the creatures (at least some) are artificially engineered (the manual makes references to genetics).

In my experience I feel like the current lore heavily implies that AGM is the main antagonist (possibly with JoD pulling the strings but that's very vague), and the monsters are genetic experiments. The very first story text talks about "AGM and all their monstrous experiments" while in E1 there's only zombies, Serpentipedes and Flesh Worms, also it says that they were going to turn Karl into an "evil horror", implying the monsters are modified humans. This might have been written back when the monsters were completely different, f.eg. the snake guys instead of Serpentipedes.

> I agree with the other comment that it's best to avoid the word "alien". It sort of makes them look more cookie-cutter, like coming from a sci-fi flick. I feel Freedoom was going for a more Lovecraftian angle, in which the world "alien" basically just means "otherworldly", not a category of creatures.
> 
> If the AGM was colluding/conspiring together with Horizon's creatures, I feel it would be more appropriate for the AGM to become cultists. The contact with those ancient beings and their eldritch knowledge blew their minds, and they started worshiping the creatures as Gods (just like in Lovecraft works). The AGM people working in Horizon would become servants, they'd lose all sanity. One could say that some of the monsters we fight could be the result of AGM members themselves attempting to merge with them, as they became aware of the imperfections of their own human flesh, so their research shifted to try to achieve "Godhood" by entering in communion with the ancient "divine" creatures they found. Turning themselves into aberrations.

I see what you're going for here, but in my opinion the current Freedoom roster does not sell the whole "eldritch/Lovecraftian" theme very well. A lot of the monsters are just straight-up alien animals (like Trilobytes), some others use rather straight-forward technology and the only thing "eldritch" about them is the Necromancers and JoD. And with the way AGM was portrayed so far, I feel like "stupid greedy fucks who accidentally ruined everything" would be a better role for them than "enthralled cultists". Oh, and as for them conspiring with the aliens, that part only comes into play in Phase 2 once Earth is being invaded, they didn't consider it beforehand and even then they're only weighing their options at that point (I'll probably update the lore document to make that more clear).

Either way, I feel that introducing this plot point would require rewriting pretty much the entire thing from scratch so I won't be doing that.

Josephus-DH4050-Astartes:
> imo i dig this one, my suggestion would be to not lean too much on the word 'aliens' as i find them somewhat overused, try to spice it up with something like 'extraterrestrial' or some other term you could come up with, otherwise pretty good so far.

Xenos, perhaps? Or just go with the term Horizonites like I did in my retelling of the story?

Ferk:
> In my experience I feel like the current lore heavily implies that AGM is the main antagonist (possibly with JoD pulling the strings but that's very vague), and the monsters are genetic experiments. The very first story text talks about "AGM and all their monstrous experiments" while in E1 there's only zombies, Serpentipedes and Flesh Worms, also it says that they were going to turn Karl into an "evil horror", implying the monsters are modified humans. This might have been written back when the monsters were completely different, f.eg. the snake guys instead of Serpentipedes.

Yes, that's the one and only part where the word "experiments" appears, in contrast, the word "alien" appears 3 times. In your changes the word "experiments" appears only in the README, but the word "alien" appears 15 times.. there's no intermission where it does not show.

The previous text left it more open to interpretation, it did imply there was an alien component, and at the same time implied there was a human component. This is consistent with the look of the environment and the monsters, who mix parts of human technology and strange origin. In yours there's a lot more alien and a lot less human.

> the current Freedoom roster does not sell the whole "eldritch/Lovecraftian" theme very well. A lot of the monsters are just straight-up alien animals (like Trilobytes), some others use rather straight-forward technology and the only thing "eldritch" about them is the Necromancers and JoD.

I'd argue most of Lovecraft monsters (at least when they are clearly described) are "alien" animals, the most obvious being the deep ones who are essentially fish people.. but even the elder things are sort of winged squids that can stand straight like a plant and have a starfish on top of their head. In any case, the point was to direct the narrative towards eldritch horror, I don't think the monsters need to look the exact same as any particular lovecraftian thing. I'd argue the most lovecraftian is the worm, but I could easily imagine lovecraft trying to describe the trilobyte things like he did describe a lot of other strange looking creatures. What makes for the horror in lovecraft is the descriptions, the mistery and the narrative. It's more about the interpretation of it, imho.

With the technology I agree.. the art shows human technology, it does not look like the kind of "alien" technology you'd expect from a truly alien specie. But that's explained by the AGM's influence. Imho, in the same way Freedoom tries to match each hellish texture from Doom with a weird-looking equivalent and each techbase texture with a techy equivalent. The monsters should not be made to look more techy than the originals were, because this would introduce inconsistencies for wads. And for the most part, it has all been kept consistent in that direction (though the lord of pain / pestmeister is not great in that department, it does not use straight up tech but the armor is too metallic-looking for my taste.. it's my least favorite in the rooster, personally).

> Xenos, perhaps? Or just go with the term Horizonites like I did in my retelling of the story?

I actually do prefer that than to call them aliens. But even then, I feel that it should not be overused either, because it can break a lot of the mystery behind it if too much is shown.

Anonysussy:
> > imo i dig this one, my suggestion would be to not lean too much on the word 'aliens' as i find them somewhat overused, try to spice it up with something like 'extraterrestrial' or some other term you could come up with, otherwise pretty good so far.
> 
> Xenos, perhaps? Or just go with the term Horizonites like I did in my retelling of the story?

Xenos is actually pretty cool, imo horizonites sounds more like organism to me.

Josephus-DH4050-Astartes:
> Merge

Yeah fair, my Horizon folk are more like dragon people instead of insectoids, so that's fair.

Tiramisu100:
> Why would all this human-looking technology be created by aliens? why the wires and bolts look like human wires and bolts? somehow a human can command an "alien spaceship"? if the creatures created it, then why would it have interfaces that a human can operate or even understand? why use buttons fit for human fingers, and leds/screens that conveniently display colors in the same specific wavelength that happens to be convenient for human eyes to read? Why does it look like they use the same materials to build their electronics than the ones we have in Earth? ..I feel this makes a lot of the assumptions that many sci-fi films make when representing "aliens". They assume they necessarily need to be human-like, just with some differences that, when analyzed deeply, are very superficial, or just different enough to try and justify the plot, as a MacGuffin.

You know what, I agree with this one - I sort of left this part vague because I couldn't think of any explanation on the technology that fit within the constraints of what I'd already written, but this is indeed an inconsistency that can only be explained by AGM involvement. I also agree that the monsters could fit into an eldritch (plus AGM modifications) interpretation, even though I personally never really saw them that way.

In general I am not exactly opposed to giving the story a more eldritch theme, but as I already said I don't think I could do it justice with anything less than a full rewrite, which I'm not interested in doing. When I wrote my version, I intentionally went for a "generic sci-fi" feel (as partially explained in the first post) because I thought this was the best way to appease the multitude of different visions involved in this project but honestly I'm not so sure about that anymore. I think this is might be too big a task for one person after all, and should probably not even be attempted before the overall theme and lore is set in stone (maybe via vote or something like that?).

With that in mind I will be closing this PR, feel free to use and edit any part of it, no credit needed. As my closing remark I recommend against using the word "xenos" because it is too strongly associated with Warhammer 40k.

Donowa93:
did you really need to close this after giving your 2 cents?

PR #1403, opened 2024-07-31
Revert MAP19 music changes to 0.12 version

### LINKED ISSUE #1400: Revert MAP19 music changes to 0.12 version
At some point between 0.12 and 0.13, the music in MAP19 has been altered to use strings instead of the piano as its main instrument. Unless there is an important reason for this I believe it should be reverted, because in my opinion it makes an otherwise good song sound droning and unpleasant.

Tiramisu100:
Resolves #1400


Xindage:
let me check when and why changed, so i can progress with this pr.

Xindage:
can you please update this file as well? [CREDITS-MUSIC](https://github.com/freedoom/freedoom/blob/master/CREDITS-MUSIC)

Xindage (changes requested):
Please edit the Credits music or else it'll be closed.

PR #1356, opened 2024-03-14
Revert E1M5 music change; clear "Previously Featured in 0.12.1" section

unithecutie:
Closes #1348 by proxy.

I don't know what was going through my head when I restored Ralphis' old E1M1 track even though I vehemently wanted it replaced for being an "At Doom's Gate" remix.

Maintaining the "Previously Featured in x" section isn't really a good idea now that we've reached a stable release: it's extra clutter and will be a PITA to manage what's what.

PR #1404, opened 2024-07-31
Potential E1M5 replacement

Tiramisu100:
I made a potential replacement for E1M5.

Arguments for:
- More interesting layout and fights
- Maintains some elements from original E1M5

Arguments against:
- Potentially too complicated and difficult for E1
- Doesn't fit the current E1M5 music

As for the last point, it seems that the current E1M5 music is on its way to being reverted to 0.13 (see #1356), so if you want to test the map then I suggest you do it in the 0.13 version instead of the current dev build.

I am open to feedback, although if the proposed changes end up being too extensive then I would rather withdraw the level instead (which is fine by me too, so please don't feel pressured).


Xindage:
This map is great but there's 2 buts
DOOM01
this torch placement in vanilla is painful, the second is I think we need more bullets, I felt that there was a lot of shells but I was bullet starving.

That aside this map is great and very unusual in a good way, i'm inclined to add that in the game.

Maybe a little extra rockets would be good too.

Xindage:
As you can see by the screenshots i played in CRL basically chocolate doom whit debug features, so it's the same thing as playing original doom, that torch is placed in the middle of a lift that you must lower and them jump in to progress and that torch is infinite tall in vanilla and that is uncomfortable when you want to drop down there, since it block you.

PR #1409, opened 2024-08-12
New TNT DOSLVR textures

SenhorFlibble:
Here are DOSLVR animated texture sets that match TNT originals. The current patches are non-animated placeholders.

I based these on the old DOSHA placeholder texture -- apparently, someone started making at least DOSLVR2 patches but then used them as a basis for a placeholder instead.


mc776:
I had just noticed these ones on TNT Map20 this morning!

Xindage:
can you provide a gif or an ingame gif so we can see how it looks, the stationary images wont give us the idea of the animation.

SenhorFlibble:
> can you provide a gif or an ingame gif so we can see how it looks, the stationary images wont give us the idea of the animation.

Here you go:
ezgif-4-46e5366e41

Xindage (approved):
good good.

PR #1411, opened 2024-08-18
README: Remove outdated paragraph

fragglet:
This is left over from before the change to vanilla as the target for levels. The change should inherently ensure that scenes will never be too complicated.


SafinW (approved):
I see almost no errors.

Xindage (approved):
should be good to go.

PR #1414, opened 2024-08-20
New TNT wood crate patches

SenhorFlibble:
Here's a set of wood crate textures (CRWD and CRLWD). I thought I'd just update the old ones to add the AGM logo and such, but I actually had to recreate them anew based on Freed∞m's wood textures (RW9* and RW13_1) to properly align them with TNT's original stuff (in case there's a PWAD that includes derivatives from the original TNT set).

Here's what the new AGM logo on a small crate looks while the side still uses the original TNT texture:
crates-1

Tall wooden crate -- original TNT and new version:
crates3-1
crates3-2

I also made the light versions to match the palette of TNT originals:
crates2-1
crates2-2

The AGM logo and infinity symbol are derived from CYL1_1, the crate markings that replace the TNT logo are from Fredrik's regular crate textures, BCRATEL1 and BCRAT32.


mc776:
Those definitely look easier on the eyes than either the old FD patches or the TNT ones.

(Just to be sure, though: none of the TNT patches were yellow like that, right?)

Could you run these through pngcrush? Some of these are 300% of the previous file size.

mc776:
Force-palettized in SLADE just in case and then optimized with pngcrush: [crates.zip](https://github.com/user-attachments/files/16686012/crates.zip)

Xindage (approved):

SenhorFlibble:
> Force-palettized in SLADE just in case and then optimized with pngcrush: [crates.zip](https://github.com/user-attachments/files/16686012/crates.zip)

Thanks!

PR #1418, opened 2024-08-22
dehacked: make minigunner and SS call A_PosAttack.

### LINKED ISSUE #897: dehacked to make minigunner use DSPISTOL?
I'm probably getting way too carried away here but having these guys actually use the right sound by default would be such a relief... and a boon to new players who might spend years totally misgauging the threat they pose when they hear the same effect as the spider mastermind (which really *is* a chain-shotgun doing a shotgunguy's 3 bullets every firing frame)

mc776:
Closes #897.

Only tested in Chocolate but I really can't see any reason why the behaviour would break in Crispy or dsda-doom or anything else.

### DIFF
--- a/lumps/dehacked/dehacked.txt
+++ b/lumps/dehacked/dehacked.txt
@@ -27,22 +27,22 @@ Sprite subnumber = 32773
 Frame 689
 Sprite subnumber = 32773
 
-# The RNG code for the minigunner's / SS's attack
-# does the same thing as the zombieman's, the only
-# substantive difference being the sound played.
-# By calling the zombieman attack instead, the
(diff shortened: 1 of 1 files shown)


mc776:
Just gave this a try on demo2, works without a hitch.

Xindage:
i'll do a deep test here before if you dont mind, give me a day or two to post the results.

Xindage:
image
i tested few selected maps alone, map02, map07 and ran these d2all runs one TAS and one human, and comparated with plutonia.wad itself and seems like that these changes do not affect the playback.

Xindage (approved):

PR #1410, opened 2024-08-17
patches: change colours/fix alignment of banners (rw16_1, rw16_2, rw16_3)

Richard-Stayman:
Removes the random purple pixel on all banners, and corrects the shading in the blue and green banners where dark red was used

PR #1427, opened 2024-09-01
Levels: minor improvements: MAP01 and MAP21

Richard-Stayman:
Adds debris and slightly 'open' the door in MAP01. In MAP21 it adds more depth to the bookshelves, changes banner to red in the secret weapon room, and adds details to marble tower.
The map21 is shown with the patches from #1410 
MAP01_1
MAP21_1
MAP21_2


mc776:
The original author made a *huge* stink when I tried something similar with Map01. What we have now was the compromise.

If you want to do it this way though I'd suggest doing more to indicate that the way through is blocked because that *definitely* looks like you should be able to walk right through.

mc776:
this definitely looks better now
DOOM00

PR #1428, opened 2024-09-01
Adding back Fun is Infinite at AGM + moving E1M9 music to E1M5.

Anonysussy:
This also solves out the E1M5 At Doom's Gate remix issue.


Josephus-DH4050-Astartes:
Gonna be honest, I would rather restore "The Raging Sun" as E1M5

PR #1429, opened 2024-09-01
sprites: use Lotan's Tomb invulnerability sphere

GeorgePieVG:
Given that freedoom gets rid of pentagrams and other satanic symbols, it's worth replacing the sphere as well


KukoBerry:
Aesthetically speaking I believe the colors should be changed for a pair that fit each other better than navy blue and bubblegum pink, but this might be a good idea

mc776:
Here's Fredrik's original, for reference: https://github.com/freedoom/attic/tree/master/sprites/fredrik/pinv

SenhorFlibble:
> Here's Fredrik's original, for reference: https://github.com/freedoom/attic/tree/master/sprites/fredrik/pinv

To be honest, I did not immediately recognize that you've added shading to the original invulnerability sphere. BTW, I like how you changed the invulnerability colourmap in LT to match its appearance somewhat, although I believe that it could still use less red and more pink, purple and blue.

> Aesthetically speaking I believe the colors should be changed for a pair that fit each other better than navy blue and bubblegum pink, but this might be a good idea

Actually I'd say the colours go rather well with one another -- especially considering that the sprite is supposed to represent some kind of reality-warping artefact, not a mundane object. So it could have some colours "out of space", if y'know what I mean.

GeorgePieVG:
> Actually I'd say the colours go rather well with one another -- especially considering that the sprite is supposed to represent some kind of reality-warping artefact, not a mundane object. So it could have some colours "out of space", if y'know what I mean.

I agree, plus it's a nice homage to the historical versions of freedoom.

UltimateCraneo:
No way, Bisexualsphere

rostuhan (approved):
Bisexual sphere.

PR #1431, opened 2024-09-07
lumps: Add MUSINFO for Phase 1

GeorgePieVG:
Supersedes #1372


GeorgePieVG:
Since freedoom will get umapinfo anyway, I don't think there's any point in this PR

PR #1433, opened 2024-09-10
build: Move all source port specific stuff to a separate extras pwad

GeorgePieVG:
As suggested in discord. Advantages of this approach:
1. Pulls back the lump limit, allowing more additional textures to be included.
2. Reduces the size of the iwads, since the extras are shared between them. 
3. Allows the project to include lumps that are incompatible with vanilla but compatible with source ports. This PR also restores the widescreen statusbar removed a few versions ago, but it's now not ZDoom specific and works with any port. 

Additional rotations removed in #1380 can also be moved to this pwad.


Xindage:
The check seems to be failing, i'll test this pr manually later when possible.

fragglet:
I don't like this idea. There's a significant advantage to having Freedoom work out of the box with common source ports like ZDoom. Most people aren't going to bother loading the "extras" WAD on startup, so they're going to get menus displayed with the wrong font, etc. This isn't good.

To respond to your arguments, you argue that your change:

* Pulls back the lump limit, allowing more additional textures to be included.

We have no pressing need to add new textures to the IWADs

* Reduces the size of the iwads, since the extras are shared between them.

It isn't 1993 any more; we have terabyte-sized storage and fast internet connections. The savings here aren't significant.

* Allows the project to include lumps that are incompatible with vanilla but compatible with source ports. This PR also restores the widescreen statusbar removed a few versions ago, but it's now not ZDoom specific and works with any port.

Creating a separate "extras" PWAD allows this, but moving *all* source port lumps to the PWAD is not necessary to achieve it.

Anonysussy:
> It isn't 1993 any more; we have terabyte-sized storage and fast internet connections. The savings here aren't significant.

Kinda disagreed on this one, i sometimes had to deal with slow internet  download, also it would be nice to have a smaller file size as a whole it'll be more efficient on storage, i think the better solution would be leave out some of the less necessary extras and kept some for ZDoom or Boom-compatible ports. Compressing the maps would also probably helps out as i heard complaints about how Freedoom maps were too big at times, having some smaller maps would also helps out on the file size and the game as a whole.

fragglet:
> Kinda disagreed on this one, i sometimes had to deal with slow internet download, also it would be nice to have a smaller file size as a whole it'll be more efficient on storage

Okay, let's look at the numbers. Current version:
```
Archive:  before.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
       0  Stored        0   0% 2024-09-15 14:53 00000000  before/
28820796  Defl:N 10418303  64% 2024-09-15 14:45 a151f665  before/freedoom1.wad
28881188  Defl:N 11062668  62% 2024-09-15 14:45 d068f0db  before/freedoom2.wad
--------          -------  ---                            -------
57701984         21480971  63%                            3 files
```
With this change:
```
Archive:  after.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
       0  Stored        0   0% 2024-09-15 14:53 00000000  after/
  197252  Defl:N    76910  61% 2024-09-15 14:47 94eee838  after/fd-extras.wad
28645960  Defl:N 10354610  64% 2024-09-15 14:47 2fc204f2  after/freedoom1.wad
28706352  Defl:N 11000118  62% 2024-09-15 14:47 17884780  after/freedoom2.wad
--------          -------  ---                            -------
57549564         21431638  63%                            4 files
```
So the saving is 148 KiB on disk, 48KiB to total download size. If you're using a 56Kbps dialup connection that's an extra 6.8 seconds of download time according to [this online calculator](https://www.omnicalculator.com/other/download-time). Balance that against a worse experience for everyone if we remove these resources from the IWADs, I don't think it's worth it. I'm going to close this PR but if one of the other Freedoom devs has a different strong opinion on this matter, feel free to reopen and continue discussion.

> Compressing the maps would also probably helps out as i heard complaints about how Freedoom maps were too big at times, having some smaller maps would also helps out on the file size and the game as a whole.

I think this is a better idea that is definitely worth exploring. ZokumBSP is pretty good at compressing levels now and I can't remember if we've applied it to the Freedoom levels yet.

GeorgePieVG:
> I don't like this idea. There's a significant advantage to having Freedoom work out of the box with common source ports like ZDoom. Most people aren't going to bother loading the "extras" WAD on startup, so they're going to get menus displayed with the wrong font, etc. This isn't good.
> 

Some source ports can automatically load pwads with additional data, such as GZDoom. When ID24 is standardized, this will be a common feature (plus the Doom + Doom II rerelease also has extras.wad, which for example contains combo key graphics for Boom, idk if ports automatically load it). Other source ports that can't do this can't automatically load freedoom itself either, command line parameters will be needed for that (please correct me if I'm wrong, but all source ports I have immediately load doom2.wad (the ones that don't have a launcher)). 

> To respond to your arguments, you argue that your change:
> 
>     * Pulls back the lump limit, allowing more additional textures to be included.
> 
> 
> We have no pressing need to add new textures to the IWADs
> 
>     * Reduces the size of the iwads, since the extras are shared between them.
> 
> 
> It isn't 1993 any more; we have terabyte-sized storage and fast internet connections. The savings here aren't significant.
> 

Fair enough.

>     * Allows the project to include lumps that are incompatible with vanilla but compatible with source ports. This PR also restores the widescreen statusbar removed a few versions ago, but it's now not ZDoom specific and works with any port.
> 
> 
> Creating a separate "extras" PWAD allows this, but moving _all_ source port lumps to the PWAD is not necessary to achieve it.

Agreed, and I won't insist on moving absolutely all the lumps from the main iwads. By the way, Anonysussy suggested a pretty good idea, move only things that are too specific to some individual ports. I think you can reopen this PR and have only the widescreen statusbar in extras, it's just too good.

BethOfDeath:
I second the comment that compressing the maps with ZokumBSP would be a better approach, but I've heard of some bugs caused by using it, at least with some settings? The super-fast texture spin issue has been fixed, at least, but this would create extra work testing and re-testing maps on every change.

It may be to some advantage to begin finalizing and freezing some Phase 1 maps, compressing them with ZokumBSP, thoroughly testing them to ensure they still work in vanilla and don't have any problems, and then only touching them again to fix any bugs missed until past 1.0, to avoid introducing new bugs.

(It's been what, twenty years now? I feel it may time to start doing that (officially freezing the best assets, and polishing the rest to match the same standard) anyway; "the perfect is the enemy of the good" and all that. It's not like iD's Doom was released fully polished. But that may be a discussion for elsewhere, and I don't wish to come off as trying to be bossy.)

As for the suggested extras.wad, that would be an excellent place to put additional rotation sprites for monsters which wouldn't be used in vanilla and the widescreen HUD. Crashing some source ports what don't support widescreen assets is _not_ a good look (assuming that claim in the Discord that it happens is accurate). 

I support the approach of having:
- a _purely_ 1-to-1 lump names version freedoom*.wad for all the vanilla assets, for compatibility with PWADS which put assets to more clever uses, plus the AQUATEX textures and anything else Freedoom-specific of course -- essentially, what's already the current guidelines and has an issue and a PR (#1380 and #1381, respectively) to fix existing issues, so there's little work to be done here.
- a separate fdextras.wad for ports with extended support for de-mirrored sprites, widescreen assets, and the like. This puts control in the hands of the Freedoom project, and not dozens or more port authors who may not care to -- or just refuse due to personally liking the older versions -- update the "extras" to match any later changes on Freedoom's side.

I do not, however, support pushing this to the point of a full-on Smooth Freedoom mod at this time -- essentially, if iD didn't create de-mirrored frames for it back in the day and only cut them from release for performance/memory reasons, we should not waste artist time trying to create and polish extra frames as part of the official project, nor should it be _required_ that those extra frames be created at all. No need to burn out contributing artists, after all.

Just my two cents.

Anonysussy:
> I think this is a better idea that is definitely worth exploring. ZokumBSP is pretty good at compressing levels now and I can't remember if we've applied it to the Freedoom levels yet.

By compressing the maps I suppose to said to replace the bigger maps with some smaller maps as some maps were just way too big (Outpost Outbreak is guilty of this), though I do in favor of using ZokumBSP to help out with map compression, wadptr could probably help out as well in that regard.

GeorgePieVG:
> > I think this is a better idea that is definitely worth exploring. ZokumBSP is pretty good at compressing levels now and I can't remember if we've applied it to the Freedoom levels yet.
> 
> By compressing the maps I suppose to said to replace the bigger maps with some smaller maps as some maps were just way too big (Outpost Outbreak is guilty of this), though I do in favor of using ZokumBSP to help out with map compression, wadptr could probably help out as well in that regard.

I disagree, Freedoom has already chosen pretty hard limits for mapping (vanilla) and from simplifying the maps even more the project will clearly lose out. Especially the first episode, it was quite good. 

> (It's been what, twenty years now? I feel it may time to start doing that (officially freezing the best assets, and polishing the rest to match the same standard) anyway; "the perfect is the enemy of the good" and all that. It's not like iD's Doom was released fully polished. But that may be a discussion for elsewhere, and I don't wish to come off as trying to be bossy.)
> 

By the way, freezing resources is a very interesting topic. Imho, if it's worth freezing anything, it's only the monster/weapon designs of the most iconic of them, like trilobite. Not sprites, since they can be redrawn in a different style.

Anonysussy:
> I disagree, Freedoom has already chosen pretty hard limits for mapping (vanilla) and from simplifying the maps even more the project will clearly lose out. Especially the first episode, it was quite good.

It's a common complaints I heard alot so I thought of bringing that up, while I enjoy parts of Episode 1 there's moments where it felt like I was just meandering around and levels were way bigger than they need to be, considering Episode 1 were supposed to be an introductory episode.

GeorgePieVG:
> > I disagree, Freedoom has already chosen pretty hard limits for mapping (vanilla) and from simplifying the maps even more the project will clearly lose out. Especially the first episode, it was quite good.
> 
> It's a common complaints I heard alot so I thought of bringing that up, while I enjoy parts of Episode 1 there's moments where it felt like I was just meandering around and levels were way bigger than they need to be, considering Episode 1 were supposed to be an introductory episode.

Well, that changes the meaning of your suggestion. I originally thought that the only point was just to save space. If simplifying the maps will make them objectively better - why not?

jerk-o:
Another way to think about it might be "How much time is spent doing 'Run and Gun' tactics vs 'Switch/Key hunting' tactics?"

GeorgePieVG:
Anyone with the authority to do this, can you please reopen the PR? I'll revert all movement changes and leave just the widescreen hud for now. Moving everything else can be discussed later if needed. I'll create a separate PR for de-mirrored sprites when (and if) #1381 is merged.

mc776:
Given all the edits and the massive sprawling discussion I think if you just want to submit the widescreen status bar a new PR might be the best.

Anonysussy:
> So the saving is 148 KiB on disk, 48KiB to total download size. If you're using a 56Kbps dialup connection that's an extra 6.8 seconds of download time according to [this online calculator](https://www.omnicalculator.com/other/download-time). Balance that against a worse experience for everyone if we remove these resources from the IWADs, I don't think it's worth it. I'm going to close this PR but if one of the other Freedoom devs has a different strong opinion on this matter, feel free to reopen and continue discussion.

Just want to say that I didn't get my brains wired at the time of writing that, dehacked should logically wouldn't be that big in size, also ig the IWAD size mostly stems from how freedoom had more custom textures than regular Doom had.

PR #1434, opened 2024-09-15
CI: Bump to upload-artifact v4.

fragglet:
Version 1 is deprecated and now fails when run:

https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/


SafinW (approved):
Looks good to me!

PR #1438, opened 2024-09-19
patches: Rename exit3 patch to exitsecr + several backports from LT

GeorgePieVG:
Renamed the patch to a more obvious name. Since I didn't change the texture name, this won't affect maps that already use secret exit sign. Also I replaced the same patch and also poster1 with their variants from lotan's tomb, as the person who submitted them I'm not very satisfied with their state.


Xindage:
can we remove the poster so? because so far no map uses it already.

GeorgePieVG:
> can we remove the poster so? because so far no map uses it already.

Yeah, sure. Done

Xindage (approved):

Issue #1442, opened 2024-09-21
graphics: add widescreen variants as `*_WS` graphics.


elf-alchemist:
As seen on [NyanDoom](https://github.com/andrikpowell/nyan-doom/), a new feature some source ports are likely to start implementing as time moves on, are non-replacing widescreen graphics.

https://github.com/andrikpowell/nyan-doom/blob/master/docs/ws.md

Doomworld user [PhobosVB](https://www.doomworld.com/profile/43501-phobosvb/) has recently started working on widescreen for Freedoom as seen [here](https://www.doomworld.com/vb/post/2847702), [here](https://www.doomworld.com/vb/post/2847978) and [here](https://www.doomworld.com/vb/post/2849118).

At least updating the build script to be ready to include the named lumps as per the NyanDoom spec will be great improvement, even if temporarily being a duplicate of the normal ones.

PR #1437, opened 2024-09-19
build: Add extras pwad

GeorgePieVG:
Same as #1433, but doesn't touch anything in the main iwads and just restores the widescreen statusbar from Blastfrog.

PR #1443, opened 2024-09-21
graphics: add widescreen variants as `*_WS` graphics.

elf-alchemist:
First time contributing.
Laying down initial foundation, hence draft.


elf-alchemist:
As per the work @GeorgePieVG has done on #1437, it seems only appropriate to ditch having the wide assets built-in, and instead best to have them in the new-fangled `fd-extras.wad` file. Once their changes are upstream, I'll merge them back to mine and reconfigure the build system to include the wide graphics as replacements to the normal lumps on `fd-extras.wad`, instead of a set of new lumps on the IWADs.

GeorgePieVG:
> As per the work @GeorgePieVG has done on #1437, it seems only appropriate to ditch having the wide assets built-in, and instead best to have them in the new-fangled `fd-extras.wad` file. Once their changes are upstream, I'll merge them back to mine and reconfigure the build system to include the wide graphics as replacements to the normal lumps on `fd-extras.wad`, instead of a set of new lumps on the IWADs.

I don't disagree with your approach either, but I think the spec for non-replacing widescreen graphics should be standardized first.

elf-alchemist:
Looking at the NyanDoom repo #1442 there doesn't seem to anything still under active development as far as the non-replacing assets are concerned, I don't believe.

But at any rate, yeah I'm thinking it's more appropriate to keep the wide assets separate for the moment, unless others agree that it should be included on the IWAD itself, it's still an early moment for the non-relpacing graphics feature, and the vanilla 4000 lump limit remains dangerously close, as it stands.

elf-alchemist:
> Why did you close this one?

By mistake. I was trying to update my fork's master branch, to prevent merge conflict issues, but wasn't familiar with GitHub specifically did it, so it was an accident. I'll reopen once I get the time to work on this one again.

PR #1444, opened 2024-09-22
Reduced lines and sectors + candle fix for E2M3.

Anonysussy:
This also include a reworked version of the AGM logo in E2M2 and added wall scrolling for the portals.
image_2024-09-22_143341105


mc776:
What's the "candle fix"?

Anonysussy:
> What's the "candle fix"?

Previous edit only feature 1 candle instead of the symmetrical 4. Me just forgetting things.

PR #1446, opened 2024-09-27
sprites: Use Korp's non-gib flesh worm death animation

GeorgePieVG:
Some people prefer this animation to the current one. It also has some advantage, the flesh worm corpse actually looks like a corpse instead of a bloody mess, allowing for better navigation in already cleared areas.


mc776:
Could it be tweaked so that it lies flatter against the bottom of the image? This is not going to look right in GL (either at some angles you'd be able to see the body floating above the ground or the bottom bit of the body and the blood pool are clipping into the ground),

Saver310:
The death sound also should be changed to fit the animation
Fun fact: The current death sound is the old pre-0.7 gib sound

GeorgePieVG:
> Could it be tweaked so that it lies flatter against the bottom of the image? [This is not going to look right in GL.](https://github.com/freedoom/freedoom/pull/1024)

Thank you for reminding me of the old necromancer death animation, added it to the attic. I have almost zero drawing skills, so I'd rather not touch anything and either add edits if someone makes them or close this PR.

PR #1447, opened 2024-09-28
Musics: new e3m4 midi

Mr795:
I decided to remix dave3d28.mid because I thought it had som pretty good potential. I decided to reference dave3d21.mid as well in the beginning of the short ending solo. This midi also marks the first time I've ever used the modulation midi effect. Here's a PK3 to preview it in it's location: 
[RealmofHell.zip](https://github.com/user-attachments/files/17175794/RealmofHell.zip)


x64x2 (approved):

Xindage:
i'm not too sure about this replacement i still prefer the old one.

PR #1450, opened 2024-10-01
pass regular expressions as raw strings in Python scripts

fabiangreffrath:


mc776:
Non-programmer question: what difference does this make? Is it a performance thing, or a security thing, or futureproofing against something?

Xindage:
i have the same question too, some explaining in what it does would be great.

fabiangreffrath:
These strings currently trigger a warning that will become an error in a future python version.

mc776 (approved):

PR #1451, opened 2024-10-02
CI: use deutex and ruby-asciidoctor-pdf from Ubuntu repositories

fabiangreffrath:
Currently, ubuntu-latest is at Ubuntu 24.04.
Deutex 5.x and ruby-asciidoctor-pdf are in Ubuntu since at least 20.04.


fabiangreffrath:
Never mind, this takes 20 seconds longer than before, and deutex pulls in the freedoom Ubuntu package as a dependency, which might lead to confusion...

PR #1456, opened 2024-10-08
patches, textures: Minor aquatex optimizations

GeorgePieVG:


mc776:
This converts a lot of single-patch textures into multi-patch textures. Anything that uses the affected textures on a 2-sided linedef will now be medusa'd.

While this is less likely to break a shitton of stuff than my earlier cockup with the sprite rotations (on that note can we *please* merge #1381 already) given how long aquatex has been around there probably is something out there that will be broken by this.

GeorgePieVG:
Yeah, definitely worth more testing, which is why I marked the PR as a draft.

Xindage:
i'll be pushing it, for a long time it is open and if it broke something better early than late, if you want to add more stuff fell free to open a new pr to update it.

mc776:
wtf

Adding "actual concern for compatibility" to the Lotan's Tomb webpage differentiating selling points then...

PR #1338, opened 2024-02-14
New TEKGRENs by @KorpKat

unithecutie:
These are edited from #1166; and RW39_3/TEKGRENW is intended for #1309.
I really don't want to touch `textures.cfg` again...

PR #1472, opened 2025-02-02
Restored Korp's TEKGRENs from #1338

Anonysussy:
I find these textures to be cleaner for me and I like it compared to the dirtier compressed current one, this didn't include the newly added textures as I was too lazy about it.

Screenshots for comparison:
Current:
DOOM20
DOOM26
Korp's:
DOOM00
DOOM01


Calinou (approved):
I agree these look better overall.

PR #1482, opened 2025-02-08
lumps: add GLDEFS.

### LINKED ISSUE #1479: Add `GLDEFS` lump for dynamic lighting in OpenGL renderred source ports
Some sprites in Freedoom differ in color from their Doom counterparts, leading to a mismatch in the default Doom `GLDEFS` definitions in such ports.

From @Xindage:

```
the sprites that i know of that glow wrong are bonus pickups (both), red armor (blue), Bfg projectiles
```

Image

mc776:
Resolves #1479.

This is copied verbatim from GZDoom and has not been tested on any other port.


elf-alchemist:
Looks good; only a nitpick, all the `attenuate` entries are using tabbed indentation instead of spaced indentation in the rest of the file. The `BLURSPHERE1` definition also has it's `subtractive` and `interval` entries with such mixed indentation.

mc776:
I think the best policy is to just leave it until there's opportunity for a vote which one to go with... I don't even want to share my own preferences on here.

Xindage:
for now i'll be pushing, it can be changed later if it's really an issue.

PR #1486, opened 2025-02-09
Sound: Punching

HorrorMovieRei:
Back when I and a couple other submited the Octaminator, I included a new punching sound for it in the package. I wanted to make it's punch sound stronger than the player's punch (like it is in Doom with the revenant) so I gave it some more bass and lowered the pitch a tiny bit, but in the end it just sounds kinda muddy.

To solve this I went the opposite direction this time. I gave the current more over-the-top punching sound we have as is (unedited) to the Octaminator, and for the player I edited it to be a bit shorter and more subdued, but still satisfying.


mc776:
I just gave this a try and I think I like the current sounds better - I can't hear the difference in the octaminator punch muddiness at all and the louder current player punch gives better feedback for whether your attack connected (which is already a bit difficult given doom's fist doesn't actually launch the hitscan attack when the sprite is showing full extension but the frame before).

mc776:
Is this about the revenant punch or the player punch?

HorrorMovieRei:
> Is this about the revenant punch or the player punch?

Both based on the same sound, so it doesn't matter. Mhhh, in light of that information however I feel this PR could be closed. I'd rather the sounds be replaced with something wholly new now just to be on the safe side.

fragglet:
Closing as I have merged #1509 instead.

PR #1488, opened 2025-02-11
Add Cescade's Keen and plasma sprites.

Anonysussy:
Adding it here as a PR bcs i really like the Keen's sprite Cescade done, also plasma while still on it.

Screenshots for comparison:
Current:
DOOM0014
Cescade's sprites:
DOOM0015


UltimateCraneo:
I think it's neat, I like the red to green meter on the side, like a heat thing, which makes sense for the cooldown sprites

HorrorMovieRei:
> Keen looks like a big improvement. I'd like to see more opinions of the plasma rifle before merging (just because weapons are so prominent)
> 
> (It's usually best to only do one thing per PR)

Definite improvement. Keeps the same basic design but makes it more cohesive. I like that the keypad has been replaced with a dial that presumedly indicates weapon heat. The placement of the keypad above the gun like that never made much sense, even as a sci-fi weapon. The tubes running along the side are a nice touch too, you could probably imagine they're running energy from one part of the gun to another, or helping disperse heat, or doing insert sci-fi jargon here

PR #1490, opened 2025-02-17
patches: new DOSWxxx replacements

quarry24:
the current DOSWxxx patches are very obviously placeholders so here are new ones that use some freedoom resources, the skull is taken from the attic (hopefully i did all this right)


Talon1024 (commented):
Could you make the blue ones brighter?

quarry24:
before i post the commit, how does this look?
DOSWX1C

Talon1024:
Much better! :+1:

mc776:
Any objections to me adding this to Lotan's Tomb before/whether or not this gets added to FD?

quarry24:
> Any objections to me adding this to Lotan's Tomb before/whether or not this gets added to FD?

for my input, nope, you're very much allowed to use this for whatever project

PR #1491, opened 2025-02-23
Improved readability for M_DOOM and reverts the infinity color.

Anonysussy:
Went back to reedit the M_DOOM, i realized it might not be the best idea to color the infinity symbol the more brownish color as it blends in with the background (i initially did it so it doesn't look overly bright like my prior attempts pre-PR, i also went back to add a bit of tiny white outline so that the silver color scheme didn't blend in with the black outline.
Screenshot 2025-02-23 195022
Screenshot 2025-02-23 200155


Anonysussy:
> Is this an improvement? The colors look rather washed out. It's nice to have a title logo that stands out.

I think in action it look good enough, though more opinions would certainly be welcome.

SeventhSentinel:
This is both more stylish and more legible, +1 from me.

Calinou (approved):
This is a lot more readable in source ports that don't apply a background color to the pause menu (including most vanilla-like source ports).

PR #1492, opened 2025-03-03
New E1M1 song (Let's add Lola's new submission to e2m1 or somewhere it could better fit)

Josephus-DH4050-Astartes:
I love Lola's submission, but it does indeed hit a bit too hard for e1m1, however, it should, in my opinion, be added to phase 1 or 2 in E2M1, or etc. It is a very good song, but as I said before, it does hit a wee bit too hard and sudden for E1M1.


Anonysussy:
> I love Lola's submission, but it does indeed hit a bit too hard for e1m1, however, it should, in my opinion, be added to phase 1 or 2 in E2M1, or etc. It is a very good song, but as I said before, it does hit a wee bit too hard and sudden for E1M1.

Imo, I find the song very first level-ish, and it would replaces the current E2M1 track which is amazing on its own. Imo I prefer something like Korp's Offbrand Starlight being used in E1M1 instead and E2M1 being kept as it is.

Josephus-DH4050-Astartes:
> > I love Lola's submission, but it does indeed hit a bit too hard for e1m1, however, it should, in my opinion, be added to phase 1 or 2 in E2M1, or etc. It is a very good song, but as I said before, it does hit a wee bit too hard and sudden for E1M1.
> 
> Imo, I find the song very first level-ish, and it would replaces the current E2M1 track which is amazing on its own. Imo I prefer something like Korp's Offbrand Starlight being used in E1M1 instead and E2M1 being kept as it is.

Fair point but tbh i feel that e2m1 is pretty weak (no offense to Korp, the man does great work) and I feel this style of song (Fast paced, riff-heavy rock) is better for e1m1.

Anonysussy:
A bit of embarrassing thing to admit, but i haven't really listened to the song until now because i thought this was posted as an 'Issue' instead of a 'Pull Request', got to listen to the song now and i really like it.

Josephus-DH4050-Astartes:
> Quite ashamed to admit, but i haven't really listened to the song until now because i thought this was posted as an 'Issue' instead of a 'Pull Request', got to listen to the song now and i really like it. I also find it to be weirdly interchangeable, it could fit in with later levels on Phase 1/2 or DM and could even fit in as a boss theme, not that i complained about it.

Yeah! An accidential success on the interchangeable part but I tried going the classic e1m1 style route, but just wrap it in new era Freedoom lol

PR #1496, opened 2025-03-24
Updated Charli's E1M1 map.

Anonysussy:
Got permission from Charli to do a PR for the updated map.


fabiangreffrath:
The map is not bad, but so is the current E1M1. I don't see why it should get replaced?

Anonysussy:
> The map is not bad, but so is the current E1M1. I don't see why it should get replaced?

From what I read, it has something to do with the current one being overly detailed and too much of an impression for a first level, the current one I also find really short if you're going for the exit or decently long if you're going 100%, I find Charli's version to have better balance on the length and progression, it eases things up to prepare for later levels.

jerk-o:
The current e1m1 uses the blue key, but this one doesn't use any keys, which is nice for new players on a "First Level" map.  This one is also easier for wandering straight to the exit.  However, I do think having an "Exit" sign next to the door at (860,-276) would help new players.

PR #1498, opened 2025-03-29
new jump sound

superdave938:
a new jump sound that is a little more subtle than the current one


fragglet:
I like the idea of a more subtle jump sound but I'm not sure I like this one. It sounds a bit artificial and it cuts off very suddenly.

PR #1499, opened 2025-04-18
levels: change E3M2 big lift line triggers.

### LINKED ISSUE #1401: E3M3's confusing lift
At some point in E3M3, you have to raise a lift (tag 26) to get to the yellow key. However, the lift's texturing does not suggest it to be interactable at all. Anecdotally, when I first played the map this caused me to get stuck to the point where I thought I was soft-locked. In my opinion the lift's front texture should be changed to a switch texture in order to prevent potential confusion.

mc776:
Does not address any issue anyone else has posted but was one of my pet peeves about this map.

The visible switch that lowers the yellow key now lowers the lift instead, the line on the lift itself is no longer usable, and the yellow key is lowered by a second walk line next to the one that releases the painlord.

This PR should not change the flow of the map in any way, except to remove one compulsory press of the use key and to change the direction you're facing when that lift starts moving.


mc776:
I think I'll just close this for now to prevent accidental closure of the issue I mistakenly tagged.

mc776:
(fwiw I still think this is an improvement on the feel of the map, but it's not an objective improvement unless other people are actually confused by this lift as well)

PR #1503, opened 2025-05-21
Simplify the help screen

### LINKED ISSUE #1489: Skull on help screen obscures keys
This is the skull with the flashing eyes on the bottom right slightly obscuring the keys on the help screen.

Happens on both Phase 1 & 2.

fragglet:
This fixes #1489, where the keys were being obscured by the menu skull that is drawn on the help screen by vanilla Doom. Rather than just shuffle around the contents of the help screen, I opted to simplify it by removing a lot of the clutter.

The old help screen was created before the Freedoom manual was added, so there's no need for us to include every in-game item. Instead, we can delegate to the manual by including a link to it, and cut things back to just a small selection of example items.

### DIFF
--- a/graphics/text/Makefile
+++ b/graphics/text/Makefile
@@ -70,59 +70,23 @@ help.png: text_strings helpbg.png
 	./smtextgen help.png 320x200                 \
 	   -background "helpbg.png"                  \
 	   150,2   "file:helpttl.png"                \
-	   10,20   "Weapons"                         \
-	   20,33   "file:../../sprites/csawa0.png"   \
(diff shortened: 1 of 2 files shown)


elf-alchemist:
Looks much better already.

woof0013

Calinou:
Looks great! I'd add a comma to `to open them, you will have to find the right colored key`

mc776:
Sorry for the belated suggestion, but could we consider getting rid of the word "colored" and rearranging the key sprites so the yellow/vertical, red/horizontal and blue/X keys are visually paired with each other?

PR #1504, opened 2025-05-26
GFX:flat/mflr8_4 replacement

lavenderdotpet:
### Description of Changes
---
replacement for the "flat/mflr8_4" texture 
sourced from a nasa public domain mars image 

### Visual Sample
---
image

PR #1505, opened 2025-05-28
flats: add old MFLR8_4 as MFLR8_5.

mc776:
See comments on #1504.


mc776:
One surprisingly common use case: Freedoom maps that assume the old MFLR8_4 and use it as a ceiling, which looks bad when it's visibly supposed to be sand that's being moved by the elements.

Here's both flats together on Map25 with the ceiling swapped out:
DOOM00

lavenderdotpet:
> @mc776 Got any screenshots of the new texture in action for comparison's sake?

Anonysussy:
> > @mc776 Got any screenshots of the new texture in action for comparison's sake?
> 
> 

I mean in-game where I could determine whether one or the other works in 3D space better, I did playtest it on GZ in E2M3 and it seems fine for me.

mc776:
Both flats are present in that screenshot.

PR #1509, opened 2025-06-10
sounds: new punch sounds.

mc776:
I'm still stuck with grossly subpar recording gear right now, but this is urgent (or rather *long* overdue) as per discussion in #1486.

DSPUNCH is just me punching myself in the arm run through a *bunch* of stuff in Saucedacity until it started to sound like a punch ingame instead of a weird tinny click.

DSSKEPCH is that same sound remixed with a bit of DSSWTCHN for a vague impression of a metallic echo.


mc776:
11025 Hz versions: [punch11025.zip](https://github.com/user-attachments/files/20665564/punch11025.zip)

lavenderdotpet:
I def think theses are good fits for replacing the punch sound

PR #1511, opened 2025-06-11
Deprecate MD5 hash for JSON metadata file

fragglet:
MD5 has been deprecated for years and it's time we switch to something newer. Add BLAKE2b and SHA3 digests, with a TODO to remove the MD5 digest in a future version.

This also fixes a warning when running the script (and potentially a bug?), and deduplicates some duplicated code.


fragglet:
@selliott512 can you take a look and review if you have time? Thanks

PR #1515, opened 2025-06-14
Very minor pickup edit.

Anonysussy:
Slightly altered the sprites for the pickup.
Screenshot 2025-06-14 173134
Screenshot 2025-06-14 171119


Anonysussy:
> Can you provide a screenshot to compare with the current sprites? It's barely visible from GitHub's preview 😅 I can just about make out some added shading and texture, and if it's just that, LGTM.

Yeah it just that, hence why I called it very minor.
spritecomlarison

mc776:
The shells don't look round anymore

mc776:
I mean more that the contrast is cranked up to such extremes with this change that up close I'm no longer parsing that as a gradient but just two unrelated shapes...

Cescade:
Upper areas of boxes could be brighter or retained from the original

The shells look alright but the contrast of the green could be toned down cuz of the material not being that metallic

PR #1516, opened 2025-06-14
sprites: simplify FCAN.

mc776:
(also update the flame, apparently)

When the current fire can sprite was first introduced it came with a small, straight blue flame, in keeping with the appearance of a pressurized canister full of flammable gas: https://github.com/freedoom/attic/blob/master/sprites/raymoohawk/old-fcan/fcana0.png

It made perfect sense internally, but broke a lot of maps that used this actor (including FD's own Map16) as a generic rough yellow flame effect.

This was amended so that the flame's shape looked more like the slanted id one, but stayed blue: https://github.com/freedoom/attic/blob/master/sprites/raymoohawk/newer-fcan/fcana0.png

Eventually it was decided we could differentiate from id enough by doing the exact opposite - keeping the colour but changing the shape, resulting in the current yellow flames going up. Unfortunately by this point it was clear that the pressurized canister had long stopped making sense, but I'd already done the flickering lighting animation effects on it and didn't feel like making any changes.

This bugged me for years until today I decided to splice together something from the nukage barrel's bands and here we are.


elf-alchemist:
I like this one quite a bit :)

I find some of the FD sprites to be rather over-detailed IMHO (especially if compared to their Doom counterparts), so this new design feels much more sensible to me.

uhbooh:
Nice job! Those sprites look almost better than the original.

PR #1517, opened 2025-06-15
Added Illost's Wolf portrait + new frame.

Anonysussy:
I like these and i think these were nice replacements of the current one.


Anonysussy:
> These are really good! Definite improvement over the existing set, but I believe Illostrious is soon to work on the other portraits, as well? Might be worth waiting a moment, and changing the PR to a Draft PR in the mean time.

That could work, though imo i find it less essential as i find the other portraits fine enough and do a good job at referencing older materials.

mc776:
The porportions and perspective look better but could it be less shiny? It looks very off with that much shine on people's foreheads and hats and fabric like that.

superdave938:
I yeah I helped out on this a little

HorrorMovieRei:
Big improvement. I like that it's clearer now how these are supposedly AGM higher ups. CEO maybe? General? God emperor?

mc776:
> The porportions and perspective look better but could it be less shiny? It looks very off with that much shine on people's foreheads and hats and fabric like that.

Looking better now!

UltimateCraneo:
> Big improvement. I like that it's clearer now how these are supposedly AGM higher ups. CEO maybe? General? God emperor?

I like to think the old guy is the current CEO and the hat guy is the founder of the company like decades ago.

Anonysussy:
Gonna draft this for a bit to solve the contrast stuff a bit.

Donowa93:
wait actually, i think it might be a good idea to remake the prisoner graphics (the one with the red sky and the one with the breaking shackles) using some of illost's other fanarts

KukoBerry:
I disagree with Donowa sadly, I think the prisoner paintings are both quite good in the sense that they're callbacks to previous versions of Freedoom and are already visually quite appealing. They can stay as they are.
Though these ones are a great improvement!

HorrorMovieRei:
I second the above, I think the callback to the old titlepic is very cute.

Anonysussy:
> > I second the above, I think the callback to the old titlepic is very cute.
> 
> well, i mean illost did remake the brown-armored prisoner from the old titlepic

Imo, it wouldn't hit as much as the original portrait.

KukoBerry:
Yea and the point is that it IS a collage of those old assets
...but that's neither here nor there, this PR is abt the AGM Boss Portraits not the prisoner ones, right?

PR #1519, opened 2025-06-18
Restore "Derelict" by Ivan Stanton for E1M5

uhbooh:
This song was previously featured in E1M5, but was replaced when the old E1M9 was moved to that slot. Now E1M5 is a completely new level that uses the old E1M1 song by Ralphis. I think this song fits the new level much better than the current song, which is also just a rip-off of vanilla Doom's E1M1.


uhbooh:
Are you thinking of Armory? That's the previous E1M9 which was moved to E1M5 and then replaced with Cloudspire Keep. I don't think the current level is a "shooting gallery" which is why I ask.

northivanastan:
"Derelict" was made for an old E1M5 called "Main Control", which was so sprawling it could take up to 30 minutes to beat, and often felt empty. Cloudspire Keep is a lot more of a shooting gallery than that, at least.

Between The Raging Sun, Derelict, and the old E1M1 music, I think Derelict fits best (though I'm biased since I wrote that song). I'd much rather someone make a new MIDI though.

PR #1523, opened 2025-07-03
Slight rewording and fixed typo.

Anonysussy:


mc776:
This is more words for less effect.

And wasn't the recent rewrite specifically prompted in part by someone calling FD a "doom clone"?

Anonysussy:
> And wasn't the recent rewrite specifically prompted in part by someone calling FD a "doom clone"?

Personally I don't feel like it's worth much trouble with, considering the person in mind doesn't even provide credible source to back up their claims and goes silent once people start calling them out, it's written in a slanderous manner, intentionally dismissing the project's intent and goal, while the 'look and sounds cheap' could be a proper criticism on its own right, the lack of much arguments towards it and extremely subjective opinion made everything felt like a slander at the project more than anything else.

mc776:
The new version is better than the original PR, but the "efforts is" is still there and now the word "clone" is used twice.

What even is wrong with the original wording anyway? Can you explain what the thought process is behind these proposed changes? Because frankly every single thing I've ever been taught about good writing would make me think a good edit would be in the other direction.

mc776:
For what it's worth, here's my suggestion:
```
goal is just to make a "free Doom clone". This is not the case!
While compatibility with Doom is a goal, we are building our own game.
We want works that are original and distinct, rather than trying to
recreate anything from Doom.
```

- The original placement of "just" felt awkward to me. (Purely subjective)
- Might as well face the "Doom clone" shit upfront and call it out by name.
- Changing the last sentence to "I/we" language to avoid telling the reader what to do (who may or may not even have any intention of submitting anything) but making it clear that there are standards.
- The "effort to ensure" stuff was kinda clunky anyway.
- It already goes without saying that *literally* copying is bad, so changing that to "recreate".
- Moving the "While" to a new line so it's not stuck at the end of the line there like a runt.

PR #1526, opened 2025-07-15
levels: fix DM24 and DM25 vanilla issues.

mc776:
Just a couple things I accidentally found tonight.

DM24: In the northwest, if you hit the light grey pillars (the ones with the corners filed off) at just the right angle, the projectiles would disappear. Turns out a huge part of that upper trim's sky-facing linedefs were [facing the wrong way](https://github.com/freedoom/freedoom/issues/996).
dm24-disappearingproj

DM25: The use of BIGDOOR6 for those steps was causing tutti-frutti. This is exactly the sort of thing BIGDOR6A was made for - all 21 occurrences of this texture replaced accordingly.
dm25-tutti
dm25-notutti


Xindage:
where is the map25 changes on this pr?

PR #1527, opened 2025-07-15
patches: redo BIGBRIKs.

mc776:
These were my first submissions to Freedoom, an urgent rush job to replace a patch that appeared to have been the id patch with some bits drawn over it. #600 I've never stopped hating looking at them.

This new patch is based on Espi's old MFLR8_1 https://github.com/freedoom/attic/blob/master/flats/espi/mflr8_1.png.

Before
DOOM01

After
DOOM00


Anonysussy:
Couple of edits i did to add a bit of lighting, haven't tested these in vanilla, i don't know / forgot the level that uses these textures.
rw23_3
rw23_4
rw38_4

Donowa93:
i think korp's fit with the more intricate shading of a lot of other freedoom textures (especially the ones that cescade's been making recently)

PR #1528, opened 2025-07-17
Replace Keen/Fetus-in-a-tube sounds with glass breaking

Donowa93:
sounds sourced from https://github.com/lavenderdotpet/CC0-Public-Domain-Sounds , the specific sample sources are bfh1_glass_breaking_02, bfh1_glass_breaking_01, bfh1_glass_falling_02, mud_02, creature_slime_01, glass_05, and bfh1_glass_falling_03

i made these because the synth sounds just DID NOT fit the tube sprites, and that felt especially weird on map11 (where the keens are critical to obtaining a few major secrets)
these might be a bit weird for mods that use the vanilla keen sounds as-is, but it makes more sense for freedoom as an original game

Old sounds:

https://github.com/user-attachments/assets/da1a5b06-d99c-4fd8-bdff-dbd231bfe3fc

New sounds:

https://github.com/user-attachments/assets/ba99fcee-dc94-4aa6-9094-179b18a62025


HorrorMovieRei:
cc0 is "no restrictions" in simpler terms, so it can be used for anything. Also, these new sounds are pretty satisfying. Even if it might sound weird in mods that use the keen sounds as is, I think a glass breaking sound is overall a lot more valuable as a resource for people to use should they make a project that uses Freedoom as a base.

KukoBerry:
Is there still an issue with the source of these sound effects? :p or something else

Donowa93:
nope, these are sourced from @lavenderdotpet 's cc0 sound repo (cc0 is basically international public domain)

PR #1529, opened 2025-07-18
Altered fist offsets.

Anonysussy:


elf-alchemist:
Does this still look fine on 4:3? It looks like it will get clipped on the side.

PR #1534, opened 2025-07-30
New spiderdemon

Cescade:
Spiderdemon replacement called "Dreadnought" I made
Dreadnought

Ingame
doom19


Saver310:
Yes...

elf-alchemist:
Yes, finally! We've been waiting for this for so long. 🥳 These new ones are such an improvement!

elf-alchemist:
True, true. I wasn't expecting this to immediately trigger 1.0 to come out. But at least that goal is now attainable.

Anonysussy:
I'd suggest also checking out the dehacked and lumps stuff, as they could still probably deadnamed the Dreadnought as Large Technospider instead.

elf-alchemist:
> I'd suggest also checking out the dehacked and lumps stuff, as they could still probably deadnamed the Dreadnought as Large Technospider instead.

Need some feedback on the name choice for the Spanish version of the manual
https://github.com/freedoom/freedoom/pull/1535/files

uhbooh:
This should definitely be merged; it looks very professional.

PR #1536, opened 2025-08-04
Musics: New MAP16 Music

Mr795:
I decided to make some new music for map16, this time it's an original song, about just a bit over 3 minutes in length, i decided to try going with a more "electronic rock"-esque style with this one.


mc776:
I've always been partial to "Reflection" because of the way its jazzy intro fits the nighttime urban skyline motif of this map so nicely, but honestly actually *playing* the map is way more enjoyable with this track.

Mr795:
I decided on the latter choice, slightly adjusted the church organ volume, added a little drum part for the beginning and added sawtooth to the beginning as well, made the delay for the square a tiny bit louder as well.

perihelioon:
I like this.

PR #1538, opened 2025-08-11
textures: create WALLFONT and WALLFON2.

mc776:
These are dedicated textures for doing stuff like that screen message hack in E2M4 - one one-sided, the other transparent for midair or as an easier hack to do multiple rows of text.

Made everything 8x8 for ease of use.


mc776:
Alternate scanline version:
DOOM01
wallfont

PR #1543, opened 2025-09-07
Replace rock, slime, and dobwire

Cescade:
Screenshots
doom18
doom23


elf-alchemist:
These new textures are really good, I really like the polished look on the new rocks, but those chicken wire(?) grid fences looks so much better than their previous entries.

PR #1550, opened 2025-10-31
Readded some old musics.

Anonysussy:
Found some tracks from 0.11 that i thought would be good enough to bring back, with them occupying different slot from 0.11, i don't recall FreeDM equivalents of these tracks.


unithecutie:
> Credits/names of tracks?

MAP10 = untitled (map01.mid) by Ralphis
MAP27 = untitled.mid by Varis Alpha

--

Don't entirely agree with this, current tracks work perfectly for their levels.

perihelioon:
I'm a bit unsure about this. These tracks don't exactly "add" anything to these maps.

PR #1554, opened 2025-11-05
levels: replace all AASHITTY in E3M7 and E3M4.

mc776:
Helion will render AASHITTY as all black even in Phase 1.


mc776:
Yeah I've been wondering if it's even worth reporting to the Helion devs or if we can all just start agreeing never to use the AA textures...

nstlaurent:
[Fixed](https://github.com/Helion-Engine/Helion/pull/1393) in Helion for the next release.

PR #1569, opened 2026-01-06
gldefs.lmp: (Q-)Zandronum does not support attenuate

pmjdebruijn:
(Q-)Zandronum fails to load the wad if attenuate is present in gldefs.lmp


Xindage:
already fixed in #1581

Issue #1394, opened 2024-07-22
AQMETL textures produce distortions at distance in vanilla software renderer


SenhorFlibble:
I'm not sure if this has been brought up before, but I was playing Phase 1 recently in MBF 2.04 (in DOSBox), and there is some very noticeable gradient-like wavy distortion that appears on walls with AQMETL textures in C1M1 when viewed from a certain distance:
![mbf_007](https://github.com/user-attachments/assets/a5843952-536f-4de2-934a-f91a3827ab3b)
![mbf_008](https://github.com/user-attachments/assets/3e77ae51-300f-4625-abfb-6e6ca9fc1df1)
The same effect can be also observed in FastDoom, Chocolate Doom, Woof, but somehow not in GZDoom.
![choco](https://github.com/user-attachments/assets/007254bb-a1a2-48b7-9db8-e812841e43f6)
My guess is that this is somehow related to how COLORMAP works with the particular combination of pixels that form the metallic pattern on these textures, although I loaded ColdPal and had effectively the same visual artifacts as with the default palette/colormap.

Do you think this can be fixed by editing the textures? I'm not sure if any other textures outside the AquaTex set act like this. Looking at AQMETL07, there's a very consistent pattern of diagonal lines forming on the metallic part of the texture, which I suspect might have something to do with the artifacts observed.

Anonysussy:
> The same effect can be also observed in FastDoom, Chocolate Doom, Woof, but somehow not in GZDoom.

This is quite stupid frankly, GZDoom runs on higher resolutions with some of the texture filtering options offers mipmapping which mask the artifacts from further away by rendering them on lower resolutions and all kinds of luxury that Vanilla and Chocolate Doom simply didn't have.

SenhorFlibble:
> > The same effect can be also observed in FastDoom, Chocolate Doom, Woof, but somehow not in GZDoom.
> 
> This quote is rather stupid frankly, GZDoom runs on higher resolutions with some of the texture filtering options even offers mipmapping which helps to mask the artifacts from further away by rendering them on lower resolutions and all kinds of luxury that Vanilla and Chocolate Doom simply didn't have.

[...]

PR #1570, opened 2026-01-08
Smoother and less moiréy textures. 

Anonysussy:
Giving my shot at somewhat partially mitigating the issue at #1394 where textures causes moiré patterns and shimmering in vanilla low-res render. 
Screenshot 2026-01-08 224537


Calinou (commented):
I think this goes in the right direction overall (especially for the stripes repeating less often), but I'm not sure about stone/grunge patterns losing their contrast.

Blastfrog:
I've got mixed feelings about this. Moire is just a part of how textures look in a low screen resolution and without mipmapping. While I do welcome some polish, I feel like most of this results in a very blurry and pillowshaded look. Freedoom's art has long been plagued with pillowshading, and it's something I'd prefer to see it move away from.

Definitely good to avoid extremely harsh contrast concentrated in small spaces, but sometimes you've got to just let the moire be in favor of more clarity. Something in between this PR and the current textures is ideal IMO (closer to current).

Anonysussy:
Hasn't given much time to work on this unfortunately, though I did try to remedy a bit with more recent PR, I guess I'm open to submission in case someone wants to help improve it further.

mc776:
> Something in between this PR and the current textures is ideal IMO (closer to current).

I'd left my own suggestion on the issue but it was objected to because *even the thick diagonal warning stripes* were considered too problematic...

Issue #1572, opened 2026-02-15
RWDMON7 and RWDMON9 appear to be swapped


TheGoomba98:
The new boss brain textures in pull request #1164 look good, but it looks like the two middle-side textures were swapped around. RWDMON7 in the vanilla Doom 2 connects to RWDMON2 on the latter texture's right side (the viewer's right), and has a monitor on the upper-left (closer to the skull itself when put together), but Freedoom's RWDMON7 looks like it was intended to be RWDMON9, since it looks like it's meant to visually connect to the left side of RWDMON2, and has a monitor on the upper-right; and vice versa. The monitor placements are how I noticed this, since the creator of the new textures intended to put the monitors in the same spots as the vanilla textures for PWAD compatibility, but the mix-up defeats that purpose - and also causes a slight graphic incongruity when it comes to the full, assembled boss texture for that matter.

I might put together a PR for this when I have the time, since it's an easy fix - I just wanted to get this jotted down first.

PR #1573, opened 2026-02-16
Corrected Jaws of Defeat middle-side textures

TheGoomba98:
As I noticed in #1572, the patches RWDMON7 and RWDMON9 looked like they were swapped from their intended positions. This should improve PWAD compatibility when it comes to the monitor parts of the textures, as well as make the whole boss face look nicer.


Talon1024:
MAP30 (Jaws of Defeat) will need to be fixed as well. Currently, the middle row uses ZZZFACE6 4 8, from left to right, but it should be ZZZFACE8 4 6.

PR #1576, opened 2026-03-24
sprites: fit spheres to Doom pixel aspect ratio.

mc776:
before
before

after
after


Calinou (approved):
Makes sense to me.

Note that most source ports render the game world in `1:1.2` aspect ratio, even if they don't aspect-correct the HUD. So this change should appear fine on just about any source port out there.

Talon1024:
I still have the source files for my stealth sphere, so here's a "widened" version for you:

PINSA0
PINSB0
PINSC0
PINSD0

I'd prefer you use these instead of modifying the existing sprites, since there's no distortion caused by editing the existing sprites.

Also, it's worth noting that these sprites are not in the Doom palette yet.

fawtytoo:
This makes no sense to me. In vanilla Doom, these sprites are round before any aspect correction. Therefore the sprites in Freedoom should follow suit.
This forces players who choose other aspect ratios to view distorted sprites just to suit the players who view in 4:3 aspect.

Anonysussy:
> just to suit the players who view in 4:3 aspect.

Doom was intended for 4:3 aspect ratio, a lot of the assets, sprites and the title screen itself were designed with it in mind.

Many source ports default the aspect ratio to said 4:3, making the wider aspect obsolete.

Also other Doom games show them as full circle instead of oval.

Anonysussy:
> Therefore the sprites in Freedoom should follow suit.

Also when does this rule were even stated?

mc776:
d1soulsphere

fawtytoo:
> > Therefore the sprites in Freedoom should follow suit.
> 
> Also when does this rule were even stated?

This PR suggests a rule. Where is that stated?

> > just to suit the players who view in 4:3 aspect.
> 
> Doom was intended for 4:3 aspect ratio, a lot of the assets, sprites and the title screen itself were designed with it in mind.

Doom wasn't _intended_ to be that way. **All** of the item sprite assets in Doom are round. In fact, **all** of the monster projectile sprite assets are round. So there is no reason why these particular assets in Freedoom should be designed differently than it was in Doom. Unless you're suggesting that these sprites were _intended_ to look oval?
Doom was created on a computer that, not only had a much bigger resolution (1120x832), it didn't require "aspect correction". In the day, the monitor was used to "correct the aspect", not the computer's hardware or software. And that was down to personal preference.

Anonysussy:
> This PR suggests a rule. Where is that stated?

I never said this PR suggest a rule, i only said that Freedoom never has any obligation in this specific part, what are you trying to say?

> And that was down to personal preference.

I know and i don't blame anyone who didn't play on 4:3, it's just that many source ports opted for it, and 4:3 is the intended way (check the wiki for proof), otherwise id wouldn't be bothered with monster sprites and title screen.

Also further proof:
rhcZdi2

fawtytoo:
> > This PR suggests a rule. Where is that stated?
> 
> I never said this PR suggest a rule, i only said that Freedoom never has any obligation in this specific part, what are you trying to say?

I never suggested a rule either but there is a lack of obligation in this area causing a leaning of direction just because of the majority.
> 
> > And that was down to personal preference.
> 
> I know and i don't blame anyone who didn't play on 4:3, it's just that many source ports opted for it, and 4:3 is the intended way (check the wiki for proof), otherwise id wouldn't be bothered with monster sprites and title screen.
> 
So were just going with the majority. That's not exactly direction. And which wiki are you on about? I've never read anything from id to state that was their intention.
> Also further proof: rhcZdi2

Your proof is flawed but it doesn't explain Doom's round sprite assets.

mc776:
At this point I think it's best to just summarily dismiss this guy who just seems to be here to argue instead of, like, literally *look at* anything.

fawtytoo:
> At this point I think it's best to just summarily dismiss this guy who just seems to be here to argue instead of, like, literally _look at_ anything.

Not to argue, just to understand why it's being pushed. I'm guessing there was a discussion on this before the PR was raised. Would you care to point me in the direction of that?

> and 4:3 is the intended way (check the wiki for proof)

And I would at least like to read this proof. Just tell me where it is.

Anonysussy:
> And I would at least like to read this proof. Just tell me where it is.

https://doomwiki.org/wiki/Aspect_ratio this also applies to many DOS games, though not all of them. 

If you want to argue further, I'm going to dismiss you like mc said. You're just ragebaiting and confusing yourself at this point.

mc776:
If someone's going to object to this on the basis that Doom's tall oval "spheres" are the better way to go then that would be valid (if not necessarily universally agreed upon), but that doesn't seem to be what's going on here.

Anonysussy:
> If someone's going to object to this on the basis that Doom's tall oval "spheres" are the better way to go then that would be valid (if not necessarily universally agreed upon), but that doesn't seem to be what's going on here.

I think the person is more concerned about the sphere looking ass considering how they're insisted on how 4:3 is wrong and it didn't need aspect ratio correction. 
Screenshot_20260526_111751_Brave
Imo it's a selfish assertion to enforce personal preference as the correct way despite contradicting evidence and majority consensus. Alongside the defense being pretty clumsy all around.

mc776:
....wait, there are people going around playing Doom with *everything* squished with the *playsim* pixels perfectly square? o_O

Anonysussy:
> ....wait, there are people going around playing Doom with _everything_ squished with the _playsim_ pixels perfectly square? o_O

Most newcomer probably did, and some do like pixel perfect in their games which I don't mind (it helps scale pixels cleanly in case integer scaling were involved), it's just become a problem when they insist on it as an all correct solution when it isn't. 

Another case being PS1 games tend to use weird resolutions
Crash Bandicoot - Warped (USA)-260526-150632

Blastfrog:
I'm fond of the general idea of aspect ratio correction on the spheres, but I don't like the resulting assets in this PR. First off, it looks blurry. I think upscaling first with a decent/appropriate ESRGAN model and then downscaling with fant sampling (or some other algorithm that retains sharpness while still decently antialiasing) would be better.

Given that the PAR in Doom is 1.2, I think targeting 1.1 width (smack in the middle between 1.0 and 1.2) and 0.91666~ height is ideal in order to retain the existing perceptual volume while redistributing it.

As for any concerns that Doom (or any other old games on various platforms) were designed to look "acceptable" in differing aspect ratios, I think that can all be safely disregarded. Doom PAR is defacto 1.2, and any source ports worth a damn treat it as such (forget Doom 95). User experience in the vast majority of likely use cases will treat PAR as 1.2, so that's the target to work with. Any minor edge case is exactly that.

PR #1579, opened 2026-03-30
textures/flats: darken Aquatex green grass.

mc776:
AQDIRT01 is basically totally unused. It's only found on Map05, but not in any visible part of the map, and even that mysteriously disappeared when I fixed a seemingly unrelated bug in Lotan's Tomb.

It's pretty clear why when you try to pair it with AQF022: the resulting bright green looks really unnatural and off. I've not seen any map where it would have been appropriate. Darkening it seems to make it fit much better.

AQF022 shouldn't be darkened directly, however, since it's actually used (and AFAICT *only* used) for a few spots to match another unnaturally bright green thing (cf. the spectre ambush portal on Map11), and so a new flat would need to be added.

Before, using old AQDIRT01 and AQF022:
DOOM00

After, using darkened AQDIRT01 and AQF022A:
DOOM00


Xindage:
i would rather replace the flat instead, maybe something could be used in map11 no, or that change affect it do badly?

mc776:
That would actually be what I'd rather do... I'll take a look later to see what other bright green flats there are.

mc776:
There is no other FD flat that is bright green like this.

Here's how much it changes when you pair the new darkened flat with existing bright green walls:
DOOM00
DOOM02

DOOM01
DOOM03

mc776:
After some thought I believe it best if this PR were a straight-up edit of the two existing graphics, and if the change in green for those two specific spots bothers anyone something can be done later.

PR #1583, opened 2026-05-04
Plutonia door-friendly small BOSF

SenhorFlibble:
Apologies for the false-start, I forgot about the capital letters.

Anyway, here's a scaled-down version of Eriance's Guardian for the "boss cube" replacement that I created for the [Theseus](https://www.doomworld.com/forum/topic/158037-freedoom-theseus-a-conservative-fork/) fork.

I couldn't think of a better solution for PWAD compatibility, as some _Plutonia_ PWADs, like [Urania](https://www.doomworld.com/idgames/levels/doom2/megawads/urania), use custom _Plutonia_-style doors, and the current BOSFA0 looks off and rather bad on them.

Here's an example of a door in _Urania_ with my version:
urania
Not perfect, but I'd say it's better than the unedited sprite. I didn't want to go the straightforward route and create a rotating box with a skull texture on it, although that is certainly possible and I even whipped up a very crude test version. But it's too similar to the original, I think.


mc776:
Could you post a couple ingame boss brain level screenshots? Aesthetically I like this a lot better than the current FD sprite but I do appreciate how the current one is so much more visible especially from far away on big open maps like Map30.

SenhorFlibble:
> Could you post a couple ingame boss brain level screenshots? Aesthetically I like this a lot better than the current FD sprite but I do appreciate how the current one is so much more visible especially from far away on big open maps like Map30.

Here you go, these were made in MBF in DOSBox with v0.10 MAP30:
map30-1
map30-2

PR #1592, opened 2026-05-23
Beefier punch sound

SeventhSentinel:
I could barely hear the new punch sound over the music and enemies; it came across more like a dry clap. I decided to do some post-processing, adding a tiny bit of reverb and layering on some low FM synth with some bassy EQ. Melee impacts sound much more substantial now.


Saver310 (commented):
I don't hear a difference

SeventhSentinel:
Direct comparison: [Before](https://github.com/user-attachments/files/28182506/DSPUNCH.wav)/[After](https://github.com/user-attachments/files/28182483/dspunch.wav). I pretty much added as much bass frequencies to it as I could without going overboard.

mc776:
It feels like you're punching a hollow metal wall instead of a heavy bag.

EDIT: I remember also feeling like current sound was hard to hear when it was new, but after I got my current setup with a decent subwoofer it was fine. Have you tried increasing the treble instead of the bass?

PR #1601, opened 2026-06-15
Adjustments to MAP01

Blastfrog:
I figured MAP01 was in need of a small tuneup. I saw that it had some vertices and other details removed last year, and I felt that some of the simplifications went too far. I figured this was also a good time to tune up some aspects of the stage in general.

I've readded some detail, and using CRL the highest I could get the visplane count was 112, so it should still be perfectly playable in Vanilla without a visplane overflow crash. If there is some (highly unlikely) extreme edge case where it somehow exceeds 128, I'm not terribly concerned to be honest - someone would knowingly have to go well out of their way to trigger that, and I doubt it's even possible on this map. Some merged sectors have been un-merged (those with randomly blinking lights). It doesn't save on visplanes, and it breaks the "glitchy" effect that the randomization provides (it's very strange to see broken lights flickering in sync).

I've removed the strange broken flooded area behind the blast doors (and closed them) in the exit room for numerous reasons. First of all, it was only added because I wanted to keep it as another homage to Jon's The Cursed Hangar. In a much earlier iteration of my MAP01 overhaul before I revealed it publicly, I had recycled the long nukage filled hallway leading up to the exit room but had retextured it to water and had the height already raised. I felt it was a bit too large and filler-y for a densely packed first stage, so I replaced it with that Mega Man-esque "pre boss room" hallway.

It really never made any sense to be behind the blast doors to begin with. What was this room pre-disaster? Some enclosed brick filled space? Why would blast doors lead to that? I've never been satisfied with the attempts to open the crack further, the chains just look weird and out of place. And since the player can't get to it anyhow, why tease them with it? So gone it is.

As an aside, I reduced the height of the red armor secret from 80 tall to 72 tall, as I felt it fit the texture dimensions better. I also tweaked the texturing a bit, using FLAT20 for the top of the door and DOORTRAK for consistency with other doors.

Some sector heights of the rocky background in the northern courtyard were slightly reduced, so as to prevent the sky from bleeding through when facing the psuedo-3D floor hovering above the teleport pad. I've also adjusted the use of not-seen-on-automap linedef flags to more closely represent what the player sees, rather than so aggressively culling it to only the playable space.

There were still many remnants of my old attempt to use Boom's transfer light function. I presume these were removed for a reason (consistency between engines, I suppose), so I've retained its removal for the time being. I removed all sector and linedef tags relating to it, and merged a sector from the light cast from the ceiling in the starting area as it is redundant. I was unaware when I added the light transfers that unknown line types would crash vanilla if the player attempted to press use against them. However, if there is no opposition to re-adding them so long as the lines are entirely inaccessible to the player, I'd be interested in doing so.

I re-split the blue light textured lines on the corners of the armor pedestal; it didn't sit well with me in how they were awkwardly cut off on their right sides. I also re-added the caution line corner sectors on the platforms the imps stand on in the middle courtyard; as above, I felt that it was over-simplified to the detriment of the aesthetics. The waterfalls all use 3 lines each now for consistency and aesthetics.

I felt that there weren't enough marine corpses in the stage, so I dotted a few more in, including a gibbed one for variety. In retrospect, I feel that the two stimpacks in the starting area were pretty useless; most players don't even realize that they're there (I've watched many YouTube playthroughs of the map to study player tendencies) and it's awkward to traverse back to the very start just to grab 20 health. I've replaced them with two multiplayer-only medikits, primarily for deathmatch so that room isn't empty and purposeless there. I suppose it works well as extra health if needed in co-op as well.

I felt that the two clips behind the lamps in the alcoves in the second room were a weak reward for going into the alcoves, so I replaced them with stimpacks. I moved the two clips to be in front of the window with the imp on the higher platform, so that they're still present in the room, and it guides the player toward the window as both a trap to be attacked and to demonstrate more clearly to new players that they can A. be attacked through windows and B. that there's a risk and reward factor to grabbing goodies in vulnerable spots. Also worth noting that there was a strange bug in vanilla where two of the armor bonuses in the pit would move to a higher height when the two side windows lower, moving them inward by 8 units resolves this issue.

I used wadptr as Anonysussy also did in their commit. Conveniently, this version comes down to about 15.7kb less. If there are any issues with this pull request, please let me know, I'm open to any and all feedback.


Anonysussy:
I'm the one who did the simplifications (except for the open area that were separate PR done by someone else) and I did made some weird changes, some of these I like to see reverted like the thick door near the end that wasn't that thick originally and the silver texture in the outer area that was made around the time I'm being allergic to AQUATEX vanilla rendering quirk, I'm working on edits and remake for those textures, so it's safe to revert. Also yeah the simplifications in geometry and vertices were also influenced by IDCLIP crash which is why some went farther.

Blastfrog:
Players don't use no-clip in standard play, so that wasn't a consideration for me, nor do I think it should be a consideration for render limits in general. Yeah, the extra thick door is a little odd, but I didn't necessarily mind it. I agree that it should probably be turned back to 16 units thick and given the DOORTRAK texture.

You mentioned a silver texture in an outer area, but I'm not sure what you're specifically referring to, can you please clarify?

Anonysussy:
> You mentioned a silver texture in an outer area, but I'm not sure what you're specifically referring to, can you please clarify?

It's this one right here.
Screenshot_20260615_144945

The other changes I forgot to mention was the STEP texture replaced with FLAT texture in the area that had water between, but I don't mind it as much, just thought of mentioning it. 
Screenshot_20260615_145334

Blastfrog:
Just made a few more tweaks, mostly minor texture alignment (and some optimization that got it down a few more kb) and changed some texture assignments. To be honest, I feel okay with the silver texture on the platforms, I like it. 

I did change the stairs to match their trim more, as well as to distinguish it more from the trim around the water (I didn't like the idea of them both using the same flat). It also avoids some ugly cutoff on the last stair due to its different depth. I also did away with that ugly mess of geometry in the first hallway at the start area and changed the texture.

Xindage:
be carefull with wadptr it has a know issue of causing maps to fail in zandronum (i dont know why but it does sometimes, i know because it happened to maps that i did when i used it) can you test it?

Anonysussy:
> be carefull with wadptr it has a know issue of causing maps to fail in zandronum (i dont know why but it does sometimes, i know because it happened to maps that i did when i used it) can you test it?

Imo, it's outdated ZDoom versions quirk (which Zandro is based on), vanilla port like choco is pretty much fine.

(EDIT : Ig you could make an issue in wadptr GitHub)

mc776:
I might as well mention it since you're making changes: there's an old slime trail that you can see by moving straight forwards from the start. Moving some of the vertices on the light gradient gets rid of it. This might be fixable without compromising that straight line by scaling that entire gradient structure.

Blastfrog:
I fixed the slime trail. I think this iteration of the map is ready to go in, barring any objections.

I tested it in Chocolate Doom, UZDoom and Zandronum, I didn't encounter any problems with the map.

PR #1603, opened 2026-06-16
Minor shotgun improvements

Invictus812:
Sprite: Pushed the handle out to keep it consistent before the pumping frame (you pump a shotgun _fast_; with three frames you shouldn't see the handle between "all the way out" and "all the way in").

BuildCFG: Adjusted shotgun offsets for a more natural, punchier motion.

~DeHackEd: Order and timing adjustments for Shotgun and SSG flashes.~
^ Reverted


Blastfrog:
The Dehacked patch should only be used to replace text strings and fix minor bugs in the original state table. Altering frame assignments and durations for subjective taste is outside of that scope.

Invictus812:
Yeah, y'know I may have gotten too ambitious with that.  The sprite changes were the real point of this to start with.  Reverted my changes to the DeHackEd file.

Blastfrog:
For what it's worth, if you wanted to swap the firing frames, you could still do so by swapping the sprites themselves. I'm of the opinion that a shrinking muzzle flash looks better than a growing one.

Invictus812:
Aye, but default behavior for both weapons is that the first flash state calls A_Light1 and the second calls A_Light2.  It's subtle but noticeable depending on the lighting conditions in-game.  In order to be consistent, I could swap the frames around but I would feel obligated to also change the pointers in DeHackEd.  Maybe that method doesn't go overboard for y'all the way swapping the whole states does though?

Xindage (changes requested):
why the dehacked strings mostly related to freedm are removed, it does not seems to relate even to the shotgun

Invictus812:
I don't know what happened with the dehacked file tbh.  I'd meant to try to remove it from this pull request but I guess I just borked it instead.  I've just taken the file from the master branch and pushed that to my fork.  It should be back to normal now as if I'd never touched it.  Still kinda new to forks and pull requests, sorry :/

PR #1605, opened 2026-06-21
look at #1606

quarry24:
BRNOPEN uses an old version of STONE6 as its base, and CAVERN5 shares the same texture (for placeholder reasons maybe). this updates both of them to be consistent with their parent textures
if i did something wrong in the pr (im very paranoid) let me know
 
before and after for BRNOPEN in tntgb.wad's MAP02:
nugg0000
nugg0001

PR #1606, opened 2026-06-23
updated BRNOPEN & CAVERN5 (again)

quarry24:
same as #1605 but without all the old commit junk

btw, i made a possible alternative for BRNOPEN that uses STONE7 as its base instead of STONE6, due to it already having a cracked looked to it. i just cut out all the rocks using the cracks as a guide
nugg0003
if people say it looks better i could replace it


quarry24:
(are the thumbs up a responce to the patches in the pr or the patch in the screenshot...?)

Xindage:
credits seems to be missing again

quarry24:
that is intentional, as i am already listed in the credits file

PR #699, opened 2020-08-14
sprites: new gib animations (zombie and shotgunner)

HorroMovieGuy:
Improved gib animations for the aforementioned enemies with realistic gore as opposed to the abstract melting currently in place.

PR #1607, opened 2026-06-28
sprites: better sailor gib animation

leikaisho:
honestly it's kind of hard for me to say anything about this one that isn't too similar to what horrormovierei said in #699


Saver310:
I've made some minor touch ups too

Blastfrog:
No offense, but while this is a clear improvement on the existing assets, I honestly fail to see the point of this since the sailor is barred from use in Freedoom maps, as well as the sailor assets being of poor quality in general and unfitting with the current aesthetic direction of the project.

I'd rather see the sailor replaced wholesale with a more interesting set of sprites. There have been attempts before, but it appears they never gained traction.

I'm still in favor of making an homage to Woolball in regards to the Wolfenstein assets.

mc776:
SS still appears in PWADs

Talon1024:
> SS still appears in PWADs

Yes, and one example is on Plutonia 2 MAP29, and Plutonia 2 works well with the Freedoom IWAD.

